Issue #130, opened 2015-10-07
The release version is not reset in the rebased SPEC


thozza:
When I copied the rebased SPEC to the dist-git directory I got:
```
thozza@thozza-pc ~/devel/fedora/unbound (master *$%=)
$ git diff
diff --git a/unbound.spec b/unbound.spec
index 448f1b8..6f68cdb 100644
--- a/unbound.spec
+++ b/unbound.spec
@@ -20,7 +20,7 @@
 
 Summary: Validating, recursive, and caching DNS(SEC) resolver
 Name: unbound
-Version: 1.5.4
+Version: 1.5.5
 Release: 5%{?extra_version:.%{extra_version}}%{?dist}     <--- THE OLD RELEASE
 License: BSD
 Url: http://www.nlnetlabs.nl/unbound/
...
```

PR #132, opened 2015-10-22
Minor issues

### LINKED ISSUE #129: Rebased SPEC contains broken changelog entry
After I copied the rebased SPEC back to the dist-git directory, the changelog entry was broken:
- didn't contain the release number after the version
- the email was missing, instead of it there was 'None'

diff:
```
... 
%changelog
+* Wed Oct 07 2015 Tomas Hozza <None> - 1.5.5
+- New upstream release 1.5.5
+
 * Tue Sep 15 2015 Tomas Hozza <thozza@redhat.com> - 1.5.4-5
 - Removed dependency and ordering on unbound-anchor.service in unbound.service
...
```

nforro:
Closes #129, #130

### DIFF
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -160,8 +160,7 @@ class Application(object):
             logger.debug("argument passed as a new source is a version")
             version, extra_version = SpecFile.split_version_string(self.conf.sources)
             self.rebase_spec_file.set_version(version)
-            if extra_version:
-                self.rebase_spec_file.set_extra_version(extra_version)
+            self.rebase_spec_file.set_extra_version(extra_version)
 
     def _initialize_data(self):
         """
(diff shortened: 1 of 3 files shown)


rebasehelper/application.py
```diff
@@ -142,8 +142,7 @@ def _prepare_spec_objects(self):
             logger.debug("argument passed as a new source is a version")
             version, extra_version = SpecFile.split_version_string(self.conf.sources)
             self.rebase_spec_file.set_version(version)
-            if extra_version:
-                self.rebase_spec_file.set_extra_version(extra_version)
+            self.rebase_spec_file.set_extra_version(extra_version)
```
phracek: I guess, that it was correct.
What about in case if extra_version doesn't exist?
Will it work?

nforro: It will work, there is another check for empty extra_version in set_extra_version method.
Also, when archive name is passed as an argument to rebase-helper and set_extra_version is called from set_version_using_archive method, there is no check for empty extra_version there either ([code](https://github.com/phracek/rebase-helper/blob/3820af9ffe8ff62e76d3afed467bd52d06860fd3/rebasehelper/specfile.py#L858)).

PR #133, opened 2015-10-23
Fix logging of report from output tools

nforro:
Changes introduced in commit [3820af9](https://github.com/phracek/rebase-helper/commit/3820af9ffe8ff62e76d3afed467bd52d06860fd3) are causing rebase-helper to traceback because of non-existent logger_report in output_tool.py. This PR fixes that.


phracek:
THX

PR #135, opened 2015-11-10
Removed patchtool option

uhliarik:


rebasehelper/application.py
```diff
@@ -351,7 +351,7 @@ def patch_sources(self, sources):
         git_helper = GitHelper(sources[0])
         if not self.conf.non_interactive:
             git_helper.check_git_config()
-        patch = Patcher(self.conf.patchtool)
+        patch = Patcher("git")
```
phracek: What do you think to define it globally.

Git command is already used here https://github.com/phracek/rebase-helper/blob/master/rebasehelper/utils.py#L490

uhliarik: Ahh, I was not able to find anything like this, thank you for comment, I will fix it ASAP.

thozza: I think the Patcher class should be completely removed and GitPatchTool used directly. To minimize the need to refactor the code you could rename the ```run_patch()``` to ```patch()```. The Patcher tool was mostly a wrapper to allow us to execute different patch tools in the same manner.

thozza:
Lubos, since you created PR from master to master, there is unrelated commit to the change you are claiming you are doing. I mean the commit "Fixed undefined variable (defined in different scope)". I don't think it is worth the effort to split it in this case, but keep this in mind when creating PR in the future. It is good practice to clearly separate unrelated changes.

rebasehelper/application.py
uhliarik: Ah OK, I thought, one day, you can want to add new patcher tool, so I left it there as is, only removed the option. I can cleanup the code in next iteration.

uhliarik:
@thozza as we talked today, I will create a separate branch for each changeset in future. Thanks for the note.

PR #171, opened 2016-04-27
Copr support

nforro:
This is initial support for Copr as a build tool. It is based on fedpkg implementation, so the usage is similar. For example:
```
rebase-helper --buildtool copr \
              --builds-nowait \
              --copr-build-ids $OLD_BUILD_ID,$NEW_BUILD_ID \
              $NEW_VERSION
```

There are two new dependencies:
- copr (*python-copr* in Fedora)
- pyquery (*python-pyquery* in Fedora)

You will need to get your Copr API token: https://copr.fedorainfracloud.org/api/

Any comments are welcome.

@phracek: Could you test this, possibly with upstream monitoring?

### DIFF
--- a/rebasehelper/build_helper.py
+++ b/rebasehelper/build_helper.py
@@ -28,6 +28,7 @@ import six
 from rebasehelper.utils import ProcessHelper
 from rebasehelper.utils import PathHelper
 from rebasehelper.utils import KojiHelper
+from rebasehelper.utils import CoprHelper
 from rebasehelper.utils import TemporaryEnvironment
 from rebasehelper.logger import logger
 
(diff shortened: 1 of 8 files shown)


rebasehelper/build_helper.py
```diff
@@ -615,6 +616,154 @@ def build(cls, spec, sources, patches, results_dir, **kwargs):
...
+                             'SRPMS']:
+                self._env[self.TEMPDIR + '_' + dir_name] = os.path.join(
+                    self._env[self.TEMPDIR_RPMBUILD],
+                    dir_name)
+                logger.debug("Creating '%s'",
+                             self._env[self.TEMPDIR + '_' + dir_name])
+                os.makedirs(self._env[self.TEMPDIR + '_' + dir_name])
+
+    @classmethod
+    def _build_srpm(cls, spec, workdir, results_dir):
```
thozza: This seems to be the same code as in RpmbuildBuildTool class. It would be better to either inherint from the class and redefine the rest of the methods, or extract the SRPM building into a new class and then both classes should inherit from it. Duplicating of code is error prone.

nforro: Right. Same code is used also in FedpkgBuildTool class.
I think it may be best to introduce SRPM build tools and let users choose between *rpmbuild* and *mock* for SRPM building.

thozza: I was thinking about not using mock for SRPM at all. It basically does not make sense. I think rpmbuild should be build always used for building SRPMs.

phracek:
@nforro Thanks for implementation. Before releasing into upstream monitoring service it has to be clarified with upstream release monitoring team.
I will initiate a discussion.

We have to create a group in COPR, called like 'upstream-monitoring-service' or 'the-new-hotness' and all builds will be done there.

As we discussed with @xsuchy, I would prefer to use python API instead of bash commands.

I can test on my instance, where the-new-hotness is running, if rebase-helper goes fine.

But let me check rebase-helper and COPR together first.
I would like to include COPR and rebase-helper into upstream before a holiday.

nforro:
> As we discussed with @xsuchy, I would prefer to use python API instead of bash commands.

If you mean using Copr API, it is being used already. Just not directly, I'm taking advantage of *copr* python module.

phracek:
Please add to requirements also Copr.
Otherwise exception ImportError arise.

nforro:
I'm not exactly sure where do you want me to add it?

phracek:
I have tried to run rebase-helper <version> --buildtool copr --non-interactive but it fails with error Missing or invalid copr configuration file.

phracek:
I will add to README.md requirements like copr-cli and python-pyquery. 
These two packages are needed for proper rebase-helper and COPR functionality.

phracek:
I see the problem. Configuration file for COPR is missing. I will create the one for instance.

nforro:
> I have tried to run rebase-helper --buildtool copr --non-interactive but it fails with error Missing or invalid copr configuration file.

You need to add configuration generated [here](https://copr.fedorainfracloud.org/api/) to *~/.config/copr*.

> I will add to README.md requirements like copr-cli and python-pyquery. 

*copr-cli* is not needed, *python-copr* is.

phracek:
Well, I tried to build package called springlobby (Fedora package name) .
~~~
rebase-helper 0.247 --buildtool copr --non-interactive -v
~~~
but COPR build is not detached.

https://copr.fedorainfracloud.org/coprs/phracek/rebase-helper-springlobby/build/182958/

From master log I see traceback like:
~~~
  File "/usr/share/copr/dist_git/srpm_import.py", line 92, in actual_do_git_srpm_import
    commands.commit(message)
  File "/usr/lib/python2.7/site-packages/pyrpkg/__init__.py", line 1282, in commit
    self._run_command(cmd, cwd=self.path)
  File "/usr/lib/python2.7/site-packages/pyrpkg/__init__.py", line 903, in _run_command
    % (' '.join(cmd), proc.returncode, error))
rpkgError: Command git commit -m import_srpm -a returned code 1 with error: 
refreshing cgit listing
sending a response - success
Sending back: 
{"pkg_version": "0.195-9.fc23", "git_hash": "1d4c918f839cea91553f1a753ad4357a9c5fd094", "pkg_name": "springlobby", "task_id": "182958-master", "repo_name": "phracek/rebase-helper-springlobby/springlobby"}
~~~

It seems like we have a COPR traceback, right?

nforro:
> but COPR build is not detached

Did you forget *--builds-nowait*?

> It seems like we have a COPR traceback, right?

Yes, but despite the error it was imported and built successfully. Weird.

phracek:
You are right. --builds-nowait was missing.

phracek:
We use fedpkg-build-tasks. Shall we use copr-build-tasks instead of copr-build-ids?

I guess, it could have the same options, right?

nforro:
There are no tasks in Copr, just builds. But I guess we could call them tasks, if you want to use the same argument fort both.

phracek:
I have tried all steps and COPR seems to be working.

nforro:
Great. I'll address @thozza's hints and this could be merged.
What about *--copr-build-ids*? Should I merge it with *--fedpkg-build-tasks* and use single *--build-tasks* argument?

phracek:
@nforro I think, personally, it make sense to have the same or similar options.

thozza:
@nforro I think the merge makes sense. And then on the build tool level let the code handle it. It does not make sense to have different options for basically the same thing.

@phracek I'm not sure if it is good to add copr as hard dependency. Maybe it is good to use soft dependencies in RPM. In the code you can just catch the ImportError and maybe inform the user that it may be good to install COPR if he wants to use it. Just an idea that for build tools it may not be good to require all of them to be available. It will also make things easier when possibly using rebase-helper on different RPM distribution.

nforro:
Code duplicates have been removed and all build tools now use rpmbuild for SRPM building.

There is new *--build-tasks* argument usable with both *fedpkg* and *copr* build tools.
I left *--fedpkg-build-tasks* working for backward compatibility, but it is hidden.
Or should I remove it for good?

@phracek @thozza please review.

thozza:
@nforro as for the ```--fedpkg-build-tasks``` option, I think only the-new-hotness is using it. My opinion would be to keep it for one release and display warning when using it. In the meanwhile we need to migrate the-new-hotness to use ```--build-tasks``` and then we can remove the ```--fedpkg-build-tasks``` option fo good in the next version. @phracek does it make sense?

phracek:
@thozza You mean, like fedpkg-build-tasks and copr-build-tasks will be joined to one build-tasks? Then yes, it makes sense and I agree with this approach.

thozza:
@phracek yes to your question.

phracek:
Ok, then I guess, We can merge it.

PR #172, opened 2016-04-27
Added support for GitLab CI on Fedora Docker images and testing cleanup

thozza:
- builds are now done in GitLab CI on top of Fedora Docker images. Rebase-helper is now tested using Python2 and Python3 on top of F22, F23, and rawhide (https://gitlab.com/rebase-helper/rebase-helper/builds).
- moved tests into rebasehelper package
- removed self-made function for converting strings and using six instead
- fixed issues found by pylint
- reworked tox.ini so that one is able to run tests locally and it is also usable for Travis CI and GitLab CI
- removed setup.py test command, as it is not needed any more, TOX should be used instead.


thozza:
I see that pylint is failing in Travis CI due to rpm and pycurl. I'm not sure if ti makes sense to resolve these issues for the sake of using Travis, since it is using Ubuntu.

@phracek @nforro please review

pylintrc-travis
```diff
@@ -0,0 +1,363 @@
...
+[TYPECHECK]
+
+# Tells whether missing members accessed in mixin class should be ignored. A
+# mixin class is detected if its name ends with "mixin" (case insensitive).
+ignore-mixin-members=yes
+
+# List of module names for which member attributes should not be checked
+# (useful for modules/projects where namespaces are manipulated during runtime
+# and thus existing member attributes cannot be deduced by static analysis
+ignored-modules=distutils,rpm,pycurl
```
nforro: The modules can be added to *extension-pkg-whitelist* instead. That way pylint would still check their attributes.
Also it might not be necessary to create separate pylintrc file for Travis.

rebasehelper/tests/test_utils.py
```diff
@@ -527,7 +527,7 @@ def test_all_packages_installed_one_non_existing(self):
 class TestMacroHelper(BaseTest):
 
     def test_get_macros(self):
-        rpm.addMacro('test_macro', 'test_macro value')
+        rpm.addMacro('test_macro', 'test_macro value')  # pylint: disable=no-member
```
nforro: I guess the comment is not needed if the rpm module is ignored or whitelisted.

pylintrc-travis
thozza: Pylint can check these modules when run locally or in gitlab ci. It does not work in travis. Module members is exactly what was not working

rebasehelper/tests/test_utils.py
thozza: Dont guess and check older travis builds where it was not working even the module was ignored.

pylintrc-travis
thozza: Please describe how can we have different pylint configuration for travis without a different pylintrc file. I was unable to find better way. Also note that pylint is run as part of pytest, not executed separately.

nforro: I'm not sure why it works locally or in Gitlab CI, but whitelisting the modules makes it work in Travis. It doesn't work now because rpm and pycurl are untrusted C extensions (not from standard library).
See https://github.com/PyCQA/astroid/blob/d272438213e6235f642d864085b10768b501cade/ChangeLog#L566.

> Please describe how can we have different pylint configuration for travis without a different pylintrc file.

The point was that there would be no need to have different pylint configuration. Why couldn't the modules be whitelisted locally or in Gitlab CI?

thozza:
@nforro I whitelisted the pycurl and rpm modules and the Travis CI works OK now. My understanding is, that the reason why it worked locally and in GitLab CI, is that there are the modules installed by RPM in the system-wide site-packages directory, while in Travis CI these are installed using pip and therefore are not trusted.

@nforro @phracek please review, there should not be any issues left.

nforro:
There is one more thing.
Documentation builds on [readthedocs.org](http://readthedocs.org/projects/rebase-helper/builds/) are trying to install dependencies from requirements.txt (if present) and failing on pycurl. It's not visible now because the builds are triggered for master branch only.
I "resolved" this before by removing requirements.txt and moving dependencies to tox.ini. But I think it is possible to specify different requirements file in *readthedocs* settings, so something like empty doc-requirements.txt should to the trick.

nforro:
Also, I noticed that tests in Travis are now run under python2 only. There should be another travis environment with python3 basepython added to tox.ini.

thozza:
@nforro As to the readthedocs issue, I'll have a look at it. Removing pycurl from requirements.txt just to make RTD happy is not a good solution. I'll add the py3 version for Travis.

thozza:
@nforro @phracek all seems to be fixed now. Please check.

nforro:
Looks good to me now.

doc-requirements.txt
```diff
@@ -0,0 +1,6 @@
+# This can not be installed from pip, on Fedora you need to install rpm-python{,3}
+#rpm-python
```
phracek: What about to remove things which are not needed. I would prefer to remove it completely.

requirements.txt
```diff
@@ -0,0 +1,5 @@
+# This can not be installed from pip, on Fedora you need to install rpm-python{,3}
+#rpm-python
```
phracek: I would prefer to remove it if it is not needed.

doc-requirements.txt
thozza: The intention is to list all dependencies with comments why these don't apply to the use case. The bottom line is that when installing rebase-helper or packaging it for a distribution, you need to add these dependencies. When removed, the information will be scattered in multiple places. I would prefer to have the list complete and this is the reason why I included it with the comment. Unless you have a really strong reason why to remove this, I would like to keep it in the file. It is harmless and provides additional information.

phracek:
Can you please explain me a bit more why tests are inside rebasehelper package?
Do we want to provide it inside a rebase-helper package on Fedora?

Does it fulfill Fedora Packaging Guidelines.

requirements.txt
phracek: Ok, the same as above.

thozza:
@phracek Fedora packaging guidelines are completely separate thing to where the project stores tests. Fedora packaging is a question of SPEC file. The change I made is a question of Python common practices. It is a good practice to include tests in the package, so that anyone using your package can also use your tests and directly import them if needed. Tests were distributed also before anyway. 

Plus this way we can check also tests using pylint, which is IMHO a good thing and it was not done before.

E.g. discussion about the place where to put tests is http://stackoverflow.com/questions/61151/where-do-the-python-unit-tests-go

thozza:
@phracek I checked Fedora packaging guidelines and they don't discuss tests at all. Only mention is that if the package contains test suite, it should be executed during package build. Tests are code like any other.

phracek:
ok, then I am fine with merge.

thozza:
@phracek @nforro thank you for the review!

PR #179, opened 2016-05-05
All checkers are loaded automatically from rebasehelper/checkers folder

phracek:

This Pull Request loads all checkers from rebasehelper/checkers dir automatically.
Of course each checkers has to implement required functions.


thozza:
I think that using entry points and pkg_resources is cleaner approach (http://aroberge.blogspot.cz/2008/12/plugins-part-6-setuptools-based.html), however we can iterate on the solution later.

phracek:
Yeah, seems reasonable.
I would like to move also builders to the similar plugin stuff.

phracek:
What do you think about it. Can we merge it?

nforro:
Fine by me. Just one note: after merging with master, *TestPkgDiff.test_fill_dictionary* and *TestPkgDiff.test_process_xml* fail because tests were moved to *rebasehelper* package and the paths that are being passed to *Checker* class constructor are wrong.

Issue #177, opened 2016-05-03
please, provide documentation


TomasTomecek:
When I look at main readme, I have literally no idea how to use this tool.

I'm sorry but this doesn't do it:

```
How to execute rebase-helper from CLI

Go to a git directory of your package and execute command /path_to_rebase-helper/rebase-helper.py (e.g. /home/user/phracek/rebase-helper/rebase-helper.py "3.1.10")
```

nforro:
@TomasTomecek is it any better now?

TomasTomecek:
That's much better, thank you!

PR #180, opened 2016-05-12
Update README.md

nforro:
This is an attempt to make the readme a bit more useful and up to date.
See issue #177.

As for the changes in requirements:
- **meld**: not really necessary, users can use any merge tool
- **git**: git is essential, it should be probably added to Fedora package as a dependency too
- **pkgdiff at least 1.6.3**: I think the version limitation is not necessary, as 1.6.3 is in Fedora since Fedora 19
- **python-six**: it's already in requirements.txt

@phracek @thozza your opinions?

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,42 +2,62 @@
 
 [![Code Health](https://landscape.io/github/phracek/rebase-helper/master/landscape.svg?style=flat)](https://landscape.io/github/phracek/rebase-helper/master) [![GitLab CI build status](https://gitlab.com/rebase-helper/rebase-helper/badges/master/build.svg)](https://gitlab.com/rebase-helper/rebase-helper/commits/master) [![Travis CI build status](https://travis-ci.org/phracek/rebase-helper.svg?branch=master)](https://travis-ci.org/phracek/rebase-helper)
 
-## Rebase-helper workflow
-
-This tool helps you to rebase package to the latest version
-
-How the rebase-helper works:
-- Each action should be logged and visible by user.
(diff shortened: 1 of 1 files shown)


phracek:
Yes. meld is not needed. I agree. It depends on user, how he has specified merging tool.
GIT is right. It should be in dependencies.
Regarding pkgdiff, If so, then we can remove it.

By the way, I guess README and requirements are especially for running rebase-helper from GIT directly and therefore python-six should be mentioned there, as well as libabigail, right?

nforro:
> By the way, I guess README and requirements are especially for running rebase-helper from GIT
> directly and therefore python-six should be mentioned there, as well as libabigail, right?

We have python dependencies in *requirements.txt* file, for easy installation with pip. As I see it, the requirements listed in the readme are external programs and they are not strictly needed to execute rebase-helper, but of course running it without them is rather useless.

If you think that python dependencies should be listed in the readme as well, I will add them, and I can also add a note that they can be installed with pip.

thozza:
@phracek I agree with @nforro, that Python packages needed for running rebase-helper should be listed in requirements.txt and only external programs needed by rebase-helper should be listed in README. Given that should be the pyrpkg be removed from README? Is it really explicitly needed? Do we use it directly? IMHO it is pulled in automatically as dependency of fedpkg.

phracek:
Ok, then, I am fine with that. You can merge it.

nforro:
I removed *pyrpkg* from the list since it's a dependency of *fedpkg* and added a note about python requirements.

Issue #174, opened 2016-05-03
Wrong (nonexistent) patches mentioned in bugzilla comment


mhlavink:
Got this comment:

Following patches has been deleted:
['idris-ASTBuilding-ghc78.patch', 'idris-rts-opt-fprintf.patch']

but no such patches are present in the package.
see https://bugzilla.redhat.com/show_bug.cgi?id=1332334

phracek:
Thanks for the bug. It seems to be really problematic bug. It has to be solved during the next release.

sergiomb2:
+1
https://bugzilla.redhat.com/show_bug.cgi?id=1333597

phracek:
Another problem https://bugzilla.redhat.com/show_bug.cgi?id=1336603#c3

phracek:
Log from the-new-hotness:

~~~
@phracek does this help?

[2016-05-17 00:25:41][    fedmsg    INFO] Handling anitya msg u'2016-712114ad-c803-4991-8a9e-605e92acd776'
[2016-05-17 00:25:44][    fedmsg    INFO] Comparing upstream 1.0.13 against repo 1.0.12-1.fc25
[2016-05-17 00:25:44][    fedmsg    INFO] OK, 1.0.13 is newer than 1.0.12-1.fc25
[2016-05-17 00:25:50][    fedmsg    INFO] Created bug: #1336603 NEW        - nphilipp@redhat.com - python-sqlalchemy-1.0.13 is available
[2016-05-17 00:25:50][    fedmsg    INFO] Filed new bug 'https://bugzilla.redhat.com/show_bug.cgi?id=1336603'
[2016-05-17 00:25:50][    fedmsg    INFO] publishing topic 'update.bug.file'
[2016-05-17 00:25:50][    fedmsg    INFO] Rebase-helper is going to rebase package
[2016-05-17 00:25:50][    fedmsg    INFO] Cloning 'http://pkgs.fedoraproject.org/cgit/rpms/python-sqlalchemy.git' to '/var/tmp/thn-rhNFAJ0j'
Downloading sources from URL http://pypi.python.org/packages/source/S/SQLAlchemy/SQLAlchemy-1.0.12.tar.gz
Downloading sources from URL http://pypi.python.org/packages/source/S/SQLAlchemy/SQLAlchemy-1.0.13.tar.gz
[2016-05-17 00:26:40][    fedmsg    INFO] Rebasehelper package python-sqlalchemy 1.0.13
[...]

PR #185, opened 2016-05-26
Clear output data on Application initialization

nforro:
When running multiple Application instances consecutively,
output data persist between runs and cause results to be incorrect.

Fix that by clearing output data just after Application instance
is created.

This resolves *the-new-hotness* related issue #174.


phracek:
Make sense to have it in def __init__
the-new-hotness always calls Application(cli) https://github.com/fedora-infra/the-new-hotness/blob/develop/hotness/buildsys.py#L184

phracek:
+1

Issue #184, opened 2016-05-26
Traceback when downloaded file acrtually does not exist.


thozza:
The problem is, that the URL returns 404 File not found. The download helper should determine that there is a problem and rebase-helper should not proceed at all.

```
thozza@thozza-pc ~/devel/fedora/bind (master $%=)
$ rebase-helper 9.10.4-P1
Downloading sources from URL https://ftp.isc.org/isc/bind9/9.10.4-P4/bind-9.10.4-P4.tar.gz
100%                     
Traceback (most recent call last):
  File "/usr/bin/rebase-helper", line 9, in <module>
    load_entry_point('rebasehelper==0.7.3', 'console_scripts', 'rebase-helper')()
  File "/usr/lib/python2.7/site-packages/rebasehelper/cli.py", line 148, in run
    app.run()
  File "/usr/lib/python2.7/site-packages/rebasehelper/application.py", line 655, in run
    sources = self.prepare_sources()
  File "/usr/lib/python2.7/site-packages/rebasehelper/application.py", line 358, in prepare_sources
    os.path.join(self.execution_dir, settings.NEW_SOURCES_DIR))
  File "/usr/lib/python2.7/site-packages/rebasehelper/application.py", line 322, in extract_sources
    Application.extract_archive(archive_path, destination)
  File "/usr/lib/python2.7/site-packages/rebasehelper/application.py", line 313, in extract_archive
    archive.extract_archive(destination)
  File "/usr/lib/python2.7/site-packages/rebasehelper/archive.py", line 275, in extract_archive
    archive = self._archive_type.open(self._filename)
  File "/usr/lib/python2.7/site-packages/rebasehelper/archive.py", line 151, in open
    return tarfile.TarFile.open(filename)
  File "/usr/lib64/python2.7/tarfile.py", line 1678, in open
    raise ReadError("file could not be opened successfully")
tarfile.ReadError: file could not be opened successfully
```

PR #188, opened 2016-05-27
Issue 184

thozza:
- Extended tests for DownloadHelper
- replaced pycurl by pure Python implementation - pycurl is not needed any more
- fixed issue #184

### DIFF
--- a/rebasehelper/utils.py
+++ b/rebasehelper/utils.py
@@ -27,7 +27,6 @@ import sys
 import fnmatch
 import subprocess
 import tempfile
-import pycurl
 import shutil
 import rpm
 import locale
@@ -213,17 +212,19 @@ class DownloadHelper(object):
     """Class for downloading sources defined in SPEC file"""
 
     @staticmethod
-    def download_file(url, destination_name):
+    def download_file(url, destination_path, timeout=10, blocksize=8192):
         """
-        Method for downloading file using pycurl
+        Method for downloading file from HTTP, HTTPS and FTP URL.
 
         :param url: URL from which to download the file
-        :param destination_name: path where to store downloaded file
+        :param destination_path: path where to store downloaded file
+        :param timeout: timeout in seconds for blocking actions like connecting, etc.
+        :param blocksize: size in Bytes of blocks used for downloading the file and reporting progress
         :return: None
         """
-        def progress(download_total, downloaded, upload_total, uploaded):
+        def progress(download_total, downloaded, start_time):
             r = downloaded / download_total if download_total else 0.0
-            t = time.time() - start
+            t = time.time() - start_time
             if 0.0 < r < 1.0:
                 h, rem = divmod(int(t/r - t), 3600)
                 m, s = divmod(rem, 60)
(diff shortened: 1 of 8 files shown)


rebasehelper/utils.py
```diff
@@ -227,34 +234,46 @@ def progress(download_total, downloaded, upload_total, uploaded):
...
-            curl.setopt(pycurl.FOLLOWLOCATION, 1)
-            curl.setopt(pycurl.MAXREDIRS, 5)
-            curl.setopt(pycurl.TIMEOUT, 300)
-            curl.setopt(pycurl.WRITEDATA, f)
-            curl.setopt(pycurl.NOPROGRESS, 0)
-            curl.setopt(pycurl.PROGRESSFUNCTION, progress)
-            try:
+        try:
+            response = urllib.request.urlopen(url, timeout=timeout)
+            file_size = int(response.info().get('Content-Length', 0))
```
phracek: Does it work also for PyPi and other stuff?

rebasehelper/utils.py
```diff
@@ -227,34 +234,46 @@ def progress(download_total, downloaded, upload_total, uploaded):
...
+
+            if not file_size:
+                raise DownloadError('The file has zero size')
+
+            # file exists, check the size
+            if os.path.exists(destination_path):
+                if file_size != os.path.getsize(destination_path):
+                    logger.debug("The destination file '%s' exists, but sizes don't match! Removing it.",
```
phracek: I guess, it should be logger.info. We should inform user, size of sources are different.

phracek:
I guess, we used already urllib but for some reason, we remove it. PyCurl worked better then urllib.
Unfortunately, I don't understand why:(

phracek:
You introduced DownloadHelperError, that's fine, but we should care about it here, right? https://github.com/phracek/rebase-helper/blob/master/rebasehelper/specfile.py#L174

rebasehelper/utils.py
thozza: it works with ANY URL.

thozza:
@phracek please check all commits.

rebasehelper/utils.py
thozza: This is only to decide if the existing file is correct or not. e.g. when you interrupt rebase-helper while it is downloading a file, the file will stay there, but the sizes will be different. We could improve it to delete the file in case the download didn't finish and there is KeyboardInterrupt, but I would leave that for different patch set.

phracek:
I see, I overlook RebaseHelperError.

thozza:
@phracek BTW CoprHelper uses urllib anyway. If there are some real issues with urllib I'm all for resolving them, but if not I'm all for pycurl, as I don't think we actually used any of its features that are not already in urllib.

rebasehelper/utils.py
thozza: @phracek but if you want to expose this message to the user I can change it. I don't have strong opinion about the log level of this message. :wink:

phracek: @thozza I would be glad if you can change it to info. THX

rebasehelper/utils.py
```diff
@@ -202,22 +202,29 @@ def capture_output(func, capture_stdout=False, capture_stderr=False):
...
         :param url: URL from which to download the file
-        :param destination_name: path where to store downloaded file
+        :param destination_path: path where to store downloaded file
+        :param timeout: timeout in seconds for blocking actions like connecting, etc.
+        :param blocksize: size in Bytes of blocks used for downloading the file and reporting progress
         :return: None
         """
-        def progress(download_total, downloaded, upload_total, uploaded):
+        def progress(download_total, downloaded, start_time):
             r = downloaded / download_total if download_total else 0.0
```
nforro: `downloaded` or `download_total` (or both) have to be casted to *float*, otherwise the progress shows only 0% or 100% and no estimated time.

nforro:
I'm fine with removing pycurl.
@thozza Could you remove it also from *docs/conf.py* and *doc-requirements.txt*?

rebasehelper/utils.py
thozza: It works for me. @nforro did you try it?

nforro: Sorry, I forgot to mention the problem is with python2 only. With python3 it works as expected.

thozza: right, I just realized that it does not work with Python 2 and I was testing using Python 3. I will fix it...

thozza:
@nforro I included your comments. As a bonus I added tests for the download progress function to make sure it we don't break it in the future.

thozza:
@phracek I also changed the log message regarding the file size to INFO as you wanted.

phracek:
Great thanks.

PR #194, opened 2016-06-08
Catch exceptions and write traceback to debug log

### LINKED ISSUE #95: The debug log should contain also traceback in case rebase helper ended with it
The debug log should contain Python traceback in case rebase-helper ended with it. It may be also helpful for regular errors. This should be included only in the debug log file, so the user will not see it.

nforro:
Fixes #95

### DIFF
--- a/rebasehelper/cli.py
+++ b/rebasehelper/cli.py
@@ -21,14 +21,22 @@
 #          Tomas Hozza <thozza@redhat.com>
 
 import argparse
+import logging
 import sys
 
 import six
 
-from rebasehelper.constants import PROGRAM_DESCRIPTION
+from rebasehelper.constants import PROGRAM_DESCRIPTION, NEW_ISSUE_LINK
 from rebasehelper.application import Application
-from rebasehelper.logger import logger
+from rebasehelper.logger import logger, LoggerHelper
 from rebasehelper.exceptions import RebaseHelperError
+
+
+class ArgumentParser(argparse.ArgumentParser):
+
+    def error(self, message):
+        self.print_usage(sys.stderr)
+        raise RebaseHelperError(message)
 
 
 class CLI(object):
@@ -36,7 +44,7 @@ class CLI(object):
 
     def __init__(self, args=None):
         """parse arguments"""
-        self.parser = argparse.ArgumentParser(description=PROGRAM_DESCRIPTION)
+        self.parser = ArgumentParser(description=PROGRAM_DESCRIPTION)
         self.add_args()
         self.args = self.parser.parse_args(args)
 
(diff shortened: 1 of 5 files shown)


nforro:
Traceback could be also hidden from user and only exception type and message could be shown.
@phracek, @thozza: What do you think?

thozza:
@nforro I agree with such approach. I even think that in console we could log just some generic message like this:

"rebase-helper failed due to an unexpected error. Please report this problem using the following link https://github.com/phracek/rebase-helper/issues/new and include the content of <path_to_debug_log> file in the report. Thank you!"

And the debug log will include the full traceback and exception type. What do you think?

rebasehelper/cli.py
```diff
@@ -157,7 +157,11 @@ def run():
             logger.info('\nInterrupted by user')
         except RebaseHelperError as e:
             logger.error('\n%s', six.text_type(e))
+            logger.debug('\n', exc_info=1)
```
thozza: with RebaseHelperError I don't think we need to log the traceback. This type of exception is used as a way to express expected failures.

nforro:
@thozza: OK, sounds good, I'll make the changes.

thozza:
@nforro when I run rebase-helper without any arguments the end of the output contains:
```
No handlers could be found for logger "rebase-helper"
```
I think the issue is that the logger is not configured right at the beginning. This means we should do some basic configuration right at the beginning of ```run()``` method of ```CliHelper``` and then reconfigure the logger is necessary (e.g. when verbose mode is turned on using cli option).

Another thing is that the output is:
```
rebase-helper failed due to an unexpected error. Please report this problem
using the following link: https://github.com/phracek/rebase-helper/issues/new
and include the content of
rebase-helper-debug.log
file in the report. Thank you!
```
I think the path to the debug loq should be absolute.

The debug log file is created too late in ```Application``` constructor, so if some exception is raised too early, the file does not exist yet. I think it should be created right at the beginning of program execution. Also once the file is created, the path to debug log in the output is absolute.

If the debug log is created, the traceback is included in it, which is good. Also when I enable verbose mode, the traceback is included, which is good as well.

During the testing I found out that sources are downloaded already in the ```Applicartion``` constructor, which is IMHO strange (should be after I execute ```run()```, but this is not topic of this PR.

nforro:
@thozza: Thanks for your notes, I'll try to address them.

nforro:
All issues should be resolved.

The console logging handler is now initialized as the first thing after execution, and it is set to `DEBUG` loglevel until `rebase-helper-debug.log` is created to ensure tracebacks are shown in case they can't be written to the log file.

rebasehelper/cli.py
```diff
@@ -148,15 +159,103 @@ def __getattr__(self, name):
...
+                                          report_log_file,
+                                          None,
+                                          logging.INFO)
+        except (IOError, OSError):
+            logger.warning("Can not create report log '%s'", report_log_file)
+        else:
+            return report_log_file
+
+    @staticmethod
+    def setup(arguments):
```
thozza: arguments should default to ```None```. Since ArgumentParser.parse() can be called without any argument or with None - https://docs.python.org/2/library/argparse.html#the-parse-args-method

rebasehelper/cli.py
```diff
@@ -148,15 +159,103 @@ def __getattr__(self, name):
...
+
+        return cli, execution_dir, debug_log_file, report_log_file
+
+    @staticmethod
     def run():
         try:
-            cli = CLI(sys.argv[1:])
-            app = Application(cli)
+            debug_log_file = None
+            cli, execution_dir, debug_log_file, report_log_file = CliHelper.setup(sys.argv[1:])
```
thozza: We don't need to pass the ```sys.argv``` to the parser. The possibility to pass the arguments explicitly is more interesting for testing.

nforro:
@thozza: After thinking about it I realized I don't like the idea of moving those `add_*_log_file` methods to `LoggerHelper`, since they are just wrappers around `LoggerHelper.add_file_handler`.
I moved them back to `Application` class instead.

nforro:
@thozza: I've modified the code as we discussed. Can you please take a look at it?

rebasehelper/constants.py
```diff
@@ -21,3 +21,4 @@
 #          Tomas Hozza <thozza@redhat.com>
 
 PROGRAM_DESCRIPTION = "OS package rebase-helper"
+NEW_ISSUE_LINK = 'https://github.com/phracek/rebase-helper/issues/new'
```
thozza: @nforro we moved the repo, so thie URL should be fixed

nforro: @thozza: Done.

thozza:
LGTM. Thanks @nforro !

PR #195, opened 2016-06-10
Remove local file if KeyboardInterrupt is raised during the download. (fixes #189)

### LINKED ISSUE #189: KeyboardException in DownloadingHelper function
I guess, we should catch KeyboardException in case user press Ctrl+C during the downloading sources.

If this happens, we should delete downloaded file and raise KeyboardException again.

thozza:
Remove local file if KeyboardInterrupt is raised during the download. Test is included.


nforro:
Looks good to me.

Issue #139, opened 2015-11-27
rebase-helper should support JSON output format


phracek:
Nowadays we are supporting test based format.
For future it would be good to support JSON format for later usage.

phracek:
@uhliarik Do you have a time to solve this issue?
Or are you busy with the network stuff?

I guess, either Nikola or me can look on it.

phracek:
Fixed by Pull Request #202  .

PR #202, opened 2016-06-22
JSON support.

phracek:

This Pull Request introduce JSON support.
Reference issue #139

### DIFF
--- a/rebasehelper/output_tool.py
+++ b/rebasehelper/output_tool.py
@@ -23,6 +23,7 @@
 import sys
 import os
 import six
+import json
 
 from rebasehelper.exceptions import RebaseHelperError
 from rebasehelper.logger import LoggerHelper, logger, logger_report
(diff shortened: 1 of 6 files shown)


rebasehelper/settings.py
```diff
@@ -50,3 +50,5 @@
 
 ABIDIFF_ERROR = 1
 ABIDIFF_USAGE_ERROR = 2
+
+REBASE_HELPER_OUTPUT_PREFIX = ".txt"
```
nforro: Shouldn't this be called `REBASE_HELPER_OUTPUT_SUFFIX`?

rebasehelper/output_tool.py
```diff
@@ -175,6 +176,31 @@ def print_pkgdiff_tool(cls):
                             logger_report.info("%s See for more details %s", output, checker)
 
 
+@register_output_tool
+class JSONOutputTool(BaseOutputTool):
+
+    """ Text output tool. """
```
nforro: Here it should be `""" JSON output tool """`.

rebasehelper/tests/test_output_tool.py
```diff
@@ -99,6 +168,19 @@ def test_text_output(self):
...
+    def test_json_output(self):
+        output = OutputTool('json')
+        data = self.get_data()
+
+        logfile = os.path.join(self.TESTS_DIR, REBASE_HELPER_RESULTS_LOG)
+        output.print_information(logfile)
+
+        with open(logfile) as f:
+            lines = [x.strip().replace("'", "") for x in f.readlines()]
+            assert '\n'.join(lines) == self.get_expected_json_output()
```
nforro: I think it might be better to have the expected output in a dict, JSON from the log file could be loaded to another dict and then simple dict comparison could be made.

rebasehelper/settings.py
phracek: Sure I will update it.

rebasehelper/tests/test_output_tool.py
phracek: You are right.

PR #203, opened 2016-06-24
Introduce --build-retries option

### LINKED ISSUE #152: Number of retries on failed builds should be configurable + reties not done in non-interactive mode
When building RPMs, rebase-helper can inspect the build logs in case it failed. If it finds some common error, it can fix the SPEC and will try to rebuild the package. AFAIK this does not work in non-interactive mode.

I think **the number of retries could be configurable** with something like `--build-retries` (and should default to some reasonably small number e.g. 2) and the **retries should be done also in non-interactive mode**.

This is a bug and RFE at the same time. The configuration of retries is a RFE. The fact that the retries are not done in non-interactive mode is a bug in my opinion.

phracek:

This Pull Request fixes #152 
Please review it.
It includes also tests

### DIFF
--- a/man/rebase-helper.1
+++ b/man/rebase-helper.1
@@ -50,5 +50,10 @@ output is more verbose
 \fB\-w\fR, \fB\-\-keep\-workspace\fR
 select the choice if you want to keep workspace directory after finishing
 
+.TP
+\fB\-\-build\--retries\fR=\fIBUILD_RETRIES\fR
+specifies number of retries rebase-helper will do in case of failed build.
+If build fails, rebase-helper analyzes the build log, tries to fix the issue if possible and retries the build.
+
 .SH AUTHOR
 Written by Petr Hracek and Tomas Hozza
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -440,8 +440,9 @@ class Application(object):
             build_dict['builds_nowait'] = self.conf.builds_nowait
             build_dict['build_tasks'] = self.conf.build_tasks
 
-            failed_before = False
-            while True:
+            files = {}
+            number_retries = 0
+            while self.conf.build_retries != number_retries:
                 try:
                     if self.conf.build_tasks is None:
                         build_dict.update(builder.build(spec, sources, patches, results_dir, **build_dict))
@@ -477,6 +478,7 @@ class Application(object):
                                     if status not in ['succeeded', 'skipped']:
                                         logger.info('Copr build {} did not complete successfully'.format(build_id))
                                         return False
+                    # Build finishes properly. Go out from while cycle
                     OutputLogger.set_build_data(version, build_dict)
                     break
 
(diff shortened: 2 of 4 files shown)


man/rebase-helper.1
```diff
@@ -50,5 +50,8 @@ output is more verbose
 \fB\-w\fR, \fB\-\-keep\-workspace\fR
 select the choice if you want to keep workspace directory after finishing
 
+.TP
+\fB\-\-build\--tries\fR=\fIBUILD_RETRIES\fR
```
thozza: This should be named "build-retries".

man/rebase-helper.1
```diff
@@ -50,5 +50,8 @@ output is more verbose
...
+specify count of builds in case of failure. Rebase-helper tries to solve some build issues automatically.
```
thozza: I propose the following text or something similar: "specifies number of retries rebase-helper will do in case of failed build. If build fails, rebase-helper analyzes the build log, tries to fix the issue if possible and retries the build."

rebasehelper/application.py
```diff
@@ -440,8 +440,8 @@ def build_packages(self):
             build_dict['builds_nowait'] = self.conf.builds_nowait
             build_dict['build_tasks'] = self.conf.build_tasks
 
-            failed_before = False
-            while True:
+            files = {}
+            while self.conf.build_retries >= 0:
```
thozza: I don't think we should decrease the configuration value. Instead I propose to save the value in a different variable and decrease it there or just store number of retries in a separate variable and check the value against the configuration value. The possible problem with current implementation is, that if rebase-helper retries builds of OLD RPMs, then it will not do the same number of retires for NEW RPMs. In a corner case the building of NEW RPMs will not start at all if the value is decreased to -1.

rebasehelper/application.py
```diff
@@ -517,24 +519,25 @@ def build_packages(self):
...
+                if os.path.exists(os.path.join(results_dir, 'SRPM')):
                     shutil.rmtree(os.path.join(results_dir, 'SRPM'))
-                    return False
-                #  build just failed, otherwise we would break out of the while loop
-                failed_before = True
-
-                shutil.rmtree(os.path.join(results_dir, 'RPM'))
-                shutil.rmtree(os.path.join(results_dir, 'SRPM'))
+                self.conf.build_retries -= 1
+        if self.conf.build_retries == 0:
```
thozza: If the build fails also after configured number of retries, we should raise RebaseHelperError. I don't see any reason why returning a value from this function would be beneficial or necessary. Can you please explain?

rebasehelper/application.py
```diff
@@ -706,6 +709,8 @@ def run(self):
                         elif self.conf.buildtool == 'copr':
                             self.print_copr_logs()
                         return 0
+                    if not build:
+                        return 1
```
thozza: This should not be needed, since if building fails, the RebaseHelperError should be raised instead.

thozza:
Please correct the name of the option in the commit message.

I'm also thinking if it would be possible to write a unit test using monkeypatching to test if the build is really retried exactly the number of times as configured.

Issue #147, opened 2016-02-05
Csmock as a checker


phracek:
I would like to discuss if covscan does make sense to be a checker in rebase-helper.

@kdudka: Is it already covscan packaged in Fedora?
How to use it? Do you have a documentation for it?

kdudka:
If you want to make sure that a rebase has not introduced any new defects, I would suggest to use csbuild. It is available in Fedora and a brief tutorial on how to use it can be found here:

https://kdudka.fedorapeople.org/devconf15.pdf

phracek:
This is close by PR #205

PR #205, opened 2016-06-30
Introduce csmock checker.

phracek:

#147

### DIFF
--- a/man/rebase-helper.1
+++ b/man/rebase-helper.1
@@ -27,7 +27,7 @@ Supported [mock (default), rpmbuild, fedpkg].
 .TP
 \fB\-\-pkgcomparetool\fR=\fIPKGCOMPARETOOL\fR
 select a tool which compares both RPM packages.
-Supported [pkgdiff, rpmdiff, abipkgdiff].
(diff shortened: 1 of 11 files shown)


phracek:
@nforro @thozza  Feel free to comment it.

nforro:
@phracek: Why is `CommandHelper` class needed? `subprocess.Popen` can search for an executable in `PATH`.

nforro:
@phracek: Actually my point was you can run "csmock" directly with `ProcessHelper.run_subprocess`, you don't have to pass absolute path. Maybe there is some reason why you want to use `CommandHelper.check_command`, but I don't see it.

phracek:
@nforro My point is only to detect whether csmock command exists on system. That's all.
I would like to predict unexpected failure.

nforro:
@phracek: I see. The same thing could be achieved by catching `FileNotFoundError` exception from `ProcessHelper.run_subprocess`, but perhaps your approach is better. The detection could be used also with other checkers, not just with csmock. Also csmock should be listed among other available checkers in command line argument help.

phracek:
@nforro You are right.

phracek:
@nforro Done by the last commit.

Issue #192, opened 2016-06-08
Problem with downloading file


phracek:
Archive is moved to different location.
~~~
phracek@unused-4-101: ~/work/Fedora/libpng ((50a7191...) %)$ rebase-helper 1.6.22
Results directory 'rebase-helper-results' exists, removing it

Failed to download file from URL ftp://ftp.simplesystems.org/pub/libpng/png/src/libpng16/libpng-1.6.21.tar.gz. Reason: '<urlopen error ftp error: [Errno ftp error] 550 libpng-1.6.21.tar.gz: No such file or directory>'
phracek@unused-4-101: ~/work/Fedora/libpng ((50a7191...) %)$ 
~~~

phracek:
How to reproduce this issue:
~~~
fedpkg clone libpng && cd libpng
fedpkg switch-branch f24
# This checkout is the latest upstream update related to 1.6.21
git check 50a719132b8fe34293c8ab4d58e503a35159680c
# update to 1.6.22 version
rebase-helper 1.6.22
~~~

File libpng-1.6.21.tar.gz exists in directory:
~~~
phracek@unused-4-101: ~/work/Fedora/libpng ((50a7191...) %)$ ls -la *.tar.gz
-rw-rw-r--. 1 phracek phracek 1419334 Jul 19 16:16 libpng-1.6.21.tar.gz
phracek@unused-4-101: ~/work/Fedora/libpng ((50a7191...) %)$ 
~~~

~~~
phracek@unused-4-101: ~/work/Fedora/libpng ((50a7191...) %)$ rebase-helper 1.6.22
Results directory 'rebase-helper-results' exists, removing it

Failed to download file from URL ftp://ftp.simplesystems.org/pub/libpng/png/src/libpng16/libpng-1.6.21.tar.gz. Reason: '<urlopen error ftp error: [Errno ftp error] 550 libpng-1.6.21.tar.gz: No such file or directory>'
phracek@unused-4-101: ~/work/Fedora/libpng ((50a7191...) %)$ 
~~~
I would suggest if tarball exists and downloading failes then use tarball which exists in the directory.

Jurisak:
After small investigation I found why we are twice downloading the first tarball. At first we are doing in _prepare_spec_objects on line 162. There is a creation of SpecFile and as a part of __init__ is downloading of tarball. Second case is in the same method on line 169. 

[...]

PR #208, opened 2016-07-20
Added new feature for tarball downloading

Jurisak:
Previously when you failed wo download tarball from Sources you
end up with exception. But I added easy check if tarball exists in
destination_source then it can proceed.

Fixes: #192

### DIFF
--- a/rebasehelper/specfile.py
+++ b/rebasehelper/specfile.py
@@ -170,12 +170,15 @@ class SpecFile(object):
             archive = [x for x in Archive.get_supported_archives() if src[0].endswith(x)]
             # if the source is a remote file, download it
             if archive:
-                if remote_files_re.search(src[0]) and self.download:
+                if remote_files_re.search(src[0]) and self.download and not os.path.isfile(abs_path):
+                    logger.debug("Tarball is not in absolute path {} "
+                                 "trying to download one from URL {}".format(abs_path, src[0]))
+                    logger.info("Tarball is not in absolute path. Trying to download it from URL")
                     try:
                         DownloadHelper.download_file(src[0], abs_path)
                     except DownloadError as e:
-                        raise RebaseHelperError("Failed to download file from URL {}. Reason: '{}'".format(src[0],
-                                                                                                           str(e)))
+                        raise RebaseHelperError("Failed to download file from URL {}. "
+                                                "Reason: '{}'. ".format(src[0], str(e)))
                 tar_sources.append(abs_path)
         return sources, tar_sources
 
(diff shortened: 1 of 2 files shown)


Jurisak:
On the other hand it seems to be a problem with Sources parsing. You can see my output after fix.

[jkulda@dhcp-24-102 libpng]$  rebase-helper 1.6.22 
Results directory 'rebase-helper-results' exists, removing it
Failed to download file from URL trying to find it in absolute path
Failed to download file from URL trying to find it in absolute path
Downloading file from URL ftp://ftp.simplesystems.org/pub/libpng/png/src/libpng16/libpng-1.6.22.tar.gz


As you can see on third attempt there is a valid URL :D...

Jurisak:
After small investigation I found why we are twice downloading the first tarball. At first we are doing in prepare_spec_objects on line 162. There is a creation of SpecFile and as a part of __init_ is downloading of tarball. Second case is in the same method on line 169.

On this line we are calling self.spec_file.copy(self.rebase_spec_file_path) but what it do is that it creates new SpecFile instance with different path :D ... So no copying at all. (We are only copying .spec file by using shutil.copy()). Is there some unknown dependency which blocks doing this same copying (using shutil) for other files?

phracek:
We have had a conversation with @thozza and I guess, we should firstly check if file exists and then rebase-helper will use it.
If source file does not exist then rebase-helper will download it.

Nowadays, rebase-helper downloads source tarball always, which is not needed. If user breaks downloading, then part of downloaded file is deleted. Which was implemented 3 month ago, I guess.

@Jurisak Could you please update Pull Request according to my suggestions?

Jurisak:
@phracek It's done as you wanted since yesterday ... Yesterday I talked with @thozza about it...

rebasehelper/specfile.py
```diff
@@ -170,12 +170,15 @@ def _get_initial_sources_list(self):
             archive = [x for x in Archive.get_supported_archives() if src[0].endswith(x)]
             # if the source is a remote file, download it
             if archive:
-                if remote_files_re.search(src[0]) and self.download:
+                if remote_files_re.search(src[0]) and self.download and not os.path.isfile(abs_path):
```
phracek: what about to use os.path.exists(abs_file)?

Jurisak: What is the main difference between isfile or exists?

Jurisak: From my perspective it depends on what we are searching for. If we are searching for a file then from my perspective is better to use isfile. Cause from the name we now what we are testing. 
Otherwise we should use exists. But this depends on you guys.

Issue #178, opened 2016-05-05
Allow passing additional arguments to mock build


msimacek:
In Java world we have a special mock plugin `pm_request` that can handle automatic detection of buildrequires. It would be nice if I had some way to enable it (i.e. pass `--enable-plugin pm_request`) when using rebase-helper

thozza:
Fixed by https://github.com/rebase-helper/rebase-helper/commit/d30d3ac2a595fa6825c2fae07157b8df038ca9cf

PR #209, opened 2016-07-22
Added support for mock plugins

Jurisak:
In past we have no option for mock plugins enablement.
Therefore I added simple fix to do it.

Fixes: #178

### DIFF
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -35,7 +35,7 @@ from rebasehelper.utils import PathHelper, RpmHelper, ConsoleHelper, GitHelper, 
 from rebasehelper.checker import Checker
 from rebasehelper.build_helper import Builder, SourcePackageBuildError, BinaryPackageBuildError, koji_builder
 from rebasehelper.patch_helper import Patcher
-from rebasehelper.exceptions import RebaseHelperError
+from rebasehelper.exceptions import RebaseHelperError, CheckerNotFoundError
 from rebasehelper.build_log_analyzer import BuildLogAnalyzer, BuildLogAnalyzerMissingError
 from rebasehelper.base_output import OutputLogger
 from rebasehelper.build_log_analyzer import BuildLogAnalyzerMakeError, BuildLogAnalyzerPatchError
@@ -564,8 +564,11 @@ class Application(object):
         checker = Checker(os.path.dirname(__file__))
         if not self.conf.pkgcomparetool:
             for check in checker.get_supported_tools():
-                results = checker.run_check(dir_name, checker_name=check)
-                pkgdiff_results[check] = results
+                try:
+                    results = checker.run_check(dir_name, checker_name=check)
+                    pkgdiff_results[check] = results
+                except CheckerNotFoundError:
+                    logger.info("Rebase-helper did not find checker '%s'." % check)
         else:
             pkgdiff_results[self.conf.pkgcomparetool] = checker.run_check(dir_name, checker_name=self.conf.pkgcomparetool)
         if pkgdiff_results:
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -410,7 +410,7 @@ class Application(object):
     def build_packages(self):
         """Function calls build class for building packages"""
         if self.conf.buildtool == 'fedpkg' and not koji_builder:
-            print ('Importing module koji failed. Switching to mockbuild.')
+            print('Importing module koji failed. Switching to mockbuild.')
             self.conf.buildtool = 'mock'
         try:
             builder = Builder(self.conf.buildtool)
@@ -439,6 +439,9 @@ class Application(object):
             results_dir = os.path.join(self.results_dir, version)
             build_dict['builds_nowait'] = self.conf.builds_nowait
             build_dict['build_tasks'] = self.conf.build_tasks
+            build_dict['enable_option'] = self.conf.enable_option
+            if self.conf.buildtool in ("copr", "fedpkg"):
+                logger.warning("We are not supporting option --enable-option for builder copr or fedpkg.")
 
             files = {}
             number_retries = 0
@@ -498,7 +501,7 @@ class Application(object):
                     build_log = 'build.log'
                     build_log_path = os.path.join(rpm_dir, build_log)
                     if version == 'old':
-                        raise RebaseHelperError('Building old RPM package failed. Check log %s', build_log_path)
+                        raise RebaseHelperError('Building old RPM package failed. Check log {}'.format(build_log_path))
                     logger.error('Building binary packages failed.')
                     msg = 'Building package failed'
                     try:
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -620,8 +620,9 @@ class Application(object):
     def print_summary(self):
         output_tool.check_output_argument(self.conf.outputtool)
         output = output_tool.OutputTool(self.conf.outputtool)
-        output.print_information(path=self._get_rebase_helper_log())
-        logger.info('Report file from rebase-helper is available here: %s', self.report_log_file)
+        report_file = os.path.join(self.results_dir, self.conf.outputtool + settings.REBASE_HELPER_OUTPUT_SUFFIX)
+        output.print_information(path=report_file)
+        logger.info('Report file from rebase-helper is available here: %s', report_file)
 
     def print_koji_logs(self):
         logs = self.get_new_build_logs()['build_ref']
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -407,6 +407,7 @@ class Application(object):
                 OutputLogger.set_patch_output('Unapplied patches:', self.rebased_patches['unapplied'])
         OutputLogger.set_patch_output('Patches:', self.rebased_patches)
 
+    @property
     def build_packages(self):
         """Function calls build class for building packages"""
         if self.conf.buildtool == 'fedpkg' and not koji_builder:
@@ -501,7 +502,12 @@ class Application(object):
                     build_log = 'build.log'
                     build_log_path = os.path.join(rpm_dir, build_log)
                     if version == 'old':
-                        raise RebaseHelperError('Building old RPM package failed. Check log {}'.format(build_log_path))
+                        error_message = 'Building old RPM package failed. Check log {}.\n'.format(build_log_path)
+                        if self.conf.enable_option is not None and self.conf.buildtool not in ('fedpkg', 'copr'):
+                            help_log = os.path.join(rpm_dir, "{}_output.log".format(self.conf.buildtool))
+                            error_message += "Possible problem with added option through --enable-plugin. " \
+                                             "See log {}".format(help_log)
+                        raise RebaseHelperError(error_message)
                     logger.error('Building binary packages failed.')
                     msg = 'Building package failed'
                     try:
@@ -711,7 +717,7 @@ class Application(object):
                     Application.check_build_requires(self.spec_file)
                 # Build packages
                 try:
-                    build = self.build_packages()
+                    build = self.build_packages
                     if self.conf.builds_nowait and not self.conf.build_tasks:
                         if self.conf.buildtool == 'fedpkg':
                             self.print_koji_logs()
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -440,7 +440,7 @@ class Application(object):
             build_dict['builds_nowait'] = self.conf.builds_nowait
             build_dict['build_tasks'] = self.conf.build_tasks
             build_dict['enable_option'] = self.conf.enable_option
-            if self.conf.buildtool in ("copr", "fedpkg"):
+            if self.conf.buildtool in ("copr", "fedpkg") and self.conf.enable_option is not None:
                 logger.warning("We are not supporting option --enable-option for builder copr or fedpkg.")
 
             files = {}
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -32,7 +32,7 @@ from rebasehelper.logger import logger, logger_report, LoggerHelper
 from rebasehelper import settings
 from rebasehelper import output_tool
 from rebasehelper.utils import PathHelper, RpmHelper, ConsoleHelper, GitHelper, KojiHelper, FileHelper, CoprHelper
-from rebasehelper.checker import Checker
+from rebasehelper.checker import checkers_runner
 from rebasehelper.build_helper import Builder, SourcePackageBuildError, BinaryPackageBuildError, koji_builder
 from rebasehelper.patch_helper import Patcher
 from rebasehelper.exceptions import RebaseHelperError, CheckerNotFoundError
@@ -59,7 +59,7 @@ class Application(object):
     rebased_patches = {}
     upstream_monitoring = False
 
-    def __init__(self, cli_conf=None):
+    def __init__(self, cli_conf, execution_dir, debug_log_file, report_log_file):
         """
         Initialize the application
 
@@ -69,17 +69,10 @@ class Application(object):
         OutputLogger.clear()
 
         self.conf = cli_conf
+        self.execution_dir = execution_dir
 
-        if self.conf.verbose:
-            LoggerHelper.add_stream_handler(logger, logging.DEBUG)
-        else:
-            LoggerHelper.add_stream_handler(logger, logging.INFO)
-
-        # The directory in which rebase-helper was executed
-        if self.conf.results_dir is None:
-            self.execution_dir = os.getcwd()
-        else:
-            self.execution_dir = self.conf.results_dir
+        self.debug_log_file = debug_log_file
+        self.report_log_file = report_log_file
 
         # Temporary workspace for Builder, checks, ...
         self.kwargs['workspace_dir'] = self.workspace_dir = os.path.join(self.execution_dir,
@@ -89,18 +82,9 @@ class Application(object):
                                                                      settings.REBASE_HELPER_RESULTS_DIR)
 
         self.kwargs['non_interactive'] = self.conf.non_interactive
-        # if not continuing, check the results dir
-        if not self.conf.cont and not self.conf.build_only and not self.conf.comparepkgs:
-            self._check_results_dir()
-        # This is used if user executes rebase-helper with --continue
-        # parameter even when directory does not exist
-        if not os.path.exists(self.results_dir):
-            os.makedirs(self.results_dir)
-            os.makedirs(os.path.join(self.results_dir, settings.REBASE_HELPER_LOGS))
 
-        self._add_debug_log_file()
-        self._add_report_log_file()
         logger.debug("Rebase-helper version: %s" % version.VERSION)
+
         if self.conf.build_tasks is None:
             self._get_spec_file()
             self._prepare_spec_objects()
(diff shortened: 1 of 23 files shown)


phracek:
@Jurisak Seems to be fine. But a question from my side. Let's say I would like to build a package by rpmbuild or fedpkg? 
Does it work properly?
Variable mock_plugin is passed into all builders, but only mock handle it?

phracek:
What about if mock_plugin_option is missing or enter multiple times?

thozza:
@Jurisak the idea was to introduce something like ```--builder-params``` or ```--builder-args``` which would just take string with anything the user wants. This string then would be passed to the builder object and appended to the command before executing. It is expected that the user knows which builder he uses and what arguments it can accept.

We explicitly don't want to maintain builder specific arguments with choices which can quickly go out of sync with the reality.

Jurisak:
@phracek:
"Seems to be fine. But a question from my side. Let's say I would like to build a package by rpmbuild or fedpkg?"
    There shouldn't be problem with rpmbuild or fedpkg cause both builder methods has **kwargs argument so unknown parameters are held there ... 

"Variable mock_plugin is passed into all builders, but only mock handle it?"
Yes as you can see on line 286 in build_helper.py. 

"What about if mock_plugin_option is missing or enter multiple times?"
This one is handled by argparse. But from what I know argparse will store the latest one ... Or the most right mock_plugin_option.

Jurisak:
@thozza Yes I told @phracek that it's very easy to make it for all builders not just for single one. But from the issue description I understood: "Do it only for mock" ... 

About the specific arguments. I wanted to prevent typos in plugin names ... So no build would be executed if there will be problem with arguments. I think it's easy to maintain those lists but ... If you prefer just letting it on users then I will remove it...

Jurisak:
@thozza I searched for enabling plugins in rpmbuild or fedpgk but it seems that they don't have any ... Only mock has --enable-plugin argument ... What shall I do with them?

thozza:
@Jurisak it is in our best interest to implement more general solution than just what people are asking for in the tickets, because next month there may be another user asking for ability to pass any random argument to any other builder. The idea is not to handle the plugins use case anyhow differently. Just let the user pass any additional options to the builder and think that he know what he's doing. I would not care about typos in the string, we do not care at all. If the user passes string with additional arguments, it's his job to make sure it is accepted by the builder, otherwise it is expected that the build will fail.

Jurisak:
@thozza Yes I totally agree but as I said the first commit was only though how it could work in the meaning that i know how to make it more general ... 

So if I understood it clearly it can be used as builder specific option enabler ... Cause for fedpkg and rpmbuild there is no specific option for enabling plugins so I could use it to pass options?

thozza:
@Jurisak 
* rename the option.
* make it to accept arbitrary string without further checking.
* pass the string to all builders and append it to the command we already run.

Jurisak:
@thozza Ok so no support for mock --enable-plugin? 
Take it as it is and if user wants to enable plugin must do --add-option "--enable-plugin=pm_request"? 
Also I would like to allow multiple options. Can I do that or do it only for single option?

P.S.: Just thought for myself... Didn't we moved from plugins to options?

Jurisak:
@thozza Ok so no support for mock --enable-plugin? 
Take it as it is and if user wants to enable plugin must do --add-option "--enable-plugin=pm_request"? 
Also I would like to allow multiple options. Can I do that or do it only for single option?

P.S.: Just thought for myself... Didn't we moved from plugins to options?

Jurisak:
Guys I found problem with option parsing.
When user types:
rebase-helper 1.6.22  --enable-option "-v" 
it got:
rebase-helper: error: argument --enable-option: expected at least one argument

No matter what kind of option you will pass it. In  "--asd"  "--wtf" same result.

After speaking with ttomecek I concider to make it really specific. Cause when user runs:
rebase-helper 1.6.22  --enable-option="-v" 
All went well ... So I suggest to write into docs how to use this option and that there must be = to work it correctly ... 

How do you see it guys?

rebasehelper/application.py
```diff
@@ -439,6 +440,9 @@ def build_packages(self):
             results_dir = os.path.join(self.results_dir, version)
             build_dict['builds_nowait'] = self.conf.builds_nowait
             build_dict['build_tasks'] = self.conf.build_tasks
+            build_dict['enable_option'] = self.conf.enable_option
+            if self.conf.buildtool in ("copr", "fedpkg"):
+                logger.warning("We are not supporting option --enable-option for builder copr or fedpkg.")
```
phracek: Personally, I guess, this can be logged as debug. User can be suppressed. But it is my opinion.

rebasehelper/application.py
```diff
@@ -708,7 +717,7 @@ def run(self):
                     Application.check_build_requires(self.spec_file)
                 # Build packages
                 try:
-                    build = self.build_packages()
+                    build = self.build_packages
```
phracek: Why did you delete brackets? Is it working? I did not know, brackets are not needed? Is this python 3 specific?

rebasehelper/build_helper.py
```diff
@@ -327,19 +329,21 @@ def build(cls, spec, sources, patches, results_dir, root=None, arch=None, **kwar
...
-            rpms = cls._build_rpm(srpm, tmp_results_dir)
+            rpms = cls._build_rpm(srpm, tmp_results_dir, **kwargs)
             # remove SRPM - side product of building RPM
             tmp_srpm = PathHelper.find_first_file(tmp_results_dir, "*.src.rpm")
             if tmp_srpm is not None:
                 os.unlink(tmp_srpm)
 
         if rpms is None:
             # We need to be inform what directory to analyze and what spec file failed
+            print(rpm_results_dir)
```
phracek: Remove print and log it into debug.log

rebasehelper/cli.py
```diff
@@ -31,13 +31,20 @@
 from rebasehelper.exceptions import RebaseHelperError
 
 
+class EnableOptionFormatException(Exception):
```
thozza: Please remove this exception and use RebaseHelperError for critical exceptions.

rebasehelper/cli.py
```diff
@@ -143,6 +150,13 @@ def add_args(self):
             help="Specify number of retries in case build fails.",
             type=int
         )
+        self.parser.add_argument(
+            "--enable-option",
```
thozza: please rename it to indicate that this is intended only for extra arguments for builders. I already provided you with some examples.

rebasehelper/cli.py
```diff
@@ -143,6 +150,13 @@ def add_args(self):
...
+            dest="enable_option",
```
thozza: this argument should not be needed, the value should be placed into variable named the same way by default.

rebasehelper/cli.py
```diff
@@ -164,6 +178,9 @@ def run():
         except RebaseHelperError as e:
             logger.error('\n%s', six.text_type(e))
             sys.exit(1)
+        except EnableOptionFormatException as e:
+            sys.stderr.write("{}".format(e))
+            sys.exit(1)
```
thozza: Please remove these three lines, as the new exception is not needed. Also in the future please be consistent with the way we log things. Just have a look at the lines above your addition.

rebasehelper/application.py
thozza: Why this is even displayed? It will be displayed every time when user uses copr or fedpkg without even using the new option. This should be completely removed.

rebasehelper/application.py
thozza: This is for sure a typo. the build_packages function is not called, but instead stored in the build variable. Please fix it.

rebasehelper/application.py
```diff
@@ -407,10 +407,11 @@ def patch_sources(self, sources):
                 OutputLogger.set_patch_output('Unapplied patches:', self.rebased_patches['unapplied'])
         OutputLogger.set_patch_output('Patches:', self.rebased_patches)
 
+    @property
```
thozza: This should not be a property, please make it a function again.

rebasehelper/application.py
```diff
@@ -407,10 +407,11 @@ def patch_sources(self, sources):
...
     def build_packages(self):
         """Function calls build class for building packages"""
         if self.conf.buildtool == 'fedpkg' and not koji_builder:
-            print ('Importing module koji failed. Switching to mockbuild.')
+            print('Importing module koji failed. Switching to mockbuild.')
```
thozza: This should be modified to logging call instead of using print().

rebasehelper/application.py
thozza: So this is because you changed the function to a property, which actually does not make any sense. Please revert it.

rebasehelper/build_helper.py
```diff
@@ -327,19 +329,21 @@ def build(cls, spec, sources, patches, results_dir, root=None, arch=None, **kwar
         # build SRPM
         srpm, cls.logs = cls._build_srpm(spec, sources, patches, results_dir)
 
+
```
thozza: please remove the extra empty line

rebasehelper/build_helper.py
```diff
@@ -283,7 +283,7 @@ class MockBuildTool(BuildToolBase):
     logs = []
 
     @classmethod
-    def _build_rpm(cls, srpm, results_dir, root=None, arch=None):
+    def _build_rpm(cls, srpm, results_dir, root=None, arch=None, enable_option=None, **kwargs):
```
thozza: please remove the ```**kwargs``` as it should not be needed at all.

rebasehelper/build_helper.py
```diff
@@ -327,19 +329,21 @@ def build(cls, spec, sources, patches, results_dir, root=None, arch=None, **kwar
...
         srpm, cls.logs = cls._build_srpm(spec, sources, patches, results_dir)
 
+
         # build RPMs
         rpm_results_dir = os.path.join(results_dir, "RPM")
         with MockTemporaryEnvironment(sources, patches, spec, rpm_results_dir) as tmp_env:
             env = tmp_env.env()
             tmp_results_dir = env.get(MockTemporaryEnvironment.TEMPDIR_RESULTS)
-            rpms = cls._build_rpm(srpm, tmp_results_dir)
+            rpms = cls._build_rpm(srpm, tmp_results_dir, **kwargs)
```
thozza: Why is there ```**kwargs```?

rebasehelper/build_helper.py
```diff
@@ -371,7 +375,7 @@ class RpmbuildBuildTool(BuildToolBase):
     logs = []
 
     @classmethod
-    def _build_rpm(cls, srpm, workdir, results_dir):
+    def _build_rpm(cls, srpm, workdir, results_dir, enable_option=None, **kwargs):
```
thozza: please remove the ```**kwargs``` as it should not be needed at all.

rebasehelper/build_helper.py
```diff
@@ -425,7 +431,7 @@ def build(cls, spec, sources, patches, results_dir, **kwargs):
             env = tmp_env.env()
             tmp_dir = tmp_env.path()
             tmp_results_dir = env.get(RpmbuildTemporaryEnvironment.TEMPDIR_RESULTS)
-            rpms = cls._build_rpm(srpm, tmp_dir, tmp_results_dir)
+            rpms = cls._build_rpm(srpm, tmp_dir, tmp_results_dir, **kwargs)
```
thozza: Why is there ```**kwargs```?

rebasehelper/build_helper.py
```diff
@@ -386,6 +390,8 @@ def _build_rpm(cls, srpm, workdir, results_dir):
         output = os.path.join(results_dir, "build.log")
 
         cmd = [cls.CMD, '--rebuild', srpm]
+        if enable_option is not None:
+            cmd.extend(filter(None, enable_option.split(" ")))
```
thozza: The same code is also in Rpmbuild build tool class. Maybe it would be a good idea to extract the actual code to the BuildToolBase class and call it from there. Or create new common base class for local builders, since these are the only ones for which it makes sense.

thozza:
@Jurisak please rebase and merge the commits, so that there are only commits with the actual changes which are currently relevant.

rebasehelper/application.py
Jurisak: Agree I was only disturbed by the whitespace.

rebasehelper/application.py
Jurisak: I didn't make it intentionally as a property. I assume that I hit some combination of keys and my IDE did this behind my back. I'm very sorry for that. It's my first case...

rebasehelper/application.py
Jurisak: Well i think it's better to use warning cause it should inform you that this is not working on other builders and yes there is missing and with checking enable_option ... 

For me it's good. User forgets that enable_option cant be used for copr and fedpkg and he should be informed. If you want it through debug. I can do it through debug...

rebasehelper/application.py
Jurisak: It's done by my IDE. Sorry for that. I'll revert it.

rebasehelper/build_helper.py
Jurisak: This shouldn't be here at all...

rebasehelper/cli.py
Jurisak: --enable-extra-builder-options ? or --enable-builder-options ? Or you have any other suggestions?

rebasehelper/cli.py
Jurisak: Correct

rebasehelper/cli.py
Jurisak: Ok.

rebasehelper/cli.py
Jurisak: Well it's kinda is ... I wanted to stop using extra builder option in wrong format so I raised RebaseHelperError and I end up with no message at all. :D After some investigation I found that we are creating logger in Application.__init__.py so EVERY exception in CLI will not use logger cause it's created after argument parsing. 

I wanted to discuss with you this problem but you were not responding.

rebasehelper/build_helper.py
Jurisak: Yep.

rebasehelper/build_helper.py
Jurisak: Well my though was this. Method build is called with dictionary which represents some of method parameters. My though was that in future you will forget to add enable_option into this dict and you will end up with error. 

I can remove it and check it before the call but my second though was that in future we may use **kwargs in _build_rpm method.

rebasehelper/build_helper.py
Jurisak: In **kwargs is dict which contains enable_option parameter. So that why.

rebasehelper/build_helper.py
Jurisak: Well my though was this. Method build is called with dictionary which represents some of method parameters. My though was that in future you will forget to add enable_option into this dict and you will end up with error.

I can remove it and check it before the call but my second though was that in future we may use **kwargs in _build_rpm method.

rebasehelper/build_helper.py
Jurisak: In **kwargs is dict which contains enable_option parameter. So that why.

rebasehelper/build_helper.py
Jurisak: I will think about it.

Jurisak:
@thozza Could you please look at my comments. I'm not 100% sure about some. Thanks :)

rebasehelper/application.py
thozza: When user uses copr/fedpkg it does not make sense to inform them about some option they didn't even specified on command line.

The only situation when it would make sense it when they explicitly use copr/fedpkg with the option.

rebasehelper/build_helper.py
thozza: You won't because the option has default value. Getting error which we can fix is better than masking it. For new code explicit parameters are preferred, unless there is strong reason to take arbitrary kwargs. If there will be such situation in the future, we will add it then, now it is just dead code.

rebasehelper/build_helper.py
thozza: Then  please pass it explicitly instead of the whole dictionary, it is more readable.

rebasehelper/build_helper.py
thozza: The same as above...

rebasehelper/build_helper.py
thozza: Same as above...

rebasehelper/cli.py
thozza: I that case please think about moving the check and raising the exception after the parsing of arguments. Also the logging is little bit reworked in another PR which we will be merging shortly and it may help you with the problem.

rebasehelper/cli.py
thozza: @Jurisak So my initial ideas were ```--builder-params``` or ```--builder-args```. IMHO there is no need to use "enable" in the option name.

rebasehelper/build_helper.py
Jurisak: OK I admit that it is really better to have it explicit than implicit ... Sorry for that.

rebasehelper/application.py
Jurisak: Well this one is really strange. In my code and in my local commits I have it without property? Does anybody know how the hell the property came from?

rebasehelper/cli.py
Jurisak: Ok. I agree with the --builder at the beginning but I would rather use `--builder-option` cause it logically connects with the builder option rather than with params or args ... How do you see it?

rebasehelper/build_helper.py
```diff
@@ -272,6 +272,14 @@ def _build_srpm(cls, spec, sources, patches, results_dir):
 
         return srpm, logs
 
+    @classmethod
```
thozza: this could be static

rebasehelper/build_helper.py
```diff
@@ -272,6 +272,14 @@ def _build_srpm(cls, spec, sources, patches, results_dir):
...
+    def get_builder_options(cls, **kwargs):
+        if 'builder_options' not in kwargs:
+            return None
+        if kwargs['builder_options'] is not None:
```
thozza: This could be simplified to:
```python
builder_options = kwargs.get('builder_options', None)
if builder_options is not None:
    ...
```

rebasehelper/cli.py
```diff
@@ -163,8 +174,8 @@ class CliHelper(object):
 
     @staticmethod
     def run():
+        debug_log_file = None
         try:
-            debug_log_file = None
```
thozza: why did you do this change?

rebasehelper/build_helper.py
Jurisak: Yes it could but I make it as a classmethod cause rest of the method are classmethods as well...

rebasehelper/build_helper.py
Jurisak: I never know about this. I like it :).

rebasehelper/cli.py
Jurisak: I think catching exception is only for code which can raise an exception. Also you are using `debug_log_file` in one of the exceptions and my IDE marked it that could be referenced before assignment. I know that it's on the first line BUT still I think try: catch: clause should contain code which could raise an Exception and `debug_log_file = None` will not ..

rebasehelper/cli.py
```diff
@@ -163,8 +174,8 @@ class CliHelper(object):
...
             # be verbose until debug_log_file is created
             handler = LoggerHelper.add_stream_handler(logger, logging.DEBUG)
             cli = CLI()
```
Jurisak: I think it must be CLI(sys.argv[1:]) ... I fixed in next commit.

rebasehelper/cli.py
nforro: I think this change is ok, but it is outside the scope of this pull request.

rebasehelper/cli.py
nforro: This is intentional, since `sys.argv[1:]` is what `ArgumentParser.parse_args` uses if not provided with any arguments.

rebasehelper/cli.py
Jurisak: Agree there is more refactorization ... I'll try to avoid this next time.

rebasehelper/cli.py
Jurisak: Ok. According to this i need to do an update.

thozza:
closing in favor of PR #216 so that we can do a new release

thozza:
@Jurisak thank you for the contribution.

Issue #190, opened 2016-06-01
Rework Checker class


thozza:
I think the class should be implemented in a way, that it gets the plugins from some default place when instantiated. Having to pass the dir name every time does not seem good to me. It may be also worth to instantiate the class on module import, so that we don't have to load all checker every time we instantiate the class on multiple places.

https://github.com/phracek/rebase-helper/commit/aefb8eee94802c171f288d353b4146a9d7c31197

phracek:
Fixed by PR https://github.com/rebase-helper/rebase-helper/pull/213

PR #213, opened 2016-07-26
Refactoring of Checker class and couple of other cleanups

thozza:
* this PR resolves issue #190 
* additional clean up, refactoring was done with the intention to increase the readability of code
* no functionality change is expected

### DIFF
--- a/rebasehelper/checker.py
+++ b/rebasehelper/checker.py
@@ -101,17 +101,29 @@ class CheckersRunner(object):
                     print ("Module '%s' does not implement `register(context)`" % modname)
         return plugin_checkers
 
-    def run_check(self, results_dir, checker_name=None):
-        """Run the check"""
-        _tool = None
+    def run_checker(self, results_dir, checker_name):
+        """
+        Runs a particular checker and returns the results.
+
+        :param results_dir: Path to a directory in which the checker should store the results.
+        :type results_dir: str
+        :param checker_name: Name of the checker to run. Ideally this should be name of existing checker.
+        :type checker_name: str
+        :raises NotImplementedError: If checker with the given name does not exist.
+        :return: results from the checker
+        """
+        checker = None
         for check_tool in six.itervalues(self.plugin_classes):
             if check_tool.match(checker_name):
-                _tool = checker_name
+                # we found the checker we are looking for
+                checker = check_tool
+                break
 
-        if _tool is None:
-            raise NotImplementedError("Unsupported checking tool")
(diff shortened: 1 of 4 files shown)


rebasehelper/checker.py
```diff
@@ -52,29 +50,40 @@ def run_check(cls, results_dir):
...
 
-class Checker(object):
+class CheckersRunner(object):
     """
-    Class representing a process of checking final packages.
+    Class representing the process of running various checkers on final packages.
     """
 
-    def __init__(self, dir_name):
+    def __init__(self, dir_name=os.path.join(os.path.dirname(__file__), 'checkers')):
```
phracek: Does it make sense to hard coded dir_name? Do we need to change it?

rebasehelper/checker.py
```diff
@@ -52,29 +50,40 @@ def run_check(cls, results_dir):
...
+    def _checker_find_injector(self, module):
         injectors = []
         for n in dir(module):
             attr = getattr(module, n)
             if hasattr(attr, '__base__') and attr.__base__.__name__ == self._injector_type:
                 injectors.append(attr)
         return injectors
 
-    def load_checkers(self, dir_name):
+    def load_checkers(self, plugins_dir):
```
phracek: load_checkers can be also hidden, I guess.

rebasehelper/checker.py
```diff
@@ -52,29 +50,40 @@ def run_check(cls, results_dir):
...
+        Load checker implementations from the given location.
+
+        :param plugins_dir: Path to directory from which to load the checkers.
+        :type plugins_dir: str
+        :return: Dictionary with names of the checkers and the actual checker objects.
+        :rtype: dict
+        """
         plugin_checkers = {}
-        plugins_dir = os.path.join(dir_name, 'checkers')
         for plugin in os.listdir(plugins_dir):
```
phracek: plugins_dir is not defined. I think it should be mentioned in __init__ function.

rebasehelper/checker.py
thozza: Yes, it could be, however one can call the function to load the checkers again, possibly from different directory.

rebasehelper/checker.py
thozza: You can always pass the specific path in which the checkers are placed. This is just to simplify the default case, which works. If we move the dir in our repo, we will just edit this one line. No complication from my point of view.

phracek:
We can merge it.

Issue #144, opened 2016-01-15
The version for log4cplus "1.1.3-rc6" is not parsed correctly


thozza:
I'm creating this report so that we don't forget about it.
The version and pre-version extraction for log4cplus does not work correctly. The code for extracting the version could be improved to better reuse the SOURCE0 string for extraction regexp.

Output of rebase-helper:
```
thozza@thozza-pc ~/devel/fedora/log4cplus (master %>)
$ rebase-helper 1.1.3-rc6
Git-rebase operation to log4cplus-1.1.3-rc3 is ongoing...
Building packages for log4cplus version 1.1.3
Building SRPM
Building SRPM finished successfully
Building RPMs
^C
Interrupted by user
```

You can see that the "rc6" is not extracted. Also it may be worth indicating that basically part of the passed version was ignored and it may not make sense to continue with the rebase.

Output in verbose mode:
```
thozza@thozza-pc ~/devel/fedora/log4cplus (master %>)
$ rebase-helper -v 1.1.3-rc6
Results directory 'rebase-helper-results' exists, removing it
Updating the extra version
Extracting version from 'log4cplus-1.1.3-rc3.tar.xz' using '%{name}-%{version}%{?prever:-%{prever}}.tar.xz'
Extracting version using regex '%{name}-([.0-9]+\w*)%{?prever:-%{prever}}.tar.xz'
Failed to extract version from archive name!
Trying to extracting version using fallback regex '^\w+-?_?v?([.0-9]+\w*)(.tgz|.tar.gz|.gem|.bz2|.tar.xz|.zip)'
Failed to extract version from archive name using fallback regex!
Splitting string ''
Updating the extra version
Extracting version from 'log4cplus-1.1.3-rc3.tar.xz' using '%{name}-%{version}%{?prever:-%{prever}}.tar.xz'
Extracting version using regex '%{name}-([.0-9]+\w*)%{?prever:-%{prever}}.tar.xz'
Failed to extract version from archive name!
Trying to extracting version using fallback regex '^\w+-?_?v?([.0-9]+\w*)(.tgz|.tar.gz|.gem|.bz2|.tar.xz|.zip)'
Failed to extract version from archive name using fallback regex!
Splitting string ''
argument passed as a new source is a version
Splitting string '1.1.3-rc6'
Divided version '1.1.3' and extra string ''
Updating version in SPEC from '1.1.3' with '1.1.3'
[...]

PR #215, opened 2016-07-28
Fixed parsing of extra version from some archive names and simplified the way we replace Source0 in new SPEC if needed (resolves issue #144)

thozza:
* Fixed parsing of extra version from some archive names
* Simplified the way we replace Source0 in new SPEC if needed
* Fixed the way we construct Release string when extra version is needed. Now it complies to Fedora Packaging guidelines.
* Resolves issue #144 

It just works :laughing:


phracek:
Seems to be fine. A pretty complicated commit.

PR #218, opened 2016-08-03
Check the validity of the configuration at the beginning of Application.run()

thozza:
Check the validity of the configuration at the beginning of Application.run()

Previously some checks were done late and some options were just ignored. If the combination of used options is not valid we fail now.

Signed-off-by: Tomas Hozza <thozza@redhat.com>


nforro:
LGTM

thozza:
@nforro thanks!

PR #221, opened 2016-08-08
 Refactored OutputLogger to be ResultsStore

thozza:
Previously the global Class object has been used for storing results,
which was messy and also another Class has been used internally, which
was basically a dictionary. Now the ResultsStore can be instantiated
and the results are stored inside the instance. This is better for
testing and makes more sense. The loggers also relied on the global
Class object, but now rather take the ResultsStore object instance and
thake the results from it. We are creating a global instance of
ResultsStore class, so that it can be used internally as it was before.
This global instance is the default argument for OutputTools.

This is preparation for reworking OutputTools so that the content makes
sense, estecially for computer readable formats like JSON.


nforro:
LGTM

thozza:
Thanks @nforro

Issue #186, opened 2016-05-26
Source0: needs to bu updated for thousands of packages.


torsava:
Hi!
The Python Package Index (PyPI) has decided to change the URL address to download tarballs/sources [0]. This means that possibly thousands of Python packages need to be updated with a new Source0 link.

The new URL address is:

```
https://files.pythonhosted.org/packages/source/p/positional/positional-1.1.0.tar.gz
```

Instead of the old:

```
https://pypi.python.org/ ...
```

**Do you think it would be beneficial to work this change into the `rebase-helper`? If so, I'd volunteer for the effort.**

[0] https://bitbucket.org/pypa/pypi/issues/438/backwards-compatible-un-hashed-package

Great day to all,
Tomas

thozza:
Thank you for your interesting idea.

rebase-helper uses the Source0 from the SPEC for downloading new sources. If the URL does not exist, it would fail. In such case we could definitely try another URL if we find out the package is from PyPI. And if it works also use it in the rebased SPEC.

Having more general way of doing the changes is preferred. An idea from the top of my head would be to add some pre-download hook which can then run plugins, which modify the Source0 URL before download. Then we could have the change of URL of PyPI implemented as such pre-download plugin. However I think we will have to discuss the right approach before implementing anything.

@torsava do you know what is the timeline for PyPI changes?

torsava:
That sounds like a very good system!

> @torsava do you know what is the timeline for PyPI changes?

Well, they switched to the new system a month ago, the old system no longer works and they gave no advanced warning what so ever (claiming the URL format was not an officially supported API). So much for a timeline. :)

phracek:
Link to relevant code https://github.com/phracek/rebase-helper/blob/master/rebasehelper/specfile.py#L163

PR #223, opened 2016-08-16
Implemented URL transformations for downloading sources

torsava:
This PR adds URL transformations for downloading sources, as was discussed in issue #186 .

However, URLs are only transformed when downloading sources. The transformed URLs are NOT saved in the new spec file. Should I try to implement this as well? It will be a bit more complex.

### DIFF
--- a/rebasehelper/settings.py
+++ b/rebasehelper/settings.py
@@ -35,6 +35,10 @@ NEW_SOURCES = 'new_sources'
 OLD_SOURCES_DIR = os.path.join(REBASE_HELPER_WORKSPACE_DIR, OLD_SOURCES)
 NEW_SOURCES_DIR = os.path.join(REBASE_HELPER_WORKSPACE_DIR, NEW_SOURCES)
 
+SOURCES_URL_TRANSFORMATIONS = [
+        ('^https?://pypi.python.org/', 'https://files.pythonhosted.org/'),
+        ]
+
 # The variable for access to full information about patches
 FULL_PATCHES = 'patches_full'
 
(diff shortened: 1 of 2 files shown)


thozza:
@torsava Thanks for the PR. However I don't think we want such pypi specific code in the rebase-helper. As I states in the issue #186, more reasonable way would be to add interface so that you can have your own *.py plugin located somewhere and then it would be run. Personally I'm not in favor of the change as proposed as it adds code fixing specific thing which may change at any time.

torsava:
@thozza Hi! We've discussed this with @phracek (https://github.com/rebase-helper/rebase-helper/issues/186#issuecomment-227370191) and he agreed that a general URL transformation mechanism would be useful. So far it applies only to the PyPI URL change (affecting over 2000 packages), but I'm sure others will follow.

Thus I made the URL transformation general, so that more cases can be added to the SOURCES_URL_TRANSFORMATIONS list as needed.

thozza:
@torsava I admit that the change as proposed is the simplest way how to accomplish what you need, but I still don't think that the change as proposed is very generic and may result in even more use case specific code in the future if people will need to do any other change to SPEC files. Anyway if @phracek is willing to maintain the code for such specific use case and rework it in the future if needed, I won't block the merge.

torsava:
I understand. I'm not really all that keen on it, I just wanted to prevent the thousands of failed rebases in the new-hotness Bugzillas, giving a cool tool a bad name.

phracek:
@torsava Thanks for the Pull Request. I will remain it open. We will work on pre-hook system and your plugin will be a part of it. We will send you a comment to update PR soon. Around 2-3 weeks.

Your change does not affect the-new-hotness immediately, we have to release a update and update machine where the-new-hotness works. It needs a time around month or two.

What do you think?

torsava:
@phracek Cool, looking forward to it the pre-hook system!

skisela:
This functionality has been implemented through a pypi_url_fix hook which is run before downloading of sources specified in the Spec file.

Thanks for help!

https://github.com/rebase-helper/rebase-helper/pull/263/files

torsava:
Thank you for implementing it!

Issue #198, opened 2016-06-21
Download old sources from Fedora lookaside cache


nforro:
Old source tarballs can be downloaded from Fedora lookaside cache (`fedpkg sources`).

thozza:
Getting the idea further would be to download also the "old" RPMs from Koji and not build them at all.

phracek:
yeah, that's was a point already proposed by @ralphbean during DevConf2016.

PR #224, opened 2016-08-17
Download old sources from Fedora lookaside cache

nforro:
The `LookasideCacheHelper` class supports both Fedora and RHEL, but downloading from RHEL lookaside cache is functional only in Red Hat internal network. So I didn't enable it in rebase-helper for now. I also marked unit test setup for RHEL lookaside cache as *xfail*, but I can make it run only in case *pkgs.devel.redhat.com* domain is resolvable or remove it completely.

This PR partially resolves issue #198.

### DIFF
--- a/rebasehelper/specfile.py
+++ b/rebasehelper/specfile.py
@@ -30,7 +30,8 @@ import argparse
 import shlex
 from datetime import date
 
-from rebasehelper.utils import DownloadHelper, DownloadError, MacroHelper, defenc
+from rebasehelper.utils import DownloadHelper, DownloadError, MacroHelper
+from rebasehelper.utils import LookasideCacheHelper, LookasideCacheError, defenc
 from rebasehelper.logger import logger
 from rebasehelper import settings
 from rebasehelper.archive import Archive
@@ -143,6 +144,13 @@ class SpecFile(object):
         self.prep_section = self.spc.prep
         # HEADER of SPEC file
         self.hdr = self.spc.sourceHeader
+
+        try:
+            # try to download old sources from Fedora lookaside cache
+            LookasideCacheHelper.download('fedpkg', os.path.dirname(self.path), self.get_package_name())
+        except LookasideCacheError as e:
+            logger.debug("Downloading old sources from lookaside cache failed. "
+                         "Reason: '{}'.".format(str(e)))
 
         # All source file mentioned in SPEC file Source[0-9]*
         self.rpm_sections = self._split_sections()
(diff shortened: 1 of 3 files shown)


rebasehelper/tests/test_utils.py
```diff
@@ -629,3 +630,39 @@ def test_get_macros(self):
...
+            'url': 'http://pkgs.fedoraproject.org/repo/pkgs',
+        },
+        pytest.mark.xfail(run=False, reason='this works only inside Red Hat internal network')
+        ({
+            'tool': 'rhpkg',
+            'package': 'libtiff',
+            'filename': 'tiff-3.9.4.tar.gz',
+            'hashtype': 'md5',
+            'hash': '2006c1bdd12644dbf02956955175afd6',
+            'url': 'http://pkgs.devel.redhat.com/repo/pkgs',
```
thozza: I don't think the internal infrastructure URLs should be in public code. Please remove it.

rebasehelper/specfile.py
```diff
@@ -144,6 +145,12 @@ def _update_data(self):
         # HEADER of SPEC file
         self.hdr = self.spc.sourceHeader
 
+        try:
+            # try to download old sources from Fedora lookaside cache
+            LookasideCacheHelper.download('fedpkg', os.path.dirname(self.path), self.get_package_name())
+        except (LookasideCacheError, DownloadError):
```
thozza: I think raising only one type of exception from LookasideCacheHelper would be simpler, since we are not handling them differently.

nforro:
The *rhpkg* test setup has been removed and `LookasideCacheHelper` now throws only `LookasideCacheError` exception.

rebasehelper/specfile.py
```diff
@@ -144,6 +145,12 @@ def _update_data(self):
         # HEADER of SPEC file
         self.hdr = self.spc.sourceHeader
 
+        try:
+            # try to download old sources from Fedora lookaside cache
+            LookasideCacheHelper.download('fedpkg', os.path.dirname(self.path), self.get_package_name())
+        except LookasideCacheError:
+            pass
```
thozza: I think we could log at least DEBUG message if the download from Lookaside cache fails.

thozza:
@nforro the changes LGTM. One cosmetic advice is to add DEBUG log message in case the download from lookaside cache fails. Feel free to merge it to master once you add it.

nforro:
@thozza: Thanks. Debug message added.

Issue #220, opened 2016-08-04
Do not run csmock by default


thozza:
csmock rebuilds SRPMs again and it takes too long to finish. Therefore I don't think we should run csmock by default, but rather expect the user to enable it explicitly when required.

phracek:
Yes could be. Therefore we have to enhance CLI so that it will support several checkers.

PR #225, opened 2016-08-18
Rework CLI class

nforro:
Allow supplying sets of tools to `pkgcomparetool` option
Do not include csmock into default set of pkgcomparetools
Move argument checks to `CLI`
Reword help messages to make them consistent and more clear

Resolves: #220


rebasehelper/cli.py
```diff
@@ -54,112 +87,121 @@ def add_args(self):
...
         )
         self.parser.add_argument(
             "--pkgcomparetool",
-            default=False,
-            help="Select the tool for comparing two packages [pkgdiff, rpmdiff, abipkgdiff, csmock]. All compare tools"
-                 " are run by default."
+            # FIXME: set of choices could be obtained dynamically
+            choices={"pkgdiff", "rpmdiff", "abipkgdiff", "csmock"},
+            default={"pkgdiff", "rpmdiff", "abipkgdiff"},
+            type=lambda s: set(s.split(',')),
```
thozza: I was thinking more about using action 'append' and allow user to use the argument more than once, but I'm OK also with your solution.

thozza:
@nforro LGTM. BTW you could use ```checkers_runner.get_supported_tools()``` and similar to get the list of supported tools dynamically. There should be an equivalent for each of the *tools.

nforro:
@thozza: Implemented, including getting default choices.

rebasehelper/build_helper.py
```diff
@@ -671,3 +672,9 @@ def get_supported_tools(cls):
         :return: list of supported build tools
         """
         return build_tools.keys()
+
+    @classmethod
+    def get_default_tool(cls):
+        """Returns default build tool"""
+        default = [k for k, v in six.iteritems(build_tools) if hasattr(v, 'DEFAULT') and v.DEFAULT]
```
thozza: @nforro you can add the DEFAULT attribute to the BuildToolBase class and set it to ```False```. Then only in the default tool class definition you can set it to ```True```. This way you won't have to use hasattr and the property of the class will be then explicitly set every time.

rebasehelper/checker.py
```diff
@@ -129,6 +129,10 @@ def get_supported_tools(self):
         """Return list of supported tools"""
         return self.plugin_classes.keys()
 
+    def get_default_tools(self):
+        """Return list of default tools"""
+        return [k for k, v in six.iteritems(self.plugin_classes) if hasattr(v, 'DEFAULT') and v.DEFAULT]
```
thozza: The same as with build tools.

thozza:
@nforro I had two comments to your new changes. Otherwise the changes look good!

nforro:
@thozza: Thanks. I made the suggested changes.

thozza:
@nforro looks good!

PR #229, opened 2016-09-07
Refactorization of SpecFile class preparing baseline for pre-download hooks - volume 1

thozza:
Please see commit messages of the individual commits.
Changes include also code cleanup which was possible due to the refactorization. This includes mostly deletion of unneeded code.

@nforro please have a look when you have time.


rebasehelper/archive.py
```diff
@@ -190,42 +190,6 @@ def extract(cls, archive=None, filename=None, path=None, *args, **kwargs):
...
-
-    @classmethod
-    def extract(cls, archive=None, filename=None, path=None, *args, **kwargs):
-        if archive is not None:
-            raise RuntimeError("In Asc files, the archive (pos 1) argument is not used, but passed")
-        final_dir = os.path.join(path, os.path.basename(filename.rstrip(cls.EXTENSION)))
-        os.makedirs(final_dir)
-        shutil.copy(filename, final_dir)
-
-@register_archive_type
```
phracek: Why did you remove it? Some files are needed from rebase-helper point of view. https://github.com/rebase-helper/rebase-helper/pull/117

rebasehelper/specfile.py
```diff
@@ -125,11 +127,35 @@ def __init__(self, path, sources_location='', download=True):
         rpm.addMacro("_sourcedir", self.sources_location)
         self._read_spec_content()
         # Load rpm information
-        self.spc = rpm.spec(self.path)
-        self.patches = self._get_initial_patches_list()
```
phracek: I guess, you would like to explicitly call method _update_data, right?
SpecFile class will be only for initialization, right?

like
~~~
sp = SpecFile(something)
sp._update_data()
~~~

Did I understand it?

rebasehelper/archive.py
thozza: Previously these were needed in order for such Source files to be downloaded from URL. The reason was bad logic in the SpecFile class method responsible for downloading Sources, because it was downloading only files with extensions of known archive types. This is not needed any more as all Sources with URL are downloaded.

.gem are different as these are listed as Source0. Arguably the handling should be improved to try to extract only files which are archives and not to add pseudo classes for such use cases.

rebasehelper/specfile.py
thozza: ```_update_data()``` is called in ```__init__()``` of SpefFile on line 131 (2 lines below your comment). Those lines were redundant as they are included in ```_update_data()``` and were called twice in case of creation of class object.

rebasehelper/specfile.py
```diff
@@ -30,7 +30,8 @@
 import shlex
 from datetime import date
 from operator import itemgetter
-from six.moves.urllib.parse import urlparse
+# TODO: remove the pylint comment once pylint with https://github.com/PyCQA/pylint/issues/551 is released
+from six.moves.urllib.parse import urlparse  # pylint: disable=import-error
```
nforro: The `import-error` can be also avoided by importing only `urllib` from `six.moves` and later calling `urllib.parse.urlparse()` instead of just `urlparse()`. Like in [utils.py](/rebase-helper/rebase-helper/blob/81ae814e3056bb7d5050d42394dc0e6cf26b4c38/rebasehelper/utils.py#L44).

nforro:
@thozza: LGTM

rebasehelper/specfile.py
thozza: @nforro true, but it seemed too long to me. I can change it if you insist.

nforro: @thozza: No, it’s fine. But for the sake of consistency we should probably change it also in `utils.py` (import `urlopen` and `urlsplit` directly).

thozza: @nforro sure, I'll do the change. Thanks for feedback!

thozza:
@nforro in the end I used the import of urllib as it is done in utils.py.

PR #231, opened 2016-09-09
landscape.io cleanups

thozza:
SSIA

### DIFF
--- a/.landscape.yaml
+++ b/.landscape.yaml
@@ -6,7 +6,10 @@ test-warnings: True
 
 # tests stricktness
 strictness: medium
-max-line-length: 120
+
+pylint:
+    options:
+        max-line-length: 120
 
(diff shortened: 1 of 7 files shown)


nforro:
There are more string literals that should probably be raw strings, like [here](/rebase-helper/rebase-helper/blob/c1cec6be2a4aac6bb5554d10d6d7c9246d4749d5/rebasehelper/specfile.py#L542) and [here](/rebase-helper/rebase-helper/blob/c1cec6be2a4aac6bb5554d10d6d7c9246d4749d5/rebasehelper/checkers/rpmdiff.py#L80).
I'm not sure why `landscape.io` doesn't report those.
Other than that - LGTM.

thozza:
@nforro thanks for review. I will change the strings to raw string in some of the future change sets.

Issue #187, opened 2016-05-27
Generate rebase-helper manual page using Sphinx


thozza:
Currently the man page is not kept up-to-date. It is static and not generated based on the latest documentation. Since Sphinx is able to generate man page, we should take advantage of this functionality and rather than manually writing the man page, generate it.

Changes of the documentation will be needed before some sane man page can be generated using `make man` in the `docs/` directory

thozza:
Some more info: http://www.sphinx-doc.org/en/stable/config.html#confval-man_pages

phracek:
https://github.com/devassistant/devassistant/blob/master/distribution/manpages/generate_man.py

PR #233, opened 2016-09-23
Documentation changes

nforro:
This PR adds posibility to automatically generate program documentation from ArgumentParser instance. It introduces Sphinx extension `autoargs`, which is similar to [sphinx-argparse](https://github.com/ribozz/sphinx-argparse), but actually quite different. There is also custom manpage builder `custom-man`, which is needed because default Sphinx manpage builder is not very usable when you want the same source for both html and man.

You can see html output on [readthedocs](http://rebase-helper.readthedocs.io/en/documentation-changes/usage.html).

To see the manpage, run the following in `docs` subdir:
```
make man
man -l _build/man/rebase-helper.1
```

This resolves issue #187.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -114,7 +114,7 @@ text:
 	@echo "Build finished. The text files are in $(BUILDDIR)/text."
 
 man:
-	$(SPHINXBUILD) -b man $(ALLSPHINXOPTS) $(BUILDDIR)/man
+	$(SPHINXBUILD) -b custom-man $(ALLSPHINXOPTS) $(BUILDDIR)/man
 	@echo
 	@echo "Build finished. The manual pages are in $(BUILDDIR)/man."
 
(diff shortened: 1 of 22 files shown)


phracek:
Looks good to me.

nforro:
@phracek: Thanks.
@thozza: Can you also take a look at this when you have time?

thozza:
@nforro The changes look good. Good job! I have just a small comment. The copyright should be given to Red Hat. At the bottom of the man page I see:
```
COPYRIGHT
       2016, Petr Hracek, Tomas Hozza
```
Can it be changed to something like?
```
COPYRIGHT
       2014-2016, Red Hat
```

nforro:
@thozza: Sure, it's just a matter of changing `copyright` variable in  `docs/conf.py`. That will of course change copyright in html documentation as well.

PR #234, opened 2016-09-26
Install python3-tox on Gitlab CI

nforro:
In Fedora rawhide tox binary moved from python-tox to python3-tox.
Install python3-tox package since doing so won't do any harm on
F23 or F24 and it will prevent test failure on rawhide.


nforro:
Gitlab CI [build](https://gitlab.com/rebase-helper/rebase-helper/commit/9f63425a87547eed4ddd00a03f9b0ea89dede31b/builds).

Issue #235, opened 2016-10-04
missing \n in rebase helper spec file


jridky:
I have tried to make rebase from screen-4.1.0 to screen-4.4.0 for RHEL with following result:

```
Results directory 'rebase-helper-results' exists, removing it
Workspace directory 'rebase-helper-workspace' exists, removing it
error: line 68: Bad %setup option -p1: unknown option
error: line 68: Bad %setup option -p1: unknown option
Applying patch 'screen-4.0.3-libs.patch' to 'screen'

Patching failed
```

In rebase-hepler-results/screen.spec:

```
%prep
#%setup -q -n screen/src
%setup -q -n %{name}%patch1 -p1 -b .libs          # !!! there is missing \n !!!
%patch2 -p1 -b .screenrc
```

jridky:
https://github.com/rebase-helper/rebase-helper/blob/master/rebasehelper/specfile.py#L1154

there is probably the mistake

nforro:
Thanks for reporting. There is of course newline missing, I will fix that.

There is however another issue: all patches will fail as a result of `%setup` dirname change.
That one will probably require some time to fix.

PR #236, opened 2016-10-07
Handle %setup dirname argument containing slashes

nforro:
In case `%setup` macro is specified e.g. like this:

`%setup -q -n screen/src`

it means that when rpmbuild is running `%prep`, sources are extracted
to `src` subdirectory.

Switch to such subdirectory after extracting sources, to prevent
patching failures. Also preserve such subdirectories when updating
`%setup` dirname argument.

Add missing newline after updating `%setup` macro.

This PR resolves issue #235.


nforro:
Replaced `line.split()` with `shlex.split(line)` to make `%(auto)setup` processing more robust.

PR #237, opened 2016-11-01
Implementation spec hooks.

phracek:

This implements the spec hooks. Feel free to comment it.
Tests are missing. I don't have any hook for testing now but can think about some "foo" hook.


thozza:
@phracek please add at least some dummy testing hook and unit tests. We need to test that the code is working correctly. Also the existing checks are not passing. Please also squash the two commits with README into a single one.

nforro:
Spec hooks were implemented in PR #256.

Issue #193, opened 2016-06-08
'The file has zero size' exception when HTTP response has no Content-length header


nforro:
Sometimes github servers don't send _Content-length_ header in a HTTP response and rebase-helper fails with `The file has zero size`.

I don't have a solid reproducer, but I noticed it usually happens on the first download request after a period of inactivity. Try this URL for instance: https://github.com/fatih/vim-go/archive/v1.7.1.tar.gz

I think it would be better in this case to show some infinite progress indicator in place of the percentual one and let the download finish.

phracek:
I have the same problem in Dockerfile. I will add it to GitHub soon.

phracek:
https://github.com/rebase-helper/rebase-helper/pull/239

~~~
Rebase-helper version: 0.8.0
Updating the extra version
Tarball is not in absolute path /package_name/2.1.9.tar.gz trying to download one from URL https://github.com/upgrades-migrations/preupgrade-assistant/archive/2.1.9.tar.gz
Tarball is not in absolute path. Trying to download it from URL

Failed to download file from URL https://github.com/upgrades-migrations/preupgrade-assistant/archive/2.1.9.tar.gz. Reason: 'The file has zero size'. 
~~~

nforro:
There are also FTP servers which don't send file length, for example `ftp://ftp.gnupg.org/`.
Downloading any file from such FTP server also leads to `The file has zero size` error.

PR #244, opened 2016-11-23
Do not fail when size of downloaded file is not known

nforro:
Introduce wget-like download progress bar, which can show some statistics also in case the size of downloaded file is not known.

This PR resolves issue #193.

I thought this should be resolved rather sooner than later, since many rebases initiated by upstream monitoring are failing because of this.

@phracek, @thozza: Can you review this PR when you have time?


phracek:
Looks good. +1

PR #246, opened 2016-12-08
Fix newlines in reports from rpmdiff and pkgdiff

nforro:


rebasehelper/checkers/pkgdiff.py
```diff
@@ -229,16 +229,19 @@ def run_check(cls, results_dir):
...
+                if lines:
+                    lines.append('')
+                lines.append('Following files were %s:' % key)
+                lines.extend(val)
 
         pkgdiff_report = os.path.join(cls.results_dir, 'report-' + cls.pkgdiff_results_filename + '.log')
         try:
             with open(pkgdiff_report, "w") as f:
-                f.writelines(text)
+                f.write('\n'.join(lines))
```
phracek: You can use writelines meanwhile. Why to use only write?

rebasehelper/checkers/rpmdiff.py
```diff
@@ -130,15 +130,19 @@ def run_check(cls, results_dir):
...
+                lines.append('Following files were %s:' % key)
+                lines.extend(val)
 
-        pkgdiff_report = os.path.join(cls.results_dir, 'report-' + cls.CMD + '.log')
+        rpmdiff_report = os.path.join(cls.results_dir, 'report-' + cls.CMD + '.log')
         try:
-            with open(pkgdiff_report, "w") as f:
-                f.writelines(text)
+            with open(rpmdiff_report, "w") as f:
+                f.write('\n'.join(lines))
```
phracek: The same. We can still use writelines when the lines is array.

rebasehelper/checkers/pkgdiff.py
nforro: `write` and `join` is more convenient here (and IMHO also more clear), because `writelines` doesn't add newline characters, so I would have to add them to the list of lines "manually".

PR #250, opened 2016-12-15
Explicitly set git username and e-mail address

nforro:
Many git commands fail when `user.name` and/or `user.email` is not configured.

Don't rely on system git configuration (which may not even exist), rather set the values
explicitly on each git call.

### DIFF
--- a/rebasehelper/utils.py
+++ b/rebasehelper/utils.py
@@ -730,6 +730,8 @@ class GitHelper(object):
     """Class which operates with git repositories"""
 
     GIT = 'git'
+    GIT_USER_NAME = 'rebase-helper'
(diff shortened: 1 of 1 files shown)


nforro:
@phracek: Sure, that could be done, but I don't see the point. AFAIK username and e-mail is not visible anywhere anyway.

phracek:
@nforro I guess, this Pull Request automatically adds user name and mail. What about case, I have my own? You over write it. And I think, it is not expected.

nforro:
@phracek: But that username and e-mail is used only for commits in temporary git repositories, so I think it doesn't really matter if those are real or not.

PR #251, opened 2016-12-15
Koji compatibility changes

nforro:
Starting with `koji-1.11.0`, `koji` buildtool is not working anymore.
Make necessary changes to get it working again.


rebasehelper/utils.py
```diff
@@ -984,7 +983,11 @@ def session_maker(cls, baseurl=None):
         else:
             koji_session = koji.ClientSession(baseurl)
             return koji_session
-        koji_session.ssl_login(cls.cert, cls.ca_cert, cls.ca_cert)
+        try:
+            koji_session.krb_login()
```
phracek: Do we really want krb_login?

nforro: I don't think we have any other choice. Or do we?

PR #253, opened 2017-02-27
Use name and e-mail from Git configuration if the value is configured

pspacek:
Values provided in 19b94d1b46f7cae65a30b4898b62afb998223310 are used as
fallback.

Using system-provided values makes easies to re-use the generated spec.

### DIFF
--- a/rebasehelper/utils.py
+++ b/rebasehelper/utils.py
@@ -730,12 +730,37 @@ class GitHelper(object):
     """Class which operates with git repositories"""
 
     GIT = 'git'
+    # provide fallback values if system is not configured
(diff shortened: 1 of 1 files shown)


nforro:
Hi, thanks for contributing.

Yeah, I didn't realize that git user.name and user.email is what is being used for generating changelog lines in the spec file.

pspacek:
I think it is inevitable, `GitHelper.command_config` cannot be used because it calls `_call_git_command`, which in turn depends on the values we are trying to initialize.

nforro:
Got it. Thanks.

PR #256, opened 2017-04-27
Implement spec hooks

nforro:
Spec hooks are implemented using entry points based approach.
Sample **typo fix** hook and test for it is included.
This PR obsoletes #237.

### DIFF
--- a/rebasehelper/spec_hooks/typo_fix.py
+++ b/rebasehelper/spec_hooks/typo_fix.py
@@ -1,0 +1,47 @@
+# -*- coding: utf-8 -*-
+#
+# This tool helps you to rebase package to the latest version
+# Copyright (C) 2013-2014 Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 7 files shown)


rebasehelper/spec_hooks/typo_fix.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class TypoFixHook(BaseSpecHook):
+    """Sample spec hook that fixes typos in spec file"""
+
+    NAME = 'Typo fix'
+    REPLACEMENTS = [
+        ('chnagelog', 'changelog'),
+        ('indentional', 'intentional'),
+    ]
```
skisela: That is a handy feature. Won't it get unmaintainable in time?

nforro: This was actually meant to be just a trivial sample spec hook. But perhaps you are right, and it could actually be useful. But in that case it should probably take advantage of some dictionary of typos, like [this one from Wikipedia](https://en.wikipedia.org/wiki/Wikipedia:Lists_of_common_misspellings), and also some logic to prevent false positives should be implemented.

PR #258, opened 2017-04-28
Run tests on Travis CI inside Docker containers

nforro:
Testing on Travis CI is now done the same way as on Gitlab CI,
tests are run inside Docker containers for each Fedora version.


nforro:
@thozza: Could you take a quick look at this? Thanks.

nforro:
Nice idea, I'll try to implement that (but probably in a new PR).

PR #260, opened 2017-05-04
Add make target for running tests inside Docker containers

nforro:


thozza:
@nforro are you sure you want to merge this to the `nforro-doc-building-fixes` branch?

nforro:
@thozza: Yes, because it kind of depends on that branch (there are commits in both which modify Makefile). Actually all 7 of my latest pull requests are chained, it seemed like a good idea to me because of dependency between them. I plan to merge them together after they are all approved.

.travis.yml
```diff
@@ -5,5 +5,5 @@ env:
     - DOCKER_IMAGE=fedora:24
     - DOCKER_IMAGE=fedora:25
     - DOCKER_IMAGE=fedora:rawhide
```
nforro: Do you mean replacing `fedora:24` and `fedora:25` with `fedora:latest`? I used `fedora:latest` in Makefile.docker just to avoid the need of changing the value every time new Fedora is released.

Makefile.docker
```diff
@@ -0,0 +1,15 @@
...
+# so rather replace it with placeholder
+TEST_IMAGE := $(subst :,_COLON_,$(DOCKER_IMAGE))
+
+.PHONY: test $(TEST_IMAGE)
+
+test: $(TEST_IMAGE)
+
+$(TEST_IMAGE):
+	$(eval IMAGE := $(subst _COLON_,:,$@))
+	sudo docker pull $(IMAGE)
```
nforro: Ok, I'll add check for docker group and call sudo only in case it's really needed.

.travis.yml
thozza: I don't have preference, but the the situation with `travis.yml` is the same as with `Makefile.docker` - it will have to be updated with every new Fedora. If it is just added as additional image, we will be most probably testing one Fedora version twice, but on the other hand we won't miss new Fedora if we don't update the file on time. If we replace Fedora versions with just `fedora:latest`, we won't test rebase-helper on "latest stable version" - 1. I'm OK also if no change will be done here, but it seemed right to consolidate Makefile and config file for Travis.

nforro: I especially don't like the necessity to change it in two places. I could actually just remove it from .travis.yml, but that would mean losing the build matrix (separate build jobs for each Fedora release), which I think is nice.
But I agree that it's probably better to keep .travis.yml and Makefile.docker synchronized, so I'll probably change both to contain `fedora:24`, `fedora:latest` and `fedora:rawhide`.

thozza: +1

PR #261, opened 2017-05-18
Documentation improvements

nforro:
You can take a look at [generated documentation](http://rebase-helper.readthedocs.io/en/nforro-documentation-improvements/index.html).


nforro:
Maybe SPEC hooks deserve their own section in the documentation.
And I agree that some control over them should be implemented.

Issue #150, opened 2016-02-07
Provide git am-able patches or a pullable repo on success


simo5:
As a maintainer I want final word on pulling any changes, but if rebase-helper did a good job I'd like to just pull the changes it applied automatically and fire an official build.

That can be either:
- a patch that can be imported in dist-git with git am
- a repository from which I can pull the patch and then git cherry-pick on dist git
- a branch in dist git with the proposed commit

PR #266, opened 2017-05-29
Generate a patch after rebase

skisela:
Move modified patches and rebased specfile to a subdirectory
results_dir/rebased_sources for better organization.
Copy also unmodified used patches to the directory, to be able
to generate patch containing all changes between the current
package version patches and specfile, and the rebased version.

Generate 'changes.patch' for simple application of the result
and place it in the results_dir.

Change paths to patches in 'rebase_spec_file' back to the current
directory after rebase, to provide fully functional rebased sources
in the 'rebased_sources' directory.

Add a test case to check 'changes.patch'.
Add a test case to check removed patches.

Modify report from rebase-helper.

Resolves: #150


skisela:
Rebased specfile contains paths to patches in the subdirectory(modified) and in the base directory(unmodified). That might be changed as well, for more convenient usage of the whole new directory.

nforro:
I'm afraid that `newsources`, resp. `new_sources_dir` is not the best name for the subdirectory. Everywhere else in **rebase-helper** it's basically a synonym of upstream version source tarball.

rebasehelper/application.py
```diff
@@ -412,6 +417,23 @@ def patch_sources(self, sources):
...
+        Generates patch to the results_dir containing all needed files for
+        the rebased package version
+        """
+        # Copy applied untouched patches to the newsources directory
+        SpecFile.copy_applied_patches(self.new_sources_dir, self.spec_file, self.rebase_spec_file)
+
+        dir = os.getcwd()
+        os.chdir(self.new_sources_dir)
+        gh = GitHelper('./')
+        gh.command_init('./')
```
nforro: Is there a reason to use `'./'` and not just `'.'`?

rebasehelper/application.py
```diff
@@ -412,6 +417,23 @@ def patch_sources(self, sources):
...
+        gh.command_add_files('.')
+        gh.command_commit('Init commit.')
+        gh.command_format_patch(directory = self.results_dir)
```
nforro: There shoudn't be spaces around `=` when passing a keyword argument, see [PEP 8 style guide](https://www.python.org/dev/peps/pep-0008/#other-recommendations).

rebasehelper/application.py
```diff
@@ -412,6 +417,23 @@ def patch_sources(self, sources):
...
+    def generate_patch(self):
+        """
+        Generates patch to the results_dir containing all needed files for
+        the rebased package version
+        """
+        # Copy applied untouched patches to the newsources directory
+        SpecFile.copy_applied_patches(self.new_sources_dir, self.spec_file, self.rebase_spec_file)
+
+        dir = os.getcwd()
+        os.chdir(self.new_sources_dir)
```
nforro: I don't think changing directory is necessary.

rebasehelper/utils.py
```diff
@@ -892,6 +892,18 @@ def command_commit(self, message=None, amend=False):
...
+        """
+        Method generates patch based on last commit
+
+        :param directory: Directory where patch will be placed
+        :param patchname: Name of the patch
+        :return: return code from ProcessHelper
+        """
+        cmd = ['format-patch', '--stdout', '-1']
+
+        return self._call_git_command(cmd, output_file=directory + '/' + patchname)
```
nforro: It would be probably better to use `os.path.join(directory, patchname)` here.

rebasehelper/utils.py
```diff
@@ -892,6 +892,18 @@ def command_commit(self, message=None, amend=False):
 
         return self._call_git_command(cmd)
 
+    def command_format_patch(self, patchname="changes.patch", directory="./"):
```
nforro: Why are you using double quotes here and single quotes below? If there is no explicit reason, I think you should choose one or the other for the sake of consistency.

rebasehelper/specfile.py
```diff
@@ -403,15 +404,79 @@ def write_updated_patches(self, patches):
...
+
+        :param patches: all patches not commented in the rebase_spec_file
+        :param spec_file: spec_file object
+        :return: List of unmodified used patches
+        """
+        copy_patches = []
+
+        for patch in patches:
+            copy = True
+            for pnum in spec_file.modified_patches:
```
nforro: `spec_file.modified_patches` will always be empty, since `write_updated_patches()` method is called only on `rebase_spec_file`.

rebasehelper/specfile.py
```diff
@@ -403,15 +404,79 @@ def write_updated_patches(self, patches):
...
+                if pnum in patch:
+                    copy = False
+
+            if copy:
+                copy_patches.append(patch)
+        return copy_patches
+
+    @staticmethod
+    def get_used_patches(spec_file, rebase_spec_file):
```
nforro: What's the point of this method? Why don't you just use `rebase_spec_file.patches['applied']`?

nforro:
As you said, rebased SPEC file contains paths to patches in the subdirectory. That has to be changed, if you expect users to use whole contents of the subdirectory.

rebasehelper/application.py
```diff
@@ -413,6 +421,39 @@ def patch_sources(self, sources):
...
+        """
+        # Delete removed patches from rebased_sources_dir from git
+        for removed_patch in self.rebase_spec_file.removed_patches:
+            self.git_helper.command_rm(os.path.join(self.rebased_sources_dir, removed_patch))
+
+        # Fix paths in rebase_spec_file to patches to current directory
+        for index, line in enumerate(self.rebase_spec_file.spec_content):
+            mod_line = re.sub(settings.REBASE_HELPER_REBASED_SOURCES_DIR + '/', '', line)
+            self.rebase_spec_file.spec_content[index] = mod_line
+        self.rebase_spec_file.save()
```
nforro: This solution doesn't seem to be very robust, wouldn't it be better to modify `SpecFile.write_updated_patches()` method not to prepend the subdirectory to modified patches?

rebasehelper/application.py
```diff
@@ -413,6 +421,39 @@ def patch_sources(self, sources):
...
+
+        # Generate patch
+        self.git_helper.command_add_files('.')
+        self.git_helper.command_commit(message='Get difference between versions')
```
nforro: I think it would be better if the commit message would be the same as changelog entry.

rebasehelper/application.py
skisela: Well the subdirectories must be prepended during the build of rebased sources.
The second option is to change the current working directory to the 'rebased_sources' directory.
Then we might need to modify few other paths.

rebasehelper/application.py
skisela: You're right. I will modify that.

rebasehelper/application.py
nforro: Got it, but I still don't like this. You could at least check if a line starts with `Patch`.
I think there has to be more elegant solution, but I can't think of any right now ;-)

skisela: Sure, I will rethink it, thanks!

rebasehelper/application.py
```diff
@@ -413,6 +421,39 @@ def patch_sources(self, sources):
...
+        self.git_helper.command_format_patch(directory=self.results_dir)
+
+    def _prepare_rebased_repository(self):
+        """
+        Initialize git repository in the rebased directory
+        :return:
+        """
+        for patch in self.rebase_spec_file.patches['applied']:
```
nforro: It probably doesn't really matter in this phase, but using `self.spec_file.patches['applied']` makes more sense to me.

rebasehelper/application.py
```diff
@@ -716,8 +757,15 @@ def run(self):
         if not self.conf.keep_workspace:
             self._delete_workspace_dir()
 
+        self.generate_patch()
+
         if self.debug_log_file:
             logger.info("Detailed debug log is located in '%s'", self.debug_log_file)
+            logger.info("Modified patches and specfile needed for the rebased version "
+                        "are located in '%s'", self.rebased_sources_dir)
+            logger.info("Patch '%s' containing changes of patches nad specfile between "
```
nforro: There is a typo, and I think I would just use `rebased spec file and patches`. So `Patch '%s' containing changes between original and rebased spec file and patches ...`, and `Rebased spec file and patches are located ...`.

rebasehelper/settings.py
```diff
@@ -28,6 +28,7 @@
 
 REBASE_HELPER_RESULTS_DIR = REBASE_HELPER_PREFIX + "results"
 REBASE_HELPER_WORKSPACE_DIR = REBASE_HELPER_PREFIX + "workspace"
+REBASE_HELPER_REBASED_SOURCES_DIR = REBASE_HELPER_PREFIX + "results/rebased_sources"
```
nforro: Why not `REBASE_HELPER_REBASED_SOURCES_DIR = os.path.join(REBASE_HELPER_RESULTS_DIR, "rebased_sources")` just like below?

rebasehelper/specfile.py
```diff
@@ -397,13 +398,14 @@ def write_updated_patches(self, patches):
                     self.spec_content[index] = '#{0} {1}\n'.format(' '.join(fields[:-1]),
                                                                    os.path.basename(patch_name))
                     if patch_removed:
+                        self.removed_patches.append(patch_name)
```
nforro: Is this used anywhere?

rebasehelper/application.py
nforro: The problem is `%{_sourcedir}` macro defines location of both source tarballs and downstream patches. So you would have to copy also source tarballs before changing directory (or redefining `%{_sourcedir}` macro).

nforro: So your solution with modifying rebased spec file after build is probably acceptable.
But I think you should move that to `SpecFile` class.

skisela:
Thanks for comments. I will fix the issues and add a test case for this enhancement.

rebasehelper/specfile.py
```diff
@@ -346,6 +347,13 @@ def _comment_out_patches(self, patch_num):
                         patch_num.remove(num)
                         break
 
+    def renew_paths_to_patches(self):
+        # Fix paths in rebase_spec_file to patches to current directory
+        for index, line in enumerate(self.spec_content):
```
nforro: I'd still like to see some additional check like `if line.startswith('Patch'):` here.

rebasehelper/specfile.py
```diff
@@ -346,6 +347,13 @@ def _comment_out_patches(self, patch_num):
                         patch_num.remove(num)
                         break
 
+    def renew_paths_to_patches(self):
```
nforro: Why not `update_paths_to_patches`? It sounds a little bit better to me.

rebasehelper/tests/test_specfile.py
```diff
@@ -460,3 +461,13 @@ def test_pypi_to_python_hosted_url_trans(self):
...
+
+    def test_update_paths_to_patches(self):
+        """
+        Check updated paths to patches in the rebased directory
+        :return:
+        """
+        self.SPEC_FILE_OBJECT.update_paths_to_patches()
+
+        for line in self.SPEC_FILE_OBJECT.spec_content:
+            assert not re.search('^Patch.*rebased_sources/', line)
```
nforro: I think it is sufficient to check only `Patch5` (before and after calling the method).
You can also use `self.SPEC_FILE_OBJECT.hdr[rpm.RPMTAG_PATCH]` to access `Patch` tags.

skisela: I am afraid this approach provides us only with patch name instead of the full relative path to it.

nforro: Right, I didn't notice that. But I still think that you should first find the `Patch5` line and then make assertion on it, rather than doing that for every line.

skisela: You are absolutely right, see last commit fixing that already.

rebasehelper/tests/test_specfile.py
```diff
@@ -460,3 +461,18 @@ def test_pypi_to_python_hosted_url_trans(self):
...
         assert 'https://files.pythonhosted.org/' in self.SPEC_FILE_OBJECT._get_raw_source_string(7)
+
+    def test_update_paths_to_patches(self):
+        """
+        Check updated paths to patches in the rebased directory
+        :return:
+        """
+        for line in self.SPEC_FILE_OBJECT.spec_content:
+            if line.startswith('Patch5'):
+                assert re.search('.*rebased_sources/', line)
```
nforro: How about this?
```
line = [l for l in self.SPEC_FILE_OBJECT.spec_content if l.startswith('Patch5')][0]
assert 'rebased_sources' in line
```

skisela: Much better, thanks!

Issue #222, opened 2016-08-08
JSON output contains items that should not be included


thozza:
When using JSON outputtool, the `builds` section contains items that should not be included, since these are just internal arguments of the builder class. IOW these should not be exposed in the JSON results output.

The `builds` section in output should **NOT** contain items:
- `build_tasks`
- `builder_options`
- `builds_nowait`

The result now looks like this:

``` json
{
    "builds": {
        "new": {
            "build_tasks": null, 
            "builder_options": null, 
            "builds_nowait": false, 
            "logs": [
                "/home/thozza/devel/fedora/unbound/rebase-helper-results/new/SRPM/build.log", 
                "/home/thozza/devel/fedora/unbound/rebase-helper-results/new/RPM/build.log"
            ], 
            "name": "unbound", 
            "rpm": [
                "/home/thozza/devel/fedora/unbound/rebase-helper-results/new/RPM/unbound-munin-1.5.8-1.fc24.noarch.rpm", 
                "/home/thozza/devel/fedora/unbound/rebase-helper-results/new/RPM/unbound-debuginfo-1.5.8-1.fc24.x86_64.rpm", 
                "/home/thozza/devel/fedora/unbound/rebase-helper-results/new/RPM/python3-unbound-1.5.8-1.fc24.x86_64.rpm", 
                "/home/thozza/devel/fedora/unbound/rebase-helper-results/new/RPM/python-unbound-1.5.8-1.fc24.x86_64.rpm", 
                "/home/thozza/devel/fedora/unbound/rebase-helper-results/new/RPM/unbound-libs-1.5.8-1.fc24.x86_64.rpm", 
                "/home/thozza/devel/fedora/unbound/rebase-helper-results/new/RPM/unbound-devel-1.5.8-1.fc24.x86_64.rpm", 
                "/home/thozza/devel/fedora/unbound/rebase-helper-results/new/RPM/unbound-1.5.8-1.fc24.x86_64.rpm"
            ], 
            "srpm": "/home/thozza/devel/fedora/unbound/rebase-helper-results/new/SRPM/unbound-1.5.8-1.fc24.src.rpm", 
            "version": "1.5.8"
```

PR #271, opened 2017-06-07
Improve rebase-helper output information

skisela:
Implement entry point approach for output generators

Implement Text and JSON output tools as plugins.
Generate report.{text, json} into results directory based on --outputtool parameter.

Add colored CLI output.

Modify exception handling in Application:run for better exception message handling.

Modify RebaseHelperError exception constructor.

Resolves: #222, #292


README.md
```diff
@@ -7,7 +7,7 @@ There are several steps that need to be done when rebasing a package. The goal o
 
 ## General workflow
 - *rebase-helper-workspace* and *rebase-helper-results* directories are created
-- original SPEC file is copied to *rebase-helper-results* directory and its Version tag is modified
+- original SPEC file is copied to *rebase-helper-results/rebased_sources* directory and its Version tag is modified
```
nforro: This is probably not the right place to complain about this, but here the inconsistency really stands out. `-` is used in `rebase-helper-results` as a separator while in `rebased_sources` it's `_`.

README.md
```diff
@@ -21,7 +21,9 @@ There are several steps that need to be done when rebasing a package. The goal o
 - new sources are extracted and added as a remote repository
 - `git-rebase` is used to rebase the commits on top of new sources
 - original patches are modified/deleted accordingly
-- changes are reflected in the spec file
+- rebased sources are stored in *rebase-helper-results/rebased_sources*. These files are needed for the new version.
```
nforro: > These files are needed for the new version.

What's this supposed to mean? I'd just drop it.

README.md
```diff
@@ -21,7 +21,9 @@ There are several steps that need to be done when rebasing a package. The goal o
...
+- Patch *changes.patch* containing changes between original and rebased spec file and patches is located in *rebase-helper-results*
```
nforro: I think this sentence deserves some better wording. I'd use `File` instead of `Patch` to avoid redundancy at least. But I think it should be rewritten completely.

rebasehelper/application.py
```diff
@@ -325,8 +323,9 @@ def _check_results_dir(results_dir):
             shutil.rmtree(results_dir)
         os.makedirs(results_dir)
         os.makedirs(os.path.join(results_dir, settings.REBASE_HELPER_LOGS))
-        os.makedirs(os.path.join(results_dir, 'old'))
-        os.makedirs(os.path.join(results_dir, 'new'))
+        os.makedirs(os.path.join(results_dir, 'old_build'))
+        os.makedirs(os.path.join(results_dir, 'new_build'))
```
nforro: Same as above. We should stick to one consistent separator character.

rebasehelper/application.py
```diff
@@ -678,11 +676,29 @@ def get_rebased_patches(self):
             output_patch_string.append('Patches were not touched. All were applied properly')
         return output_patch_string
 
-    def print_summary(self):
-        output = output_tool.OutputTool(self.conf.outputtool)
-        report_file = os.path.join(self.results_dir, self.conf.outputtool + settings.REBASE_HELPER_OUTPUT_SUFFIX)
-        output.print_information(path=report_file)
-        logger.info('\nReport file from rebase-helper is available here: %s', report_file)
+    def print_summary(self, message=None):
```
nforro: Why is the argument called `message` when it's in fact `Exception` object? Wouldn't `exception` or `error` be more suitable?

rebasehelper/application.py
```diff
@@ -678,11 +676,29 @@ def get_rebased_patches(self):
...
+        Save rebase-helper result and print the summary using output_tools_runner
+        :param message: Error message from rebase-helper
+        :return:
+        """
+        log = None
+        # Store rebase helper result message
+        if message:
+            if len(message.args) > 1:
+                result = message.args[0] % message.args[1:]
+                if os.path.exists(message.args[2]):
```
nforro: This is potential out-of-bounds access. How do you know that a third argument is a log file anyway?

rebasehelper/output_tool.py
```diff
@@ -19,216 +19,185 @@
...
-        cls.print_pkgdiff_tool(results.get_checkers())
+        if cls.app.conf.non_colored_cli_output:
+            print(string)
+        else:
+            if color == 'green':
+                print(green(string))
+            elif color == 'yellow':
+                print(yellow(string))
+            elif color == 'red':
+                print(red(string))
```
nforro: You could pass the function directly, instead of its name as a string.

rebasehelper/output_tools/html_output_tool.py
```diff
@@ -0,0 +1,367 @@
...
+import os
+import six
+
+from rebasehelper.utils import ProcessHelper
+from rebasehelper.output_tool import BaseOutputTool
+from rebasehelper.results_store import results_store
+from rebasehelper.logger import logger
+
+from yattag import Doc
+doc, tag, text = Doc().tagtext()
```
nforro: What's the point of global variables?

rebasehelper/output_tools/html_output_tool.py
```diff
@@ -0,0 +1,367 @@
...
+        with tag('h3'):
+            text('%s version logs' % ver)
+
+        for log in logs['logs']:
+            with tag('div', klass='well'):
+                with tag('h4'):
+                    text(log)
+                cls.write_file_content(log)
+
+
```
nforro: What's up with all these extra newlines?

rebasehelper/tests/test_output_tool.py
```diff
@@ -131,21 +137,24 @@ def get_expected_json_output():
...
+                "success": "Success"
             }
         }
         return expected_output
 
     def test_text_output_tool(self, results_file_path, results_store):
-        output = OutputTool('text')
-        output.print_information(results_file_path, results_store)
+        assert TextOutputTool.get_name() in output_tools_runner.output_tools
+        TextOutputTool.print_summary(results_file_path, results_store)
```
nforro: OK, the test will fail if there is an exception, but that's not the point of unit tests, this kind of thing should go to functional tests.

rebasehelper/tests/test_output_tool.py
```diff
@@ -131,21 +137,24 @@ def get_expected_json_output():
...
 
         with open(results_file_path) as f:
             lines = [y.strip() for y in f.readlines()]
             assert lines == self.get_expected_text_output().split('\n')
 
     def test_json_output_tool(self, results_file_path, results_store):
-        output = OutputTool('json')
-        output.print_information(results_file_path, results_store)
+        assert JSONOutputTool.get_name() in output_tools_runner.output_tools
+        JSONOutputTool.print_summary(results_file_path, results_store)
```
nforro: Same as above.

rebasehelper/cli.py
```diff
@@ -172,6 +172,13 @@ def build_parser():
             help="disable inapplicable patches in rebased SPEC file"
         )
         parser.add_argument(
+            "--non-colored-cli-output",
```
nforro: I don't like this very much. What about simple `--nocolors`?

CHANGELOG.md
```diff
@@ -2,11 +2,14 @@
 
 ## [Unreleased]
 ### Added
+- Implemented anytia and pypi versioneers (for looking up the current upstream package version)
+- Added possibility to download latest build from Koji instead of building it
+- Added functional tests run from cron jobs on travis CI
```
nforro: I don't think this kind of thing belongs to changelog.

rebasehelper/cli.py
```diff
@@ -200,6 +200,13 @@ def build_parser():
             help=argparse.SUPPRESS
         )
         parser.add_argument(
+            "--pure-html-output",
+            dest="pure_html_output",
+            action="store_true",
+            default=False,
+            help="generate pure html when using html output tool"
```
nforro: I don't think anybody minds CSS, and if so, one can easily disable (or replace) it oneself. My concern was about javascript, and the usability of the report with it being disabled.

rebasehelper/application.py
skisela: You're right. That's just a naive expectation.

I am thinking of modifying RebaseHelperError exception to accept log filepath as an optional argument. What do you think?

rebasehelper/output_tool.py
skisela: Thanks for the idea!

rebasehelper/cli.py
skisela: In that case, this option can be removed totally, as no javascript is being generated nor used.

CHANGELOG.md
skisela: Well, I consider those changes as quite beneficial and I think they deserve to be mentioned. Or is it enough to mention them in the documentation only?

README.md
skisela: You are right. I will stick to the `-` separator with directory names.

rebasehelper/application.py
nforro: Yes, that definitely sounds better.

CHANGELOG.md
nforro: I agree that these changes are beneficial, but from user's point of view nothing actually changed.
And I always thought changelog is meant to be for users.
But I suppose keeping it won't do any harm and users can just ignore it ;-)

rebasehelper/cli.py
nforro: Sure.

rebasehelper/exceptions.py
```diff
@@ -26,7 +26,17 @@ class RebaseHelperError(Exception):
...
     catching some expected and well known exception/error.
     """
-    pass
+    def __init__(self, msg, logfile=None):
+        """
+        Store rebase error into variables
+        :param msg: Rebase Helper error
+        :param logfile: Path to a logfile containing origin of the problem
+        """
+        if len(msg[0]) > 1:
```
nforro: `msg` can be `None` or `len(msg)` can be 0. I think this method should rather look like this:

```python
def __init__(self, *args, **kwargs):
    if not args:
        self.msg = None
    elif len(args) > 1:
        self.msg = args[0] % args[1:]
    else:
        self.msg = args[0]
    self.logfile = kwargs.get('logfile')
```

rebasehelper/application.py
```diff
@@ -768,43 +790,38 @@ def run(self):
                     build = self.build_packages()
                     if self.conf.builds_nowait and not self.conf.build_tasks:
                         return
-                except RuntimeError:
-                    logger.error('Unknown error caused by build log analysis')
-                    # TODO: exception should be raised instead of returning a value - it is never checked!
```
nforro: The comment was there for a reason - return value of `Application.run` is just ignored, so it would be good to do something about it. Either handle the return value in `ClieHelper.run`, on reraise the exception.

rebasehelper/output_tools/html_output_tool.py
```diff
@@ -0,0 +1,369 @@
...
+    @classmethod
+    def print_report_file_path(cls):
+        """Print path to the report file into the CLI"""
+        cls.colored_output('yellow', '%s report' % cls.NAME)
+        print('%s' % os.path.join(cls.html_path, 'report.' + cls.NAME))
+
+    @classmethod
+    def page_setup(cls):
+        """Generates html menu and loads css styles"""
+        global doc, tag, text
```
nforro: Again, why global? If you need the variables in other methods why don't you make them class variables?

rebasehelper/output_tools/html_output_tool.py
```diff
@@ -0,0 +1,369 @@
...
+            try:
+                with tag('div', klass='well'):
+                    text(results['fail'])
+
+                if log:
+                    with tag('h2'):
+                        text('Last 20 lines of %s' % log)
+
+                # Get log containing important message
+                ProcessHelper.run_subprocess(['tail', log, '-n20'], output='error.log')
```
nforro: This could be easily replaced with python implementation.

rebasehelper/output_tools/html_output_tool.py
```diff
@@ -0,0 +1,369 @@
...
+
+                with open('error.log', 'r') as f:
+                    with tag('div', klass='well'):
+
+                        for l in f.readlines():
+                            doc.asis('%s<br>' % l)
```
nforro: I can imagine there are better ways how to present part of a log file in HTML, how about using `<pre>` tag?

rebasehelper/output_tools/html_output_tool.py
```diff
@@ -0,0 +1,369 @@
...
+        Print content of a file to HTML
+
+        :param filename: name of a file that will be printed to HTML
+        """
+        logger.debug("Writing content of %s" % filename)
+        try:
+            with open(filename, 'r') as f:
+                for line in f.readlines():
+                    doc.asis(line + '<br>')
+        except:
```
nforro: You should always catch specific exceptions, here it would be probably `IOError`.

rebasehelper/output_tools/html_output_tool.py
```diff
@@ -0,0 +1,369 @@
...
+
+        :param dir_path: path of a directory to be searched
+        :return: dictionary containing the directory tree
+        """
+        dir_content = os.listdir(dir_path)
+
+        for item in dir_content:
+            path = os.path.join(dir_path, item)
+
+            if os.path.isdir(path) and 'git' not in item:
```
nforro: This should be `.git`, not `git`.

rebasehelper/output_tool.py
```diff
@@ -19,216 +19,177 @@
...
+        :param app: Application instance
         """
-        raise NotImplementedError()
+        cls.app = app
+        cls.print_patches_cli()
 
+        cls.colored_output(yellow, '\nAvailable logs:')
+        print('{0}:\n{1}'.format('Debug log', app.debug_log_file))
+        if results_store.get_old_build() is not None:
+            print('{0}:\n{1}'.format('Old build logs and (S)RPMs', os.path.join(app.results_dir, 'old_build')))
```
nforro: This should be `old-build` now.

rebasehelper/output_tool.py
```diff
@@ -19,216 +19,177 @@
...
+        if results_store.get_new_build() is not None:
+            print('{0}:\n{1}'.format('New build logs and (S)RPMs', os.path.join(app.results_dir, 'new_build')))
```
nforro: Same as above, this should be `new-build` now.

nforro:
BTW, the first line of commit message should be heading, summary of all changes in the commit, not the first item of a list of changes.

rebasehelper/cli.py
```diff
@@ -151,6 +152,13 @@ def build_parser():
             help="continue previously interrupted rebase"
         )
         parser.add_argument(
+            "--colors",
```
nforro: This should be rather `color`, to stay consistent with other tools using argument like this.

rebasehelper/cli.py
```diff
@@ -240,6 +248,17 @@ def __getattr__(self, name):
 class CliHelper(object):
 
     @staticmethod
+    def set_colors(app):
+        """Set app.conf.colors variable"""
+        if not os.isatty(1) or os.environ.get('TERM') == 'dumb'\
```
nforro: How about using `sys.stdout.fileno()` instead of `1`? It doesn't always need to be like that.

rebasehelper/cli.py
```diff
@@ -240,6 +248,17 @@ def __getattr__(self, name):
...
+           or (sys.platform.startswith('java')):
```
nforro: This is definitely wrong.

rebasehelper/cli.py
```diff
@@ -240,6 +248,17 @@ def __getattr__(self, name):
 class CliHelper(object):
 
     @staticmethod
+    def set_colors(app):
```
nforro: I'd call this method differently, something like `colorize_output` or `should_use_colors`.

rebasehelper/cli.py
```diff
@@ -240,6 +248,17 @@ def __getattr__(self, name):
...
+        """Set app.conf.colors variable"""
```
nforro: The docstring says something different than the method actually does.

rebasehelper/cli.py
nforro: I think it would be more logical to first check value of `app.conf.colors`, and only if it's `auto`, then continue with checking other things. I also think it'd be nice to honor `PY_COLORS` environment variable.

rebasehelper/cli.py
```diff
@@ -249,10 +268,11 @@ def run():
                 raise RebaseHelperError('Wrong format of --builder-options. It must be in the following form:'
                                         ' --builder-options="--desired-builder-option".')
             cli = CLI()
-            execution_dir, results_dir, debug_log_file, report_log_file = Application.setup(cli)
+            execution_dir, results_dir, debug_log_file = Application.setup(cli)
             if not cli.verbose:
                 handler.setLevel(logging.INFO)
-            app = Application(cli, execution_dir, results_dir, debug_log_file, report_log_file)
+            app = Application(cli, execution_dir, results_dir, debug_log_file)
+            app.conf.colors = CliHelper.set_colors(app)
```
nforro: You definitely shouldn't redefine the attribute. You should use the method in some `print` or rather `sys.stdout.write` wrapper.

rebasehelper/output_tool.py
```diff
@@ -19,216 +19,177 @@
...
 
     DEFAULT = False
+    NAME = 'name'
 
     @classmethod
-    def match(cls, cmd=None, *args, **kwargs):
-        """Checks if tool name matches the desired one."""
-        raise NotImplementedError()
+    def get_report_path(cls, app):
+        return os.path.join(app.results_dir, 'report.' + cls.NAME)
```
nforro: I don't like this very much. For example `text` output tool should definitely be named `text`, but the generated file I think should rather have standard `.txt` extension.

rebasehelper/output_tools/text_output_tool.py
```diff
@@ -0,0 +1,154 @@
...
+            logger_report.info(results['success'])
+        else:
+            logger_report.info(results['fail'])
+
+    @classmethod
+    def print_changes_patch(cls):
+        """Print info about the location of changes.patch"""
+        patch = cls.results_store.get_changes_patch()
+        if patch is not None:
+            logger_report.info('\nPatch containing changes between old and new version patches:')
```
nforro: This seems a bit weird, how about `Patch with differences between old and new version source files`?

rebasehelper/output_tool.py
```diff
@@ -19,216 +19,177 @@
...
+            print('{0}:\n{1}'.format('New build logs and (S)RPMs', os.path.join(app.results_dir, 'new-build')))
+        print()
 
-@register_output_tool
-class TextOutputTool(BaseOutputTool):
+        cls.colored_output(yellow, '%s:' % 'Rebased sources')
+        print("%s" % app.rebased_sources_dir)
 
-    """ Text output tool. """
+        cls.colored_output(yellow, '%s:' % 'Patch containing changes')
```
nforro: Like in the report file, wouldn't something like `Patch with differences` sound a bit better? Or maybe just `Generated patch`.

rebasehelper/application.py
```diff
@@ -584,16 +585,16 @@ def build_packages(self):
...
                             raise RebaseHelperError("Build failed, but no issues were found in the build log %s",
-                                                    build_log)
+                                                    build_log, logfile=build_log)
                     self.rebase_spec_file.modify_spec_files_section(files)
 
                 if not self.conf.non_interactive:
-                        msg = 'Do you want rebase-helper to try build the packages one more time'
-                        if not ConsoleHelper.get_message(msg):
-                            raise KeyboardInterrupt
+                    msg = 'Do you want rebase-helper to try build the packages one more time'
```
nforro: There should be `to try to build` in this sentence, right?

rebasehelper/exceptions.py
```diff
@@ -26,7 +26,21 @@ class RebaseHelperError(Exception):
...
     catching some expected and well known exception/error.
     """
-    pass
+
+    def __init__(self, *args, **kwargs):
+        """
+        Init method for RebaseHelperError exception
+
+        :param args: error message
+        :param kwargs: more optional arguments
```
nforro: The docstring is not very accurate. `Init method for RebaseHelperError exception` - `Constructor of RebaseHelperError` sounds better to me. I wouldn't describe `args` and `kwargs` at all perhaps, or if so, I'd use `error message plus optional formatting arguments` for `args`, and `keyword arguments` for `kwargs`, with added description of `logfile`.

rebasehelper/application.py
```diff
@@ -621,7 +622,7 @@ def run_package_checkers(self, results_dir):
 
         for checker_name in self.conf.pkgcomparetool:
             try:
-                results[checker_name] = checkers_runner.run_checker(results_dir, checker_name)
+                results[checker_name] = checkers_runner.run_checker(os.path.join(results_dir, 'checkers'), checker_name)
```
nforro: This line is a bit too long. It fits into the 120 character limit, but it doesn't fit github diff visualizer. Perhaps we should make the limit a bit more strict, say 100 or 110 characters.

rebasehelper/application.py
```diff
@@ -739,8 +755,8 @@ def run(self):
                 options_used.append('--builds-nowait')
             if options_used:
                 raise RebaseHelperError("%s can be used only with the following build tools: %s",
-                                        ' and '.join(options_used),
-                                        ', '.join(tools_creating_tasks)
+                                         ' and '.join(options_used),
+                                         ', '.join(tools_creating_tasks)
```
nforro: Why the extra spaces?

nforro: Also it would be nice to make this consistent, use either just `"` or `'`.

rebasehelper/exceptions.py
```diff
@@ -26,7 +26,21 @@ class RebaseHelperError(Exception):
...
+        """
+        if not args:
+            self.msg = None
+        elif len(args) > 1:
+            self.msg = args[0] % args[1:]
+        else:
+            self.msg = args[0]
+        self.logfile = kwargs.get('logfile')
```
nforro: Also, since there is now `RebaseHelperError.msg` attribute, you could use it in [`CliHelper`](https://github.com/rebase-helper/rebase-helper/blob/master/rebasehelper/cli.py#L259-L263).

rebasehelper/cli.py
```diff
@@ -151,6 +152,13 @@ def build_parser():
             help="continue previously interrupted rebase"
         )
         parser.add_argument(
+            "--color",
+            default='auto',
+            dest='color',
+            choices=['always', 'never', 'auto'],
+            help="use colors for CLI report"
```
nforro: I'd use `colorize the output, defaults to %(default)s`.
Later we can add support for colorizing also other messages, not just the "CLI report".

rebasehelper/output_tool.py
```diff
@@ -19,216 +19,200 @@
...
+    def should_use_colors(cls):
+        """Determine whether ansi colors should be used for CLI output"""
+        if os.environ.get('PY_COLORS') == '1':
+            return True
+        if os.environ.get('PY_COLORS') == '0':
+            return False
+        if cls.app.conf.color == 'auto':
+            if (not os.isatty(sys.stdout.fileno()) or
+                    os.environ.get('TERM') == 'dumb' or
+                    sys.platform.startswith('java')):
```
nforro: This is not supposed to be here. In the original check this is used in conjunction with NT platform, which we don't support anyway.

rebasehelper/output_tool.py
```diff
@@ -19,216 +19,200 @@
...
+
+        if patch_type in patches:
+            print('\n%s patches:' % patch_type)
+            for patch in patches[patch_type]:
+                cls.colored_output(color, patch)
+
 
     @classmethod
-    def print_summary(cls, path, results=results_store):
+    def colored_output(cls, color, string):
```
nforro: I'd rather see some generic helper function or method for printing colored or non-colored output. That way it could be used in more places than just in `TextOutputTool`.

rebasehelper/utils.py
```diff
@@ -95,6 +97,36 @@ class ConsoleHelper(object):
 
     """Class for command line interaction with the user."""
 
+    use_colors = False
+
+    @classmethod
+    def should_use_colors(cls, app):
```
nforro: You are setting `cls.use_colors` attribute where previously there were **return** statements. This means after you set the attribute the execution continues, which can have undesired effects. I think for simplicity and code readability best would be to revert to returning a value and assigning it to `ConsoleHelper.use_colors` outside of the method.

rebasehelper/utils.py
```diff
@@ -95,6 +97,36 @@ class ConsoleHelper(object):
...
+        if app.conf.color == 'auto':
+            if (not os.isatty(sys.stdout.fileno()) or
+                    os.environ.get('TERM') == 'dumb'):
+                cls.use_colors = False
+        elif app.conf.color == 'never':
+            cls.use_colors = False
+        cls.use_colors = True
+
+    @classmethod
+    def colored_output(cls, color, string):
```
nforro: I'd call this method differently, maybe just `print`, set `color` as a keyword argument, by default unset, and probably used string type to avoid importing color functions in other modules. You can use this approach for instance:
```python
import colors

if color and hasattr(colors, color):
    print(getattr(colors, color)(string))
else:
    print(string)
```

rebasehelper/cli.py
```diff
@@ -249,10 +257,11 @@ def run():
                 raise RebaseHelperError('Wrong format of --builder-options. It must be in the following form:'
                                         ' --builder-options="--desired-builder-option".')
             cli = CLI()
-            execution_dir, results_dir, debug_log_file, report_log_file = Application.setup(cli)
+            execution_dir, results_dir, debug_log_file = Application.setup(cli)
             if not cli.verbose:
                 handler.setLevel(logging.INFO)
-            app = Application(cli, execution_dir, results_dir, debug_log_file, report_log_file)
+            app = Application(cli, execution_dir, results_dir, debug_log_file)
+            ConsoleHelper.should_use_colors(app)
```
nforro: I think this should be called sooner, perhaps even in `CLI` constructor. We might want to print some colorized messages in `Application.setup` or so.

rebasehelper/utils.py
nforro: Also, I don't like `string`, `message` sounds better to me.

nforro: And one more thing, this should of course be in a separate commit ;-)

rebasehelper/output_tool.py
```diff
@@ -19,216 +19,166 @@
...
-        This will be used by checkers
+        Print report of the rebase
+
+        :param app: Application instance
         """
-        raise NotImplementedError()
+        cls.app = app
+        cls.print_patches_cli()
 
+        ConsoleHelper.cprint('\nAvailable logs:', 'yellow')
```
nforro: I'd prefer `ConsoleHelper.cprint('\nAvailable logs:', color='yellow')`. What do you think?

rebasehelper/output_tool.py
```diff
@@ -19,216 +19,166 @@
...
+            elif results_store.get_patches()['success']:
+                ConsoleHelper.cprint("\nPatching failed", color='red')
 
     @classmethod
-    def print_message_and_separator(cls, message="", separator='='):
-        logger_report.info(message)
-        logger_report.info(separator * len(message))
+    def print_report_file_path(cls):
+        """Print path to the report file"""
+        ConsoleHelper.cprint('%s report' % cls.NAME, color='yellow')
```
nforro: This doesn't look good, there is no `:` at the end.

skisela: I shall add one for each section, as each misses it.

rebasehelper/application.py
```diff
@@ -678,11 +680,26 @@ def get_rebased_patches(self):
...
+    def print_summary(self, exception=None):
+        """
+        Save rebase-helper result and print the summary using output_tools_runner
+        :param exception: Error message from rebase-helper
+        :return:
+        """
+        log = None
+        # Store rebase helper result exception
+        if exception:
+            if os.path.exists(exception.logfile):
```
nforro: What if `exception.logfile` is `None`?

rebasehelper/application.py
```diff
@@ -678,11 +680,27 @@ def get_rebased_patches(self):
...
+        """
+        Save rebase-helper result and print the summary using output_tools_runner
+        :param exception: Error message from rebase-helper
+        :return:
+        """
+        log = None
+        # Store rebase helper result exception
+        if exception:
+            if exception.logfile is not None:
+                if os.path.exists(exception.logfile):
```
nforro: This could be written as oneliner: `if exception.logfile and os.path.exists(exception.logfile):`. But if you prefer to keep it on two lines, I'm okay with that.

nforro:
I guess this resolves also issue https://github.com/rebase-helper/rebase-helper/issues/292, right?

skisela:
Yes, and resolves #222 as well.

PR #274, opened 2017-06-07
Add functional tests

### LINKED ISSUE #270: Interactive question in non-interactive mode
When `rpmbuild` build tool is selected and some build dependencies are missing, `rebase-helper` always asks the following question, even in case `--non-interactive` argument was specified:

`Some build dependencies are missing. Do you want to install them now [Y/n]?`

nforro:
Note: I used `GitPython` for cloning dist-git repository, but I can also implement the missing functionality into `GitHelper` and use it instead.

Fixes #270.


skisela:
Ahoj,
vadilo by, ak by to s tym review pockalo do pondelka? Chcel by som si to
viac prakticky skusit, dnes som na to nemal velmi vela casu.

Sebastian.

rebasehelper/tests/functional/test_rebase.py
```diff
@@ -0,0 +1,107 @@
...
+        execution_dir, results_dir, debug_log_file, report_log_file = Application.setup(cli)
+        app = Application(cli, execution_dir, results_dir, debug_log_file, report_log_file)
+        app.run()
+        with open(os.path.join(REBASE_HELPER_RESULTS_DIR, 'json.txt')) as f:
+            report = json.load(f)
+            for k in ['deleted', 'modified', 'inapplicable']:
+                assert set(report['patches'].get(k, [])) == (patches[k] or set())
+        changes = os.path.join(REBASE_HELPER_RESULTS_DIR, 'changes.patch')
+        patch = unidiff.PatchSet.from_filename(changes, encoding='UTF-8')
+        assert patch[0].path == '{}.spec'.format(package)
```
skisela: I think the order of changed files depends on the alphabetical sort, which leaves spec files at the end of the list most of the time, instead of at the beginning. But you have surely found that out in the build log.

nforro: Yes, I know about that. I'll modify the code to iterate over all hunks in the patch and search for the right one (related to SPEC and containing change of Version tag).

PR #278, opened 2017-06-16
Add possibility to download old build from Koji

nforro:
Added new argument `--get-old-build-from-koji`, only available with **python 2** due to `koji` module still lacking **python 3** support. Hopefully not for long, see: https://pagure.io/koji/c/4f4639ec


nforro:
I think most of the code is self-explanatory, but it's true that the new `KojiHelper` methods lack docstrings. I'll fix that.

PR #283, opened 2017-06-22
Delete patch tags of deleted patches from spec file

skisela:
No need for commenting now, when all changes are already
displayed in the changes patch.

Delete corresponding test.

### DIFF
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -447,7 +447,7 @@ class Application(object):
         Initialize git repository in the rebased directory
         :return: git.Repo instance of rebased_sources
         """
-        for patch in patches['applied']:
+        for patch in patches['applied'] + patches['not_applied']:
             shutil.copy(patch.path, rebased_sources_dir)
 
         repo = git.Repo.init(rebased_sources_dir)
--- a/rebasehelper/specfile.py
+++ b/rebasehelper/specfile.py
@@ -331,23 +331,35 @@ class SpecFile(object):
         """
         return self.patches['not_applied']
 
-    def _comment_out_patches(self, patch_num, note=None):
+    def _process_patches(self, comment_out=[], remove_patches=[], disable_inapplicable_patches=None):
         """
-        Comment out patches from SPEC file
+        Comment out and delete patches from SPEC file
 
-        :var patch_num: list with patch numbers to comment out
+        :var comment_out: list with patch numbers to comment out
+        :var remove_patches: list with patch numbers to delete
+        :var disable_inapplicable_patches: boolean value deciding if the inapplicable patches should be commented out
         """
         for index, line in enumerate(self.spec_content):
             #  if patch is applied on the line, try to check if it should be commented out
             if line.startswith('%patch'):
                 #  check patch numbers
-                for num in patch_num:
+                for num in comment_out:
                     #  if the line should be commented out
                     if line.startswith('%patch{0}'.format(num)):
-                        comment = '# {}\n'.format(note) if note is not None else ''
-                        self.spec_content[index] = '{}#%{}'.format(comment, line)
+                        comment = '# Following patch contains conflicts\n'
+                        if disable_inapplicable_patches:
+                            self.spec_content[index] = '{}#%{}'.format(comment, line)
+                        else:
+                            self.spec_content[index] = '{}{}'.format(comment, line)
                         #  remove the patch number from list
-                        patch_num.remove(num)
+                        comment_out.remove(num)
+                        break
+                for num in remove_patches:
+                    #  if the line should be removed
+                    if line.startswith('%patch{0}'.format(num)):
+                        self.spec_content[index] = ''
+                        #  remove the patch number from list
+                        remove_patches.remove(num)
                         break
 
     def update_paths_to_patches(self):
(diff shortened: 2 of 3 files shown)


rebasehelper/specfile.py
```diff
@@ -410,9 +389,13 @@ def write_updated_patches(self, patches, disable_inapplicable):
                     patch_inapplicable = [x for x in patches['inapplicable'] if patch_name in x]
                 else:
                     patch_inapplicable = None
-                if check_not_applied or patch_removed:
+                if check_not_applied:
                     self.spec_content[index] = '#{0} {1}\n'.format(' '.join(fields[:-1]),
                                                                    os.path.basename(patch_name))
```
nforro: I'm not sure if it is necessary to keep unused patches in the spec file. Anyway, it would be nice to preserve whitespaces, so it is obvious from the diff that the line in question was not changed, only commented out.

rebasehelper/specfile.py
```diff
@@ -331,7 +331,7 @@ def get_not_used_patches(self):
         """
         return self.patches['not_applied']
 
-    def _comment_out_patches(self, patch_num, note=None):
+    def _comment_out_patches(self, patch_num, note=None, remove_patches=False):
```
nforro: I think this method and its arguments should be changed to something more sensible. Perhaps something like `_process_patches(self, comment_out=[], remove=[], note=None)` with lists of patches to comment-out and/or remove?

rebasehelper/specfile.py
skisela: Well it is questionable, as that information might be gathered from the changes patch. In any case, rebased spec file in that state will probably not be used most of the time, as another interactive rebase will need to be done. Do you agree?

nforro: I think you are confusing `not applied` with `inapplicable` patches. AFAIK the former are the patches that have `PatchN` tag in the spec file but are lacking matching `%patchN` macro.
In my opinion these tags should be either removed or rather commented-out, but with a note justifying the reason of that.

rebasehelper/specfile.py
skisela: What do you exactly mean by comment_out list variable? I guess the plan is to either delete(deleted) or leave untouched(modified/applied).

rebasehelper/specfile.py
skisela: You are right, my bad. Then of course, let's delete `not applied` patches as well.

rebasehelper/specfile.py
nforro: Well, the method is called `_comment_out_patches` and with `remove_patches=False` it does exactly that.

nforro: And it is necessary for commenting out inapplicable patches if `--disable-inapplicable-patches` option is set.

rebasehelper/specfile.py
```diff
@@ -331,23 +331,33 @@ def get_not_used_patches(self):
         """
         return self.patches['not_applied']
 
-    def _comment_out_patches(self, patch_num, note=None):
+    def _process_patches(self, comment_out=[], remove_patches=[], note=None, disable_inapplicable_patches=None):
         """
-        Comment out patches from SPEC file
+        Comment out and delete patches from SPEC file
 
         :var patch_num: list with patch numbers to comment out
```
nforro: It seems the docstring is not updated.

rebasehelper/specfile.py
```diff
@@ -331,23 +331,33 @@ def get_not_used_patches(self):
...
             if line.startswith('%patch'):
                 #  check patch numbers
-                for num in patch_num:
+                for num in comment_out:
                     #  if the line should be commented out
                     if line.startswith('%patch{0}'.format(num)):
                         comment = '# {}\n'.format(note) if note is not None else ''
-                        self.spec_content[index] = '{}#%{}'.format(comment, line)
+                        if disable_inapplicable_patches:
+                            self.spec_content[index] = '{}#{}'.format(comment, line)
```
nforro: The `%` character was there for a reason - to escape macro in comment. Without doing that `rpmlint` will complain with `macro-in-comment` warning.

rebasehelper/specfile.py
```diff
@@ -331,23 +331,33 @@ def get_not_used_patches(self):
...
+                        else:
+                            self.spec_content[index] = '{}{}'.format(comment, line)
                         #  remove the patch number from list
-                        patch_num.remove(num)
+                        comment_out.remove(num)
+                        break
+                for num in remove_patches:
+                    #  if the line should be commented out
```
nforro: The comment should be `if the line should be removed` I guess.

rebasehelper/specfile.py
```diff
@@ -331,23 +331,33 @@ def get_not_used_patches(self):
...
+                    if line.startswith('%patch{0}'.format(num)):
+                        self.spec_content[index] = ''
```
nforro: This will keep blank line in the spec file, right?

rebasehelper/specfile.py
```diff
@@ -410,14 +395,16 @@ def write_updated_patches(self, patches, disable_inapplicable):
...
                 else:
                     patch_inapplicable = None
-                if check_not_applied or patch_removed:
-                    self.spec_content[index] = '#{0} {1}\n'.format(' '.join(fields[:-1]),
-                                                                   os.path.basename(patch_name))
+                if patch_inapplicable or patch_removed:
+                    if disable_inapplicable:
+                        self.spec_content[index] = '#{0} {1}\n'.format(' '.join(fields[:-1]),
+                                                                       os.path.basename(patch_name))
+                    inapplicable_patches.append(patch_num)
```
nforro: This doesn't seem right, it will append removed patches to `inapplicable_patches` list.
Also `check_not_applied` is now ignored completely.

rebasehelper/specfile.py
```diff
@@ -427,10 +414,9 @@ def write_updated_patches(self, patches, disable_inapplicable):
                     self.spec_content[index] = ' '.join(fields) + '\n'
                     modified_patches.append(patch_num)
 
-        self._comment_out_patches(removed_patches)
-        note = 'Disabled because of conflicts, please verify'
-        self._comment_out_patches(inapplicable_patches, note=note)
-        self._correct_rebased_patches(modified_patches)
+        note = 'Following patch contains conflicts'
+        self._process_patches(inapplicable_patches, removed_patches, note, disable_inapplicable)
```
nforro: I think you can move the `note` constant into the `_process_patches` method, it's pointless to keep it here now.

rebasehelper/specfile.py
```diff
@@ -331,23 +331,37 @@ def get_not_used_patches(self):
...
+
         for index, line in enumerate(self.spec_content):
             #  if patch is applied on the line, try to check if it should be commented out
             if line.startswith('%patch'):
                 #  check patch numbers
-                for num in patch_num:
+                for num in comment_out:
                     #  if the line should be commented out
                     if line.startswith('%patch{0}'.format(num)):
                         comment = '# {}\n'.format(note) if note is not None else ''
```
nforro: Since `note` is now constant, it would be nicer to use the text directly in `comment` variable.

rebasehelper/application.py
```diff
@@ -450,6 +450,9 @@ def _prepare_rebased_repository(cls, patches, rebased_sources_dir):
         for patch in patches['applied']:
             shutil.copy(patch.path, rebased_sources_dir)
 
+        for patch in patches['not_applied']:
```
nforro: How about joining the two loops: `for patch in patches['applied'] + patches['not_applied']:`?

Issue #285, opened 2017-06-30
Installing rebase-helper on non-system Python


junaruga:
May I ask you a question about installing `rebase-helpr`?

I am trying to install `rebase-helpr` (latest master branch) on Python 3.6.1 installed from source code on Fedora 25.

```
$ pwd
/home/jaruga/git/rebase-helper

$ which python3
/usr/local/python-3.6.1/bin/python3

$ sudo /usr/local/python-3.6.1/bin/pip3 install .

$ which rebase-helper
/usr/local/python-3.6.1/bin/rebase-helper

$ rebase-helper --help
Traceback (most recent call last):
  File "/usr/local/python-3.6.1/bin/rebase-helper", line 11, in <module>
    load_entry_point('rebasehelper==0.9.0', 'console_scripts', 'rebase-helper')()
  File "/usr/local/python-3.6.1/lib/python3.6/site-packages/pkg_resources/__init__.py", line 565, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/local/python-3.6.1/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2631, in load_entry_point
    return ep.load()
  File "/usr/local/python-3.6.1/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2291, in load 
    return self.resolve()
  File "/usr/local/python-3.6.1/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2297, in resolve 
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/local/python-3.6.1/lib/python3.6/site-packages/rebasehelper/cli.py", line 30, in <module>
    from rebasehelper.application import Application
  File "/usr/local/python-3.6.1/lib/python3.6/site-packages/rebasehelper/application.py", line 28, in <module>
    import git 
ModuleNotFoundError: No module named 'git

$ sudo /usr/local/python-3.6.1/bin/pip3 install . -r requirements.txt

Traceback (most recent call last):
  File "/usr/local/python-3.6.1/bin/rebase-helper", line 11, in <module>
    load_entry_point('rebasehelper==0.9.0', 'console_scripts', 'rebase-helper')()
  File "/usr/local/python-3.6.1/lib/python3.6/site-packages/pkg_resources/__init__.py", line 565, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
[...]

PR #286, opened 2017-06-30
Update a comment for old package rpm-python in requirements.txt

junaruga:
This PR is to update `requirements.txt` correctly, and from below conversation.

> https://github.com/rebase-helper/rebase-helper/issues/285#issuecomment-312264180
> it used to be called rpm-python3, we should probably update the comment in requirements.txt

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,6 @@
-# This can not be installed from pip, on Fedora you need to install rpm-python{,3}
+# This can not be installed from pip, on Fedora you need to install
+# python{2,3}-rpm package containing the module.
+# See: https://github.com/rpm-software-management/rpm/tree/master/python/rpm
 #rpm-python
 backports.lzma
 copr


requirements.txt
```diff
@@ -1,5 +1,6 @@
-# This can not be installed from pip, on Fedora you need to install rpm-python{,3}
-#rpm-python
+# This can not be installed from pip, on Fedora you need to install python3-rpm.
+# https://github.com/rpm-software-management/rpm/tree/master/python/rpm
+#python3-rpm
```
nforro: Don't forget about python2 version of the package, previously called `rpm-python`, now `python2-rpm`.
Also, since `python3-rpm` provides `rpm3-python` and `python2-rpm` provides `rpm-python`, `dnf install rpm-python` will work on F24 as well as on F25 and F26, but `dnf install python2-rpm` won't work on F24. So I think it would be better to wait until F24 EOL before doing this change.

junaruga: OK I will wait until F24 EOL. maybe we need to wait for 1 month.

> https://fedoraproject.org/wiki/Fedora_Release_Life_Cycle
> Fedora 24 will be maintained until 1 month after the release of Fedora 26. 

By the way I could not find `rpm3-python` pip package on my F25.

```
$ rpm -q python2-rpm
python2-rpm-4.13.0.1-1.fc25.x86_64

$ rpm -q python3-rpm
python3-rpm-4.13.0.1-1.fc25.x86_64

$ /usr/bin/pip list | grep rpm
rpm-python (4.13.0.1)

$ /usr/bin/pip3 list | grep rpm
rpm-python (4.13.0.1)
```

nforro: Sorry, that's a typo, it should have been `rpm-python3`.

junaruga:
By the way, the pip package name `rpm-python` will be renamed to `rpm` in the future.
At least we can see it on the upstream.
https://github.com/rpm-software-management/rpm/commit/2797d00

requirements.txt
junaruga: @nforro 

> So I think it would be better to wait until F24 EOL before doing this change.

Now F24 is end of life.
https://fedoraproject.org/wiki/End_of_life

requirements.txt
```diff
@@ -1,4 +1,6 @@
-# This can not be installed from pip, on Fedora you need to install rpm-python{,3}
+# This can not be installed from pip, on Fedora you need to install
+# python{2,3}-rpm RPM including rpm-python pip package.
```
nforro: According to your [comment](https://github.com/rebase-helper/rebase-helper/pull/286#issuecomment-312655923), I think it would be best to say something like: `This can not be installed from pip, on Fedora you need to install python{2,3}-rpm package containing the module, see:` followed by the link.
And after the module is renamed later we will change `#rpm-python` to just `#rpm`.

junaruga: @nforro  Thanks for the review.
Updated as your way. 
If it looks good, please merge with squash to 1 commit.

nforro: Squash merging is disabled for this project. To be honest I'm not sure why, but anyway it would be better if you rebased your branch to master and squashed the two commits yourself. Note that there is a typo in the title (`Imporve`), and also the commit message
`Update RPM package name python-rpm to python2-rpm in requirements.txt`
should most likely be modified like this:
`Update RPM package name rpm-python{,3} to python{2,3}-rpm in requirements.txt`

junaruga: Thanks for the checking again. I fixed the commit message. And rebased by myself now.

junaruga: A just moment. I have to rebase on the latest master branch.

junaruga: OK done.

PR #288, opened 2017-07-04
Update KojiHelper for Koji 1.13

nforro:
Koji 1.13 finally supports python3, but it's not packaged for Fedora yet, so we should probably wait before merging this.


nforro:
Here are the Fedora updates in question:
F25: https://bodhi.fedoraproject.org/updates/FEDORA-2017-e719edea40
F26: https://bodhi.fedoraproject.org/updates/FEDORA-2017-f941efd84c

PR #299, opened 2017-08-07
Revert mock and curl workarounds

nforro:
The workarounds should no longer be needed as soon as these two updates hit stable:
https://bodhi.fedoraproject.org/updates/FEDORA-2017-4ee5b66a4c
https://bodhi.fedoraproject.org/updates/FEDORA-2017-204acb9aa4


nforro:
Yes, I know, but this PR will be deprecated by #305, where the workarounds are already removed.

PR #300, opened 2017-08-11
Do not wait for non existent koji build

skisela:
Skip waiting for old koji build,
when --builds-nowait and
--get-old-build-from-koji is used.

Fix checking for build_dict['builds_nowait']
while not being set.


nforro:
Actually, using both `--get-old-build-from-koji` and `--builds-nowait` doesn't make sense in the current implementation. If you run `rebase-helper` with both options, it downloads the old build and starts the new. Downloading the old build should be of course done on the second run with `--build-tasks`, and `rebase-helper` should tell you ID of the old build if finds on `koji`, so you can pass it along with the new build ID.

nforro:
Hm, that would obviously work only with `koji` build tool.

rebasehelper/cli.py
```diff
@@ -90,14 +90,15 @@ def build_parser():
             action="store_true",
             help="be more verbose (recommended)"
         )
-        parser.add_argument(
+        patch_build_only_group = parser.add_mutually_exclusive_group()
```
nforro: Why call this variable `patch_build_only_group` while a little bit lower similar variable is called simply `group`?

nforro:
I like the idea with mutually exclusive arguments, but I think in case of `--builds-nowait` and `--get-old-build-from-koji` it would be better to do the check in `Application.run` method, where a different check for `--builds-nowait` is already present.

Also, `--comparepkgs-only` should be also mutually exclusive with `--patch-only` and `--build-only`.

nforro:
> Show build ID of the last koji build to the user.

I think this should go into a separate commit.

skisela:
2 checks for `--builds-nowait` are present, neither is suitable in my opinion, as one is performed only in case wrong `--buildtool` is used and the other is run too late - after building packages. I added a new one.

Yes the group for mutually exclusive parameters requires better naming, but the problematic one was removed due to moving the check to `the Application.run`.

skisela:
Btw. Do you prefer having the arguments in 'CLI.build_parser' ordered alphabetically or logically together? 

I grouped the parameters within the same group of mutually exclusive arguments together.

nforro:
> Btw. Do you prefer having the arguments in CLI.build_parser ordered alphabetically or logically together?

I'd prefer both, first logically, then alphabetically. Current order seems pretty random.

rebasehelper/cli.py
```diff
@@ -90,20 +90,29 @@ def build_parser():
             action="store_true",
             help="be more verbose (recommended)"
         )
-        parser.add_argument(
+        patch_build_cmp_only_group = parser.add_mutually_exclusive_group()
```
nforro: I think simple `group` would suffice, since it is not referenced anywhere and so it doesn't have to be unique.

rebasehelper/cli.py
```diff
@@ -130,11 +139,11 @@ def build_parser():
...
-            "--keep-workspace",
-            default=False,
-            action="store_true",
-            help="do not remove workspace directory after finishing"
+            "sources",
+            metavar='SOURCES',
+            nargs='?',
+            default=None,
+            help="new upstream sources"
         )
```
nforro: I think `SOURCES` should be logically the last argument.

rebasehelper/cli.py
```diff
@@ -193,15 +195,17 @@ def build_parser():
             help=argparse.SUPPRESS
         )
```
nforro: I think we can safery remove `--fedpkg-build-tasks` now. It had been used with upstream monitoring and I don't think anyone else has ever used it.

nforro:
The order of arguments seems to be better now, however I think the change should go to separate commit, or at least the commit message should be adjusted.
We should try to avoid having unrelated (or not so-closely-related) changes in one commit. We can benefit from it later when we'll need to find out something in the history.

skisela:
You mean adjust the message title? Because reorder is mentioned in 9c47f28ba98af991dfc63aac0b6a42d538fb3c2f

nforro:
That's actually what I had in mind in https://github.com/rebase-helper/rebase-helper/pull/271#issuecomment-323995056. If a commit changes two things, the title should not be one of them and the rest the other one. The title should somehow summarize both changes, and the rest of the commit message should elaborate on that.

nforro:
Tests are failing because of [this check](https://github.com/rebase-helper/rebase-helper/blob/master/rebasehelper/application.py#L726-L730), which can now be removed.

skisela:
Yes, there were few reasons, this was one of them. Thanks!

nforro:
Sorry to nitpick again, but you are removing `parser.add_argument("--fedpkg-build-tasks", ...)` in **Reorder Rebase Helper arguments** (de458279196cce8acc8da04b4fa03ee7ed935a57), while I'd expect it in **Remove fedpkg-build-tasks argument** (f687700dddc344c9d8c31324ea22db38f3200a71).

Also it would make more sense to first reorder the arguments and then disable those mutually exclusive ones.

skisela:
Yes, you are right. Changing multiple things in a single commit may even create hunks that should have been in separate commits, which is inconvenient.

PR #239, opened 2016-11-01
Dockerfile for rebase-helper testing

phracek:
The Dockerfile can be used for rebase-helper testing and later on for rebases.

After deep testing we can include them in the-new-hotness daemon, but it is the future.

PR #305, opened 2017-08-17
Using rebase-helper Dockerfile.

junaruga:
I want to improve current `Makefile.docker` test task with creating `rebase-helper`'s Dockerfile.
The current code looks a little bit hard to manage.

```
      git archive HEAD | $(DOCKER) run \
              --cap-add=SYS_ADMIN --rm -i -v /build -w /build -e PY_COLORS=1 \
              -e TOXENV="$(TOXENV)" -e PYTEST_ADDOPTS="$(PYTEST_ADDOPTS)" $(IMAGE) \
              /bin/sh -c 'dnf -y install tar && tar -x && dnf -y upgrade openssl-libs && dnf -y install yum && dnf -y install $$(cat test-deps-fedora.txt) && tox'
```

The merits are 

- We can manage the logic such as `dnf install something ..` in the `Dockerfile`.
  We may be able to manage `test-deps-fedora.txt`'s content in the `Dockerfile` removing the `test-deps-fedora.txt` file.
- Users can test on their local by using `rebase-helper`'s docker image.
- We can manage simple `Makefile.docker`.

This PR is inspired by https://github.com/rpm-software-management/rpm/blob/master/ci/Dockerfile .
Other configurations in the file may inspires us.


junaruga:
I found that this PR might be conflict with another PR: https://github.com/rebase-helper/rebase-helper/pull/239

junaruga:
https://travis-ci.org/rebase-helper/rebase-helper/builds/265654685

The test was failed on fedora:25.
But then when I tried to run test on my local for fedora:25, it was succeeded. 
So, I guess that the test is unstable, and the failure is not related to this PR.

```
$ sed -r -i -e "/^FROM /s/FROM .*/FROM fedora:25/" Dockerfile

$ docker build -t rebase-helper .

$ docker run --cap-add=SYS_ADMIN -t rebase-helper
...
  py2: commands succeeded
  py3: commands succeeded
  congratulations :)
```

Makefile.docker
```diff
@@ -17,8 +17,10 @@ test: $(TEST_IMAGE)
 
 $(TEST_IMAGE):
 	$(eval IMAGE := $(subst _COLON_,:,$@))
-	$(DOCKER) pull $(IMAGE)
-	git archive HEAD | $(DOCKER) run \
-		--cap-add=SYS_ADMIN --rm -i -v /build -w /build -e PY_COLORS=1 \
-		-e TOXENV="$(TOXENV)" -e PYTEST_ADDOPTS="$(PYTEST_ADDOPTS)" $(IMAGE) \
-		/bin/sh -c 'dnf -y install tar && tar -x && dnf -y upgrade openssl-libs && dnf -y install yum && dnf -y install $$(cat test-deps-fedora.txt) && tox'
+	sed -r -i -e "/^FROM /s/FROM .*/FROM $(IMAGE)/" Dockerfile
```
nforro: I think it would be better to use [this approach](https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact) instead of sed.

Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM fedora
+
+WORKDIR /build
+COPY . .
+
+ENV PY_COLORS=1
+
+RUN dnf -y upgrade openssl-libs
+RUN dnf -y install yum $(cat test-deps-fedora.txt)
```
nforro: Wouldn't it be better to just remove `test-deps-fedora.txt` file and merge its contents into Dockerfile?

nforro:
The test is failing because of a bug in mock, see https://github.com/rebase-helper/rebase-helper/commit/bd2b0296b61532e6e08dd705c829b4bb5ee26b66 and https://github.com/rebase-helper/rebase-helper/pull/299.

Makefile.docker
junaruga: OH I did not know the approach! I will fix it. Thanks.

Dockerfile
junaruga: Yes, I strongly agree with your opinion, removing `test-deps-fedora.txt`.
Just in case I kept the file.
As you like removing the file, I will remove the file in this PR.

junaruga:
> The test is failing because of a bug in mock, see bd2b029 and #299.

OK thanks for the information.

nforro:
To workaround it you have to first install yum in one dnf session and then install other dependencies in another dnf session. But fixed mock package should be in F25 soon.

junaruga:
> To workaround it you have to first install yum in one dnf session and then install other dependencies in another dnf session. But fixed mock package should be in F25 soon.

OK. Thanks for the explanation.
I try to fix it in this PR to pass the tests.

nforro:
`.gitlab-ci.yml` file should probably look like this:

```yaml
stages:
    - test

.test_template: &test_definition
    stage: test
    image: docker:latest
    variables:
        DOCKER_DRIVER: overlay
    services: docker:dind
    script:
        - make PYTEST_ADDOPTS="-m 'not optional'" test-docker

test:f25:
    <<: *test_definition
    variables:
        DOCKER_IMAGE: fedora:25

test:latest:
    <<: *test_definition
    variables:
        DOCKER_IMAGE: fedora:latest

test:rawhide:
    <<: *test_definition
    variables:
        DOCKER_IMAGE: fedora:rawhide
```

But I'm not really sure whether this will work, let's see.

junaruga:
Do you want to change `.gitlab-ci.yml` in this PR too?
Is the Gitlab CI triggered by this PR?
How to check the `.gitlab-ci.yml` is correct in this PR?
https://gitlab.com/rebase-helper/rebase-helper/-/jobs

nforro:
Well, if you remove `test-deps-fedora.txt` file, it will stop working, so it would be good to prevent that.
The pipeline will be triggerred by this PR, but there will be most probably some delay.

junaruga:
> Well, if you remove test-deps-fedora.txt file, it will stop working, so it would be good to prevent that.
> The pipeline will be triggerred by this PR, but there will be most probably some delay.

OK I will modify `.gitlab-ci.yml` too in this PR.

junaruga:
> https://travis-ci.org/rebase-helper/rebase-helper/jobs/265966635
> Please provide a source image with `from` prior to commit

There is a error ARG before FROM in `Dockerfile`.
I am checking which version the feature is supported from.
https://github.com/moby/moby/issues/32457

nforro:
Hm, it's supported since `17.05.0-ce`, but docker on Travis is `17.03.1-ce`.

nforro:
This might help:
https://docs.travis-ci.com/user/docker/#Installing-a-newer-Docker-version

junaruga:
OK let me try to update docker to latest version.

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+RUN echo -e "deltarpm=0\ninstall_weak_deps=0\ntsflags=nodocs" >> /etc/dnf/dnf.conf
+RUN dnf -y update
+RUN dnf -y install yum
+RUN dnf -y install \
+  python2 \
+  python2-devel \
+  python3 \
+  python3-devel \
+  rpm-python \
+  rpm-python3 \
```
nforro: This should be `python2-rpm` and `python3-rpm`.

junaruga: ok I will fix it.

nforro: Also other `python-*` packages should be renamed to `python2-*`.

junaruga:
OK success for Travis CI.
So, we need to wait for Gitlab, right?

nforro:
> So, we need to wait for Gitlab, right?

Yes, it's currently stuck waiting for update and it's not even possible to force it manually. So I guess we will have to wait for several hours.

Dockerfile
junaruga: OK renamed to `python2-*` now.

junaruga: > https://travis-ci.org/rebase-helper/rebase-helper/builds/266157702
No package python2-pip available on only F25.

junaruga:
Which page can we check the result of Gitlab CI on?
I could not find it.

nforro:
Sorry, I was wrong. Your branch is not being synchronized to gitlab, as it is from different repo, so I cloned it and pushed as `junaruga-feature/add-dockerfile`, and it should be synchronized in a while hopefully.

junaruga:
Oh okay. Thanks for checking.

Makefile.docker
```diff
@@ -17,8 +17,12 @@ test: $(TEST_IMAGE)
 
 $(TEST_IMAGE):
 	$(eval IMAGE := $(subst _COLON_,:,$@))
-	$(DOCKER) pull $(IMAGE)
-	git archive HEAD | $(DOCKER) run \
-		--cap-add=SYS_ADMIN --rm -i -v /build -w /build -e PY_COLORS=1 \
-		-e TOXENV="$(TOXENV)" -e PYTEST_ADDOPTS="$(PYTEST_ADDOPTS)" $(IMAGE) \
-		/bin/sh -c 'dnf -y install tar && tar -x && dnf -y upgrade openssl-libs && dnf -y install yum && dnf -y install $$(cat test-deps-fedora.txt) && tox'
+	$(DOCKER) build \
+		--build-arg DOCKER_IMAGE="${DOCKER_IMAGE}" \
```
junaruga: Sorry this line `DOCKER_IMAGE` should be `IMAGE`.
By the way, there are `DOCKER_IMAGE`, `TEST_IMAGE`, and `IMAGE` in the file.
That is confusing me.

junaruga: `Makefile` is good for task, but not good for script logic.

Instead of current logic.

Makefile
```
test-docker:
    make -f Makefile.docker test
```

Below logic looks better, isn't it?

```
test-docker:
    scripts/test-on-docker.sh
```

nforro: I don't know, does it? What are the advantages of a script?

junaruga: > I don't know, does it? 

What is the meaning of "does it"?

The advantage of a script make code simple.

Actually the content of `Makefile.docker` is 3 statement.
If we use the bash shell script, we can write it on 5 lines.

1. if user does not belong to docker group, add "sudo".
2. docker build.
3. docker run. 

When we write .PHONY lone code target, we have to consider colon ":", as a result, the code line become longer.
So, long .PHONY target can be a scrip file.

But if you do not like it, I do not change this part.

junaruga: My code is using `${DOCKER_IMAGE}`, but `$(IMAGE)` or `$(DOCKER_IMAGE)` is correct?
I do not know the difference of `${FOO}` and `$(FOO)` in `Makefile`.

nforro: > My code is using ${DOCKER_IMAGE}, but $(IMAGE) or $(DOCKER_IMAGE) is correct?
> I do not know the difference of ${FOO} and $(FOO) in Makefile.

`$(IMAGE)` is correct, in Makefile you have to use `$()` syntax to reference variables. And `$(IMAGE)` is the current target (substituted from `$(TEST_IMAGE)`). 

I agree that it's confusing, I don't like the colon hack also. I'm not against a script, but with Makefile it should be able to run all jobs in parallel. But TBH I'm not sure if it really works.

nforro:
BTW I'm trying to get `.gitlab-ci.yml` to work, but with these delays between synchronizations it will take me some time. Here is your cloned branch with my commits on top:
https://gitlab.com/rebase-helper/rebase-helper/commits/junaruga-feature/add-dockerfile

Makefile.docker
junaruga: OK. In this PR I will just change to `${DOCKER_IMAGE}` to `$(IMAGE)`.
The parallel might be a merit of that.

junaruga:
> BTW I'm trying to get .gitlab-ci.yml to work, but with these delays between synchronizations it will take me some time. Here is your cloned branch with my commits on top:
https://gitlab.com/rebase-helper/rebase-helper/commits/junaruga-feature/add-dockerfile

Great! Thanks for the help.
I will do cherry-pick the your commit to my branch again. And squash it.

junaruga:
Ah maybe I should wait your work for `.gitlab-ci.yml`.

nforro:
I'd say forget about `.gitlab-ci.yml` for now, let's merge this PR without it. I'll deal with it later.

junaruga:
OK I will keep `.gitlab-ci.yml` without change in this PR.
Just moment.

nforro:
Tests are failing on rawhide because proper GPG key is missing. You can work it around by downloading and importing [RPM-GPG-KEY-fedora-28-primary](https://pagure.io/fedora-repos/raw/master/f/RPM-GPG-KEY-fedora-28-primary). We can remove such workaround later when fedora:rawhide image is fixed.

junaruga:
You mean you want to include the `Dockerfile` like below logic?

```
RUN curl -O https://pagure.io/fedora-repos/raw/master/f/RPM-GPG-KEY-fedora-28-primary
RUN mv RPM-GPG-KEY-fedora-28-primary /etc/pki/RPM-GPG-KEY-fedora-28-x86_64
RUN dnf -y update
```

nforro:
Yes, plus `rpm --import /etc/pki/RPM-GPG-KEY-fedora-28-x86_64`. And some comment about it being temporary workaround until `fedora:rawhide` image is updated with up-to-date `fedora-repos` package.

junaruga:
OK thanks let me try it.

junaruga:
@nforro we passed "dnf update and install" by the modification, but tests are failed.
https://travis-ci.org/rebase-helper/rebase-helper/builds/267104514

I have no idea. Do you have any idea?

nforro:
Yes, there is another bug:
https://bugzilla.redhat.com/show_bug.cgi?id=1483342

I guess we'll have to wait until next Rawhide compose, or install `dnf-2.6.3-3.fc28` directly from Koji as a workaround.

junaruga:
I want to wait :)

junaruga:
Can I remove the latest commit for temporary workaround in this PR?

```
+
+# Temporary workaround until fedora:rawhide image is updated with up-to-date
+# fedora-repos package.
+RUN curl -O https://pagure.io/fedora-repos/raw/master/f/RPM-GPG-KEY-fedora-28-primary
+RUN mv RPM-GPG-KEY-fedora-28-primary /etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-28-x86_64
+RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-28-x86_64
+
```

nforro:
I'm not sure, I guess not yet, because even if `fedora-repos-28-0.1` gets to the rawhide repo, it still cannot be installed because it is signed with the missing key.

junaruga:
OK, so let's wait right now :)

nforro:
BTW, this is how working `.gitlab-ci.yml` file should look like:

```yml
stages:
    - test

.test_template: &test_definition
    stage: test
    image: docker:latest
    variables:
        DOCKER_DRIVER: overlay
        DOCKER_HOST: tcp://docker:2375
    services:
        - docker:dind
    script:
        - apk update
        - apk add make git
        - make PYTEST_ADDOPTS="-m 'not optional'" test-docker

test:f25:
    <<: *test_definition
    variables:
        DOCKER_IMAGE: fedora:25

test:latest:
    <<: *test_definition
    variables:
        DOCKER_IMAGE: fedora:latest

test:rawhide:
    <<: *test_definition
    variables:
        DOCKER_IMAGE: fedora:rawhide
```

And the check in `Makefile.docker` has to be modified as follows:

```makefile
# sudo is needed in case user is not member of docker group
ifeq (,$(CI))
ifeq (,$(findstring docker,$(shell groups)))
        DOCKER := sudo docker
endif
endif
```

`CI` environment variable is defined in Gitlab CI and Travis CI environments, and this ensures sudo is not used in there.

Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+RUN echo -e "deltarpm=0\ninstall_weak_deps=0\ntsflags=nodocs" >> /etc/dnf/dnf.conf
+
+# Temporary workaround until fedora:rawhide image is updated with up-to-date
+# fedora-repos package.
+RUN curl -O https://pagure.io/fedora-repos/raw/master/f/RPM-GPG-KEY-fedora-28-primary
+RUN mv RPM-GPG-KEY-fedora-28-primary /etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-28-x86_64
+RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-28-x86_64
+
+RUN dnf -y update
+RUN dnf -y install yum
```
nforro: Fixed mock package is now in both F25 and F26, so you can remove this workaround.

junaruga: OK, I will remove it!

.travis.yml
```diff
@@ -13,4 +18,6 @@ before_script:
         else
             PYTEST_ADDOPTS="-m 'not optional'"
         fi
-script: make PYTEST_ADDOPTS="${PYTEST_ADDOPTS}" test-docker
+script:
+    - docker version
```
nforro: I think this is no longer needed, but maybe you can keep it and move it to `install`, so it doesn't show highlighted in the test output.

junaruga: The reason for `docker version` is to show the docker version that we updated manually.

In my memory, if the script in `install` section does show the stdout in the travis log file.
So, if you want to remove the line, I would just remove the line.

How do you think?

nforro: It does show, but collapsed and not highlighted in green like now:
**The command "docker version" exited with 0.**

I guess it could be sometimes useful to see what docker version has been used for testing, so I'd rather keep it.

nforro: On the other hand, the version is visible in apt-get output, so you can go ahead and just remove it.

junaruga: > On the other hand, the version is visible in apt-get output, so you can go ahead and just remove it.

Ah okay, I did not know that. I will remove the line.

junaruga:
Oh finally merged! Thanks!

PR #306, opened 2017-08-22
Fix rpmdiff and abipkgdiff checkers

### LINKED ISSUE #301: Check existence of rpm packages before rpmdiff
After this commit:
https://github.com/rpm-software-management/rpm/pull/180/commits/68d4b463859da0bb35740eb55d2b033c0e0d6bcc

debug-info subpackage is divided into debug-info and debugsource. 

Rpmdiff generates diff between old/new packages. Rebasing packages will fail, in case there will not be a corresponding debugsource/debuginfo package.

nforro:
Fixes #301.

* rpmdiff
    * Skip debuginfo and debugsource packages
    * Do not traceback when matching new version (sub)package doesn't exist
* abipkgdiff
    * Use `split_nevra` from `hawkey` module for reliable parsing of package name
    * Find proper debuginfo package for each (sub)package and continue even if it doesn't exist, as abipkgdiff works fine without it - it just shows bare symbols only

### DIFF
--- a/rebasehelper/checkers/abipkgdiff_tool.py
+++ b/rebasehelper/checkers/abipkgdiff_tool.py
@@ -24,7 +24,11 @@ from __future__ import print_function
 import os
 import re
 
-from rebasehelper.utils import ProcessHelper
+import rpm
+
+from hawkey import split_nevra
+
+from rebasehelper.utils import ProcessHelper, RpmHelper
 from rebasehelper.logger import logger
 from rebasehelper.exceptions import RebaseHelperError, CheckerNotFoundError
 from rebasehelper.results_store import results_store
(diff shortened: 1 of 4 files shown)


nforro:
I hate to introduce another Fedora-only dependency (`hawkey`), but I think it has more pros than cons. Parsing package name using regexps is unreliable, because the rules are really complex and there are many corner cases. `split_nevra` can be later used in more places than just abipkgdiff checker.

rebasehelper/checkers/abipkgdiff_tool.py
```diff
@@ -71,61 +75,56 @@ def _get_packages_for_abipkgdiff(cls, input_structure=None):
...
+        srpm = RpmHelper.get_info_from_rpm(pkg, rpm.RPMTAG_SOURCERPM)
+        debuginfo = '{}-debuginfo'.format(split_nevra(srpm).name)
+        find = [x for x in debug if split_nevra(os.path.basename(x)).name == debuginfo]
+        if find:
+            return find[0]
+        return None
+
+    @classmethod
     def run_check(cls, result_dir):
         """Compares old and new RPMs using pkgdiff"""
```
skisela: Not really related to this commit, but the comment could be fixed to the correct `abipkgdiff` name.

PR #307, opened 2017-08-23
Fix tests on Gitlab CI

nforro:
Run tests on Gitlab CI the same way as on Travis CI, using `make test-docker`.


nforro:
Here is the Gitlab CI [pipeline](https://gitlab.com/rebase-helper/rebase-helper/pipelines/11121531).

PR #309, opened 2017-08-24
Implement custom NEVRA parser

### LINKED ISSUE #308: Do not use hawkey module
According to https://github.com/rpm-software-management/libdnf/issues/321 hawkey is going to die soon.

nforro:
Remove dependency on `hawkey` module which is to be deprecated.

Resolves #308.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,6 @@ RUN dnf -y install \
   python3-devel \
   python2-rpm \
   python3-rpm \
-  python2-hawkey \
-  python3-hawkey \
   python2-tox \
   python3-tox \
   python-pip \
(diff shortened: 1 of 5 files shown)


nforro:
This implementation should actually work a bit better than `hawkey.split_nevra`, because the latter is capable of splitting only strings that match full `NEVRA` pattern.

skisela:
Yes that's great.

So the extra braces in the regex are to emphasize that fact I guess.
https://github.com/rebase-helper/rebase-helper/pull/309/files#diff-fe3e3d5f4c143113b3837700625080c8R711

nforro:
The braces are there so that the number of groups and their indexes stays the same no matter which regexp is used. The regexps are actually borrowed from [libdnf sources](https://github.com/rpm-software-management/libdnf/blob/59bc733148c36f9a85d62e01768488a8e2c2f97b/libdnf/hy-subject-private.c#L28-L34) (except for the third one, which is ambiguous and not really useful I'd say).

Issue #310, opened 2017-08-24
Crash when build of old version fails


nforro:
A list is assigned to the `logfile` argument here:
https://github.com/rebase-helper/rebase-helper/blob/c39f271b83f7af0723b6c5a7621f71bf9402a24f/rebasehelper/application.py#L561

Either `Application.print_summary` should be modified to be able to handle a list of logfiles, or only one logfile should be assigned.

Also, I'd drop `'N/A'` argument, IMHO it's better to let `logfile` to be `None`.

PR #312, opened 2017-08-25
Allow multiple logs for RebaseHelperError

skisela:
RebaseHelperError exception now allows multiple
logs to be stored within it.

Resolves: #310

### DIFF
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -558,7 +558,7 @@ class Application(object):
                         error_message = 'Building old RPM package failed. Check logs: {} '.format(
                             builder.get_logs().get('logs', 'N/A')
                         )
-                        raise RebaseHelperError(error_message, logfile=builder.get_logs().get('logs', 'N/A'))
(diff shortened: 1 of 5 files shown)


rebasehelper/application.py
```diff
@@ -567,14 +567,14 @@ def build_packages(self):
                         raise RebaseHelperError('Build log %s does not exist', build_log_path)
                     except BuildLogAnalyzerMakeError:
                         raise RebaseHelperError('%s during build. Check log %s', msg, build_log_path,
-                                                logfile=build_log_path)
+                                                logfiles=build_log_path)
```
nforro: I can see that it's not being used right now by neither `text` nor `json` output tool, but wouldn't it be better to pass it as a list, so you don't have to check type etc. After all, the argument is now called `logfiles` (plural).

skisela: The `logfiles` attribute was meant mainly for better accessibility and will be used in HTML output tool once it makes it to the master branch.

Yes you are right, list is better. Thanks.

PR #336, opened 2017-09-22
Propagate build errors to results_store

nforro:
This kind of information is missing in `results_store`.

@skisela: Maybe it could be also displayed in reports, what do you think?

### DIFF
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -537,8 +537,9 @@ class Application(object):
                     results_store.set_build_data(version, build_dict)
                     break
 
-                except SourcePackageBuildError:
+                except SourcePackageBuildError as e:
                     build_dict.update(builder.get_logs())
+                    build_dict['source_package_build_error'] = six.text_type(e)
                     results_store.set_build_data(version, build_dict)
                     #  always fail for original version
                     if version == 'old':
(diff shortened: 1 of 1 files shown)


skisela:
Well the exception message (https://github.com/rebase-helper/rebase-helper/pull/336/files#diff-55e1f3684ba905b1c1f92175ff8187d2R560) which is just slightly different from what you've added to the results_store is already set to be saved to results_store by print_summary() (https://github.com/rebase-helper/rebase-helper/pull/336/files#diff-55e1f3684ba905b1c1f92175ff8187d2R697)

But we can surely add this additional info to the report.

skisela:
Your approach is better without a doubt. The previous one was more messy as the error message could be saved/read more conveniently, than it is in print_summary().

nforro:
The point was mainly to add an information about the stage of failure - was it during srpm or rpm build?

skisela:
Well as far as I understand it, that information is already in the message string passed to the exception. Am I missing something?

nforro:
Right, I didn't realize that. But I still find checking for known key (`source_package_build_error`, `binary_package_build_error`) in `results_store` to be better than parsing exception message.

skisela:
Yes that is what I meant. You can merge it(if you want) and I can modify it to make it match the discussion here.

Issue #348, opened 2017-10-10
RFE: Add npmjs versioneer


nforro:
Add `npmjs` versioneer for nodejs packages.
Package data in JSON can be obtained from `http://registry.npmjs.org/PACKAGE`.

junaruga:
It's awesome to implement versioneer for each popular script languages!

PR #351, opened 2017-10-15
Add NPMJS Versioneer

FrNecas:
Added NPMJS versionner and test for it.

Resolves: #348

### DIFF
--- a/rebasehelper/versioneers/npmjs_versioneer.py
+++ b/rebasehelper/versioneers/npmjs_versioneer.py
@@ -1,0 +1,68 @@
+# -*- coding: utf-8 -*-
+#
+# This tool helps you to rebase package to the latest version
+# Copyright (C) 2013-2014 Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# he Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Authors: Petr Hracek <phracek@redhat.com>
+#          Tomas Hozza <thozza@redhat.com>
+
+import requests
+
+from rebasehelper.versioneer import BaseVersioneer
+from rebasehelper.logger import logger
+
+
+class NPMJSVersioneer(BaseVersioneer):
+
+    NAME = 'npmjs'
+    CATEGORIES = ['nodejs']
+
+    BASE_URL = 'https://www.npmjs.com'
+    API_URL = 'http://registry.npmjs.org'
+
+    @classmethod
+    def get_name(cls):
+        return cls.NAME
+
+    @classmethod
+    def get_categories(cls):
+        return cls.CATEGORIES
+
+    @classmethod
+    def _get_version(cls, package_name):
+        # gets the package name format needed in npm registry
+        if package_name.startswith('nodejs-'):
+            package_name = package_name.replace('nodejs-', '')
+
+        r = requests.get('{}/{}'.format(cls.API_URL, package_name))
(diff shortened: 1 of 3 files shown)


rebasehelper/versioneers/npmjs_versioneer.py
```diff
@@ -0,0 +1,67 @@
...
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Authors: Petr Hracek <phracek@redhat.com>
+#          Tomas Hozza <thozza@redhat.com>
+
+import requests
+
+from pkg_resources import parse_version
```
nforro: This import is not needed.

rebasehelper/versioneers/npmjs_versioneer.py
```diff
@@ -0,0 +1,67 @@
...
+        return cls.NAME
+
+    @classmethod
+    def get_categories(cls):
+        return cls.CATEGORIES
+
+    @classmethod
+    def _get_version(cls, package_name):
+        # gets the package name format needed in npm registry
+        package_name = package_name.split('-', 1)[1]
```
nforro: This won't work in all cases. Take for example [uglify-js](https://src.fedoraproject.org/rpms/uglify-js) package. There is no `nodejs-` prefix and there is a hyphen in package name.

rebasehelper/versioneers/npmjs_versioneer.py
```diff
@@ -0,0 +1,67 @@
...
+
+        r = requests.get('{}/{}'.format(cls.API_URL, package_name))
+        if not r.ok:
+            return None
+        data = r.json()
+
+        return data.get('dist-tags')['latest']
```
nforro: This could end with an exception. If there is no `dist-tags` key in `data`, `get()` method returns `None`, which is obviously not a dict. And if there is `dist-tags` key in data, but there is no `latest` key in resulting dict, then an exception is raised as well.

rebasehelper/versioneers/npmjs_versioneer.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @classmethod
+    def get_categories(cls):
+        return cls.CATEGORIES
+
+    @classmethod
+    def _get_version(cls, package_name):
+        # gets the package name format needed in npm registry
+        if 'nodejs-' in package_name:
+            package_name = package_name.split('-', 1)[1]
```
nforro: I think `package_name = package_name.replace('nodejs-', '')` would be sufficient here and IMHO more readable.

rebasehelper/versioneers/npmjs_versioneer.py
```diff
@@ -0,0 +1,68 @@
...
+        return cls.NAME
+
+    @classmethod
+    def get_categories(cls):
+        return cls.CATEGORIES
+
+    @classmethod
+    def _get_version(cls, package_name):
+        # gets the package name format needed in npm registry
+        if 'nodejs-' in package_name:
```
nforro: I'd rather use `if package_name.startswith('nodejs-'):`, just in case some package actually has `nodejs-` somewhere in the middle.

rebasehelper/versioneers/npmjs_versioneer.py
```diff
@@ -0,0 +1,68 @@
...
+            package_name = package_name.split('-', 1)[1]
+
+        r = requests.get('{}/{}'.format(cls.API_URL, package_name))
+        if not r.ok:
+            return None
+        data = r.json()
+        try:
+            return data.get('dist-tags')['latest']
+        except TypeError:
```
nforro: This will catch `TypeError: 'NoneType' object is not subscriptable`, but what about the case where there is no `latest` key in `dist-tags` dict?

rebasehelper/versioneers/npmjs_versioneer.py
```diff
@@ -0,0 +1,68 @@
...
+        if package_name.startswith('nodejs-'):
+            package_name = package_name.replace('nodejs-', '')
+
+        r = requests.get('{}/{}'.format(cls.API_URL, package_name))
+        if not r.ok:
+            return None
+        data = r.json()
+        try:
+            return data.get('dist-tags')['latest']
+        except (TypeError, KeyError):
```
nforro: This will work fine, however it's a bit inconsistent. You are preventing `KeyError` on `data` dict by using `get()` method, so you should either do the same with `latest` key or drop the `get()` and catch the exception.

Issue #330, opened 2017-09-15
Speed up for Travis test


junaruga:
I want to consider doing speed up Travis test to promote `rebase-helper` development.
Right now Travis tests takes 7.5 minutes for non cron mode, 35 minutes for cron mode.

At first, I check current situation.

Could you tell me which marker's tests are run for each case?
Is my recognition for "Process" correct?
Time 1, Time 2, Time 3 for case B, C is a value of "Ran for" in Travis build page.

| Case | env PYTEST_ADDOPTS | tox.ini/pytest addopts | Process | Time 1 | Time 2 | Time 3 | Average |
| ---- | ------------------ | ---------------------- | ------- | ------ | ------ | ------ | ------  |
| A. Local: “$ time tox” | "" | “-m 'not functional'” | tox (no marker) | 00:02:59 | 00:02:53 | 00:02:48 | 00:02:53 |
| B. Travis: not cron | "-m 'not optional'" | “-m 'not functional'” | build docker env + tox(no marker) | 00:07:50 | 00:07:29 | 00:07:26 | 00:07:35 |
| C. Travis: cron | "-m 1" | “-m 'not functional'” | build docker env + tox(no marker + functional + optional) | 00:35:12 | 00:36:39 | 00:34:50 | 00:35:34 |

Used pytest marker is 3 types.

- functional rebasehelper/tests/functional/test_rebase.py
- optional rebasehelper/tests/functional/test_rebase.py
- other conditional markers
  ```
  $ grep -r @pytest.mark rebasehelper/tests/
  ```
- normal (no marker) other test files

nforro:
```
$ py.test -m 'not functional' --setup-plan rebasehelper | grep "^        rebasehelper/tests"

test_application.py::TestApplication::()::test_application_sources (fixtures used: tmpdir_factory, workdir)
test_archive.py::TestArchive::()::test_archive[tar.gz] (fixtures used: archive, extracted_archive, tmpdir_factory, workdir)
test_archive.py::TestArchive::()::test_archive[tgz] (fixtures used: archive, extracted_archive, tmpdir_factory, workdir)
test_archive.py::TestArchive::()::test_archive[tar.xz] (fixtures used: archive, extracted_archive, tmpdir_factory, workdir)
test_archive.py::TestArchive::()::test_archive[tar.bz2] (fixtures used: archive, extracted_archive, tmpdir_factory, workdir)
[...]

PR #355, opened 2017-10-16
Speed up tests.

junaruga:
Related to https://github.com/rebase-helper/rebase-helper/issues/330 .

* Set ignored error types to pylintrc directly.
  Previously maybe the error code's filter was run at 2 parts that are pylintrc disable and --pylint-error-types=EF. I unified it to 1 part.

* Split pytest and pylint to set pytest's used tests directory.
  Previously `py.test rebasehelper`. But pytest searches all the files to search unit test files.
  `py.test rebasehelper/tests` is effective.

The total running time was improved from 4:19 to 2.44 (1.57 times faster) on my local.

I majored the total running time by below commands.

## Before modification

```
$ time tox
...
real  4m19.158s
user  1m53.625s
sys 0m6.112s
```

## After modification

```
$ time tox
...
real  2m44.231s
user  1m21.446s
sys 0m5.626s
```


junaruga:
Travis CI total running time (Ran for)

## Current master branch (non cron mode)
https://travis-ci.org/rebase-helper/rebase-helper/builds/288220530
09:09 sec

## After this modification.

1st time on master branch (before npmjs-versioneer commits)
https://travis-ci.org/rebase-helper/rebase-helper/builds/288379707
6:47 sec

2nd time on latest master branch.
https://travis-ci.org/rebase-helper/rebase-helper/builds/288381732
8:09 sec

pylintrc
```diff
@@ -59,7 +59,9 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=E1608,W1627,E1601,E1603,E1602,E1605,E1604,E1607,E1606,W1621,W1620,W1623,W1622,W1625,W1624,W1609,W1608,W1607,W1606,W1605,W1604,W1603,W1602,W1601,W1639,W1640,I0021,W1638,I0020,W1618,W1619,W1630,W1626,W1637,W1634,W1635,W1610,W1611,W1612,W1613,W1614,W1615,W1616,W1617,W1632,W1633,W0704,W1628,W1629,W1636
+disable=
+    C,R,W,
+    E1608,W1627,E1601,E1603,E1602,E1605,E1604,E1607,E1606,W1621,W1620,W1623,W1622,W1625,W1624,W1609,W1608,W1607,W1606,W1605,W1604,W1603,W1602,W1601,W1639,W1640,I0021,W1638,I0020,W1618,W1619,W1630,W1626,W1637,W1634,W1635,W1610,W1611,W1612,W1613,W1614,W1615,W1616,W1617,W1632,W1633,W0704,W1628,W1629,W1636
```
nforro: There is no need to disable all those warnings explicitly now, right?

junaruga: Yes, you are right. Do you like removing the warnings in line 64?
If you do, I will remove those.

nforro: Yes, remove them please. Also, it would be nice to replace error codes with symbolic names, e.g. use `old-octal-literal` instead of `E1608`. See for reference https://github.com/b0wl/readable-pylint-messages/blob/master/README.md.
Also we should consider removing some (or maybe all) of those disabled errors. But that's probably out of scope of this PR.

junaruga: > Yes, remove them please. Also, it would be nice to replace error codes with symbolic names, e.g. use old-octal-literal instead of E1608. See for reference https://github.com/b0wl/readable-pylint-messages/blob/master/README.md.

OK just moment. I will fix it.

> Also we should consider removing some (or maybe all) of those disabled errors. But that's probably out of scope of this PR.

Yes, I agree. Another PR is better.

junaruga:
Updated.

nforro:
Can you please rebase this to master? I'll merge it afterwards.

junaruga:
Rebased! Thanks.

Issue #356, opened 2017-10-16
functional_long_running test that needs a root permission to run


junaruga:
The test case: `rebasehelper/tests/functional/test_rebase.py::TestRebase::test_rebase[libtiff-4.0.7-5=>4.0.8-rpmbuild]` is likely to need a root permission to run it.

When I ran it on my local desktop environment, the dialog to require super user was appeared.

```
Authentication Required
Authentication is needed to run `/usr/bin/dnf` as the super user
```

If possible, I want to run it on normal user account.
But if it is not possible, I want below kind of logic.

https://github.com/junaruga/rpm-py-installer/blob/master/tests/test_integration.py#L15-L16

> ```
> @pytest.mark.skipif(not pytest.helpers.is_root_user(),
>                                 reason='needs root authority.')
> ```

`rebasehelper/tests/functional/test_rebase.py::TestRebase::test_rebase[libtiff-4.0.7-5=>4.0.8-mock]` is fine.
`rebasehelper/tests/functional/test_rebase.py::TestRebase::test_rebase[libtiff-4.0.7-5=>4.0.8-rpmbuild]` has the issue.

```
(venv) $ py.test --verbose -m functional_long_running rebasehelper/tests/functional/test_rebase.py::TestRebase::test_rebase
==================================== test session starts =====================================
platform linux -- Python 3.6.2, pytest-3.2.3, py-1.4.34, pluggy-0.4.0 -- /home/jaruga/git/rebase-helper/venv/bin/python3
cachedir: .cache
rootdir: /home/jaruga/git/rebase-helper, inifile: tox.ini
plugins: pylint-0.7.1, cov-2.5.1
collected 4 items                                                                             

rebasehelper/tests/functional/test_rebase.py::TestRebase::test_rebase[libtiff-4.0.7-5=>4.0.8-rpmbuild] FAILED
rebasehelper/tests/functional/test_rebase.py::TestRebase::test_rebase[libtiff-4.0.7-5=>4.0.8-mock] PASSED

========================================== FAILURES ==========================================
__________________ TestRebase.test_rebase[libtiff-4.0.7-5=>4.0.8-rpmbuild] ___________________

self = <test_rebase.TestRebase object at 0x7f11d8591ac8>, buildtool = 'rpmbuild'
package = 'libtiff', version = '4.0.8'
[...]

PR #357, opened 2017-10-16
Add warning log for the command "pkgexec dnf builddep <spec_path>"

junaruga:
Related to https://github.com/rebase-helper/rebase-helper/issues/356 .

When the command "pkexec dnf builddep <spec_path>" is run internally in `rebase-helper`, below dialog is appeared. But the text is not user friendly.

```
Authentication Required
Authentication is needed to run `/usr/bin/dnf` as the super user
```

So, I think we need to output log message too.
The message is like this.

```
2017-10-16 06:33:46,308 WARNING utils.py:662 install_build_dependencies: Authentication required to install build depenencies by command 'pkexec dnf builddep /tmp/pytest-of-jaruga/pytest-27/workdir0/libtiff.spec'
```

### DIFF
--- a/rebasehelper/utils.py
+++ b/rebasehelper/utils.py
@@ -657,6 +657,7 @@ class RpmHelper(object):
         """
         cmd = ['dnf', 'builddep', spec_path]
         if os.geteuid() != 0:
+            logger.warning("Authentication required to install build dependencies using '%s'", ' '.join(cmd))
             cmd = ['pkexec'] + cmd
         if assume_yes:
             cmd.append('-y')


junaruga:
In my experience, I could not judge if I execute it with root permission, when I saw the dialog text.

rebasehelper/utils.py
```diff
@@ -658,6 +658,8 @@ def install_build_dependencies(spec_path=None, assume_yes=False):
         cmd = ['dnf', 'builddep', spec_path]
         if os.geteuid() != 0:
             cmd = ['pkexec'] + cmd
+            logger.warning("Authentication required to install build "
+                           "depenencies by command '%s'", ' '.join(cmd))
```
nforro: I think it would make more sense to move this up by one line, so it prints the command without `pkexec` prepended.

nforro: Also, there is a typo: `depenencies`. And this sound a little bit better to me: `Authentication required to install build dependencies using '%s'`.

junaruga: OK I will fix those. Thanks for the checking.
But actually what is the max line length for this project's rule?


There is a setting in pylintrc
```
# Maximum number of characters on a single line.
max-line-length=120
```

it is checked by c0301. But as we know, all the pylint convention check is disabled by `py.test --pylint-error-types=EF` right  now.
http://pylint-messages.wikidot.com/messages:c0301

nforro: The limit is 120 characters. Yes, it's not enforced, but you shouldn't exceed it anyway.

junaruga:
Updated.

Issue #349, opened 2017-10-10
RFE: Add CPAN versioneer


nforro:
Add `CPAN` versioneer for perl packages.
API documentation: https://github.com/metacpan/metacpan-api/blob/master/docs/API-docs.md

PR #358, opened 2017-10-16
Add CPAN Versioneer

FrNecas:
Resolves: #349

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -105,7 +105,8 @@ setup(
             'anitya = rebasehelper.versioneers.anitya_versioneer:AnityaVersioneer',
             'pypi = rebasehelper.versioneers.pypi_versioneer:PyPIVersioneer',
             'rubygems = rebasehelper.versioneers.rubygems_versioneer:RubyGemsVersioneer',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -105,7 +105,8 @@ def get_requirements():
             'anitya = rebasehelper.versioneers.anitya_versioneer:AnityaVersioneer',
             'pypi = rebasehelper.versioneers.pypi_versioneer:PyPIVersioneer',
             'rubygems = rebasehelper.versioneers.rubygems_versioneer:RubyGemsVersioneer',
-            'npmjs = rebasehelper.versioneers.npmjs_versioneer:NPMJSVersioneer'
+            'npmjs = rebasehelper.versioneers.npmjs_versioneer:NPMJSVersioneer',
+            'cpan = rebasehelper.versioneers.cpan_versioneer:CPANVersioneer'
```
nforro: Keep the `,` at the end of the last item of a list, it is allowed in python and it is actually [quite convenient](https://stackoverflow.com/questions/11597901/why-does-python-allow-a-trailing-comma-in-list).

Issue #353, opened 2017-10-15
Actual comment in change log


junaruga:
For example, for `ruby.spec`.

```
(venv) $ rebase-helper
...
(venv) $ diff ruby.spec rebase-helper-results/rebased-sources/ruby.spec 
24c24
< %global release 84
---
> %global release 1
1032a1033,1035
> * Sun Oct 15 2017 rebase-helper <rebase-helper@localhost.local> - 2.4.2-1
> - New upstream release 2.4.2
```
I want to change the comment "New upstream release 2.4.2" to more actually used comment.
In a case of RubyGem package (`rubygem-foo`), the comment pattern "Update to Foo Bar X.Y.Z." is often used.

The logic is like this.
`rubygem-foo-bar` => `foo-bar` => `Foo Bar` =>  "Update to Foo Bar X.Y.Z."

How do you think?

PR #359, opened 2017-10-17
Add changelog template

FrNecas:
Resolves: #353 

Custom changelog entry can be set using --changelog-entry "message". RPM macros can be used to add version and other things (%{version}, %{name} etc.)
```

[root@9af0720655ba rubygem-afm]# python3 ../rebase-helper/rebase-helper.py --changelog-entry "Update to {VERSION}"
...
 
[root@9af0720655ba rubygem-afm]# diff rubygem-afm.spec rebase-helper-results/rebased-sources/rubygem-afm.spec
6c6
< Release: 6%{?dist}
---
> Release: 1%{?dist}
74a75,77
> * Tue Oct 17 2017 rebase-helper <rebase-helper@localhost.local> - 0.2.2-1
> Update to 0.2.2
```

@junaruga Which other options of formating would you like to have?

### DIFF
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -24,6 +24,7 @@ from __future__ import print_function
 import os
 import shutil
 import logging
+import rpm
 
 import git
 import six
@@ -88,6 +89,8 @@ class Application(object):
         self.kwargs['rebased_sources_dir'] = self.rebased_sources_dir
 
         self.kwargs['non_interactive'] = self.conf.non_interactive
+
+        self.kwargs['changelog_entry'] = self.conf.changelog_entry
 
         logger.debug("Rebase-helper version: %s" % version.VERSION)
 
@@ -172,6 +175,7 @@ class Application(object):
         self.rebase_spec_file_path = get_rebase_name(self.rebased_sources_dir, self.spec_file_path)
 
         self.spec_file = SpecFile(self.spec_file_path,
+                                  self.conf.changelog_entry,
                                   self.execution_dir,
                                   download=not self.conf.not_download_sources)
         # Check whether test suite is enabled at build time
@@ -445,8 +449,8 @@ class Application(object):
 
         # Generate patch
         self.rebased_repo.git.add(all=True)
-        self.rebased_repo.index.commit('New upstream release {}'.format(self.rebase_spec_file.get_full_version()),
-                                       skip_hooks=True)
+        self.rebase_spec_file._update_data()
+        self.rebased_repo.index.commit(rpm.expandMacro(self.conf.changelog_entry))
         patch = self.rebased_repo.git.format_patch('-1', stdout=True, stdout_as_string=False)
         with open(os.path.join(self.results_dir, 'changes.patch'), 'wb') as f:
             f.write(patch)
--- a/rebasehelper/cli.py
+++ b/rebasehelper/cli.py
@@ -230,6 +230,11 @@ class CLI(object):
             default=None,
             help="new upstream sources"
         )
+        parser.add_argument(
+            "--changelog-entry",
+            default="- New upstream release %{version}",
+            help="text to use as changelog entry, can contain RPM macros, which will be expanded"
+        )
         return parser
 
     def __init__(self, args=None):
--- a/rebasehelper/specfile.py
+++ b/rebasehelper/specfile.py
@@ -125,10 +125,11 @@ class SpecFile(object):
                         '%files',
                         '%changelog']
 
-    def __init__(self, path, sources_location='', download=True):
+    def __init__(self, path, changelog_entry, sources_location='', download=True):
         self.path = path
         self.download = download
         self.sources_location = sources_location
+        self.changelog_entry = changelog_entry
         #  Read the content of the whole SPEC file
         rpm.addMacro("_sourcedir", self.sources_location)
         self._read_spec_content()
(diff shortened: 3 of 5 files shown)


junaruga:
Thanks for the PR!

I think @FrNecas there are 2 types of the approach to implement features in `rebase-helper` in general.

1. `rebase-helper` delegates users (argument) options to run `rebase-helper`. That way means adding argument option to execute `rebase-helper` as they like. Your approach is this way. This way's demerit is many argument options. It's hard for users to learn how to run `rebase-helper`.

2. `rebase-helper` predicts the best way from input data such as RPM spec file. each programming language's versioneer and running without 1st argument are this approach. It's a kind of AI or automation's approach.

I prefer the approach No 2. Because I am trying to automate RPM development as much as possible, through `rebase-helper`. And the No 2 can solve below situation "Update to Foo Bar X.Y.Z.".

> https://github.com/rebase-helper/rebase-helper/issues/353#issue-265572269
> The logic is like this.
> rubygem-foo-bar => foo-bar => Foo Bar => "Update to Foo Bar X.Y.Z."

If we will proceed your way, at least I want `{NAME}` (= Foo Bar) too.

junaruga:
@nforro Yes, your recognition is correct. in my opinion, handling this kind of logic by SPEC Hook class for each programming language looks good to me to recognize the prefix `rubygem-`.

In my understanding, historically `rebase-helper`'s concept had been a type No 1.
But recently `rebase-helper` is starting to change the concept to a type No 2 by the requirement of the age.

rebasehelper/application.py
```diff
@@ -89,6 +90,8 @@ def __init__(self, cli_conf, execution_dir, results_dir, debug_log_file):
 
         self.kwargs['non_interactive'] = self.conf.non_interactive
 
+        self.kwargs['changelog-entry'] = self.conf.changelog_entry
```
nforro: Replace hyphen with underscore. Hyphen is not allowed in identifier names so this can't be used e.g. as function argument.

rebasehelper/application.py
```diff
@@ -172,6 +175,7 @@ def _prepare_spec_objects(self):
         self.rebase_spec_file_path = get_rebase_name(self.rebased_sources_dir, self.spec_file_path)
 
         self.spec_file = SpecFile(self.spec_file_path,
+                                  self.kwargs['changelog-entry'],
```
nforro: I'd rather see `self.conf.changelog_entry` here. You could also pass the entire `kwargs` dict so other attributes can also be consumed by `SpecFile` class (if necessary).

rebasehelper/application.py
```diff
@@ -445,8 +449,8 @@ def generate_patch(self):
 
         # Generate patch
         self.rebased_repo.git.add(all=True)
-        self.rebased_repo.index.commit('New upstream release {}'.format(self.rebase_spec_file.get_full_version()),
-                                       skip_hooks=True)
+        self.rebase_spec_file._update_data()
+        self.rebased_repo.index.commit(rpm.expandMacro(self.kwargs['changelog-entry']))
```
nforro: Same as above, I think it's much nicer to use `self.conf.changelog_entry`.

nforro:
@junaruga: RPM macros can now be used in the template, so you could use something like this to get what you want:
```
rebase-helper --changelog-entry 'Update to '%{lua: print((rpm.expand("%name"):gsub("-%l", string.upper):gsub("^rubygem%-", ""):gsub("-", " ")))} %{version}.'
```
We will try to implement the second approach in a different PR.

junaruga:
> @junaruga: RPM macros can now be used in the template, so you could use something like this to get what you want:

OK, thanks!

> We will try to implement the second approach in a different PR.

Thanks! That's helpful for me.

nforro:
@FrNecas: It seems that your commits got mixed up. Also a few notes on your commit messages:

> **Add option to change changelog output**

I think `Add option to customize changelog entry` would sound better.

> User can change the text of changelog using
> parameter --changelog-entry. RPM macros can
> be used to display name or version.

Why so short lines? Hard limit is 80 characters, it's best to try to fit into 72 characters, but ~40 characters is just needlessly little and it actually makes the text harder to read.
Also, you are using the word `option` in the subject but `parameter` in the body.
And it's not exactly accurate to say that RPM macros can be used to display name or version, because they can be used to show basically anything.

This is how I think the body should look like:
```
This enables users to customize changelog entry using
"--changelog-entry" option.
RPM macros (e.g. "%{name}" or "%{version}") will be expanded.
```

>Default is set to:
>"- New upstream release %{version}"

I don't think this is worth mentioning as the default is the same as before.

> **Fix tests after changelog changes**

I think simple `Update tests to handle "--changelog-entry" option` would be sufficient.

> One more argument was added
> so some tests failed

Sorry, but this just doesn't make any sense. Anyway, I don't think this commit deserves any more explanation. Just the subject alone is enough.

rebasehelper/cli.py
```diff
@@ -230,6 +230,12 @@ def build_parser():
             default=None,
             help="new upstream sources"
         )
+        parser.add_argument(
+            "--changelog-entry",
+            default="- New upstream release %{version}",
+            help="format shown in changelog after updating to a higher version, "
+                 "you can use RPM macros (%{version} etc.)"
```
nforro: I think something like this sounds better (and is more consistent with help text of other options): `text to use as changelog entry, can contain RPM macros, which will be expanded`.

Issue #352, opened 2017-10-15
Adding git name and email in change log


junaruga:
For example, for `ruby.spec`.

```
(venv) $ rebase-helper
...
(venv) $ diff ruby.spec rebase-helper-results/rebased-sources/ruby.spec 
24c24
< %global release 84
---
> %global release 1
1032a1033,1035
> * Sun Oct 15 2017 rebase-helper <rebase-helper@localhost.local> - 2.4.2-1
> - New upstream release 2.4.2
```

In this update, I want to include my user name, and email address.
Maybe the result of below commands.

```
$ git config user.name
Jun Aruga
$ git config user.email
jaruga@redhat.com
```

nforro:
> Maybe the result of below commands.
> `$ git config user.name`
> `$ git config user.email`

That's exactly what is being done, see: https://github.com/rebase-helper/rebase-helper/blob/master/rebasehelper/specfile.py#L1298-L1299

Could you provide more information so we can investigate why it doesn't work for you?

junaruga:
OK, let me check it.

junaruga:
@nforro I found the reason.

https://github.com/rebase-helper/rebase-helper/blob/master/rebasehelper/utils.py#L817-L829

GitHelper#get_user, get_email methods use default user name and email when the command line `git config --get --global user.name` is failed.

But on virtual environment, the result is like this. `--global` does not work on virtual environment.
I want to remove `--global` from the command line if it is no problem.

```
(venv) $ git config --get --global user.name
(venv) $ echo $?
1

(venv) $ git config user.name
Jun Aruga
(venv) $ echo $?
0

(venv) $ git config --get user.name
Jun Aruga
(venv) $ echo $?
0
```

junaruga:
Ah it's not virtualenv specified issue. It is reproduced on my normal local environment.

```
$ git config --get --global user.name
$ echo $?
1

$ git config user.name
Jun Aruga
$ echo $?
0

$ git config --get user.name
Jun Aruga
$ echo $?
0
```

nforro:
I'm not sure why we used `--global` in the first place, but I guess it's not really necessary.

PR #360, opened 2017-10-18
Fix to get git config user.name and user.email correctly.

junaruga:
This fixes https://github.com/rebase-helper/rebase-helper/issues/352

I added warnings for the case of failure to get user and email.
I added very simple unit tests.

## The case for the success to get user and email

```
(venv) $ rebase-helper

(venv) $ diff rubygem-byebug.spec rebase-helper-results/rebased-sources/rubygem-byebug.spec
126a127,129
> * Wed Oct 18 2017 Jun Aruga <jaruga@redhat.com> - 9.1.0-1
> - New upstream release 9.1.0
>
```

## The case for the failure to get user and email

```
(venv) $ rebase-helper
...
Faild to get Git user.name. Return the fallback value 'rebase-helper'
Faild to get Git user.email. Return the fallback value 'rebase-helper@localhost.local'
...
Faild to get Git user.name. Return the fallback value 'rebase-helper'
Faild to get Git user.email. Return the fallback value 'rebase-helper@localhost.local'
...

(venv) $ diff rubygem-byebug.spec rebase-helper-results/rebased-sources/rubygem-byebug.spec
126a127,129
> * Wed Oct 18 2017 rebase-helper <rebase-helper@localhost.local> - 9.1.0-1
> - New upstream release 9.1.0
> 
```

### DIFF
--- a/rebasehelper/tests/test_utils.py
+++ b/rebasehelper/tests/test_utils.py
@@ -26,11 +26,13 @@ import random
 import string
 import sys
 
+import git
 import rpm
 import pytest
 
 from six import StringIO
 
+from rebasehelper.utils import GitHelper
 from rebasehelper.utils import ConsoleHelper
 from rebasehelper.utils import DownloadHelper
 from rebasehelper.utils import DownloadError
@@ -40,6 +42,73 @@ from rebasehelper.utils import TemporaryEnvironment
 from rebasehelper.utils import RpmHelper
 from rebasehelper.utils import MacroHelper
 from rebasehelper.utils import LookasideCacheHelper
+
+
+class TestGitHelper(object):
+
+    def write_config_file(self, config_file, name, email):
+        with open(config_file, 'w') as f:
+            f.write('[user]\n'
+                    '    name = {0}\n'
+                    '    email = {1}\n'.format(name, email))
+
+    @pytest.mark.parametrize('config', [
+        # Get from $XDG_CONFIG_HOME/git/config
+        'global',
+        # Get from included file in $XDG_CONFIG_HOME/git/config
+        'global_include',
+        # Get from $repo_path/.git/config
+        'local',
+        # Get from GIT_CONFIG
+        'env',
+    ])
+    def test_get_user_and_email(self, config, workdir):
+        name = 'Foo Bar'
+        email = 'foo@bar.com'
+        env_git_config = os.environ.get('GIT_CONFIG')
+        env_xdg_config_home = os.environ.get('XDG_CONFIG_HOME')
+
+        try:
+            if config == 'global':
+                work_git_path = os.path.join(workdir, 'git')
+                os.makedirs(work_git_path)
+
+                config_file = os.path.join(work_git_path, 'config')
+                self.write_config_file(config_file, name, email)
+                os.environ['XDG_CONFIG_HOME'] = workdir
+            elif config == 'global_include':
+                work_git_path = os.path.join(workdir, 'git')
+                os.makedirs(work_git_path)
+
+                config_file = os.path.join(work_git_path, 'config')
+                with open(config_file, 'w') as f:
+                    f.write('[include]\n'
+                            '    path = included_config\n')
+                included_config_file = os.path.join(work_git_path, 'included_config')
+                self.write_config_file(included_config_file, name, email)
+                os.environ['XDG_CONFIG_HOME'] = workdir
+            elif config == 'local':
+                repo = git.Repo.init(workdir)
+                repo.git.config('user.name', name, local=True)
+                repo.git.config('user.email', email, local=True)
+            elif config == 'env':
+                config_file = os.path.join(workdir, 'git_config')
+                os.environ['GIT_CONFIG'] = config_file
+                self.write_config_file(config_file, name, email)
+            else:
+               raise RuntimeError()
+
+            assert name == GitHelper.get_user()
+            assert email == GitHelper.get_email()
+        finally:
+            if env_git_config:
+                os.environ['GIT_CONFIG'] = env_git_config
+            elif 'GIT_CONFIG' in os.environ:
+                del os.environ['GIT_CONFIG']
+            if env_xdg_config_home:
+                os.environ['XDG_CONFIG_HOME'] = env_xdg_config_home
+            elif 'XDG_CONFIG_HOME' in os.environ:
+                del os.environ['XDG_CONFIG_HOME']
 
 
 class TestConsoleHelper(object):
--- a/rebasehelper/utils.py
+++ b/rebasehelper/utils.py
@@ -817,15 +817,17 @@ class GitHelper(object):
     @classmethod
     def get_user(cls):
         try:
-            return git.cmd.Git().config('--global', 'user.name', get=True, stdout_as_string=six.PY3)
+            return git.cmd.Git().config('user.name', get=True, stdout_as_string=six.PY3)
         except git.GitCommandError:
+            logger.warning("Failed to get configured git user name, using '%s'", cls.GIT_USER_NAME)
             return cls.GIT_USER_NAME
 
     @classmethod
     def get_email(cls):
         try:
-            return git.cmd.Git().config('--global', 'user.email', get=True, stdout_as_string=six.PY3)
+            return git.cmd.Git().config('user.email', get=True, stdout_as_string=six.PY3)
         except git.GitCommandError:
+            logger.warning("Failed to get configured git user email, using '%s'", cls.GIT_USER_EMAIL)
             return cls.GIT_USER_EMAIL
 
     @classmethod


rebasehelper/utils.py
```diff
@@ -817,15 +817,17 @@ class GitHelper(object):
     @classmethod
     def get_user(cls):
         try:
-            return git.cmd.Git().config('--global', 'user.name', get=True, stdout_as_string=six.PY3)
+            return git.cmd.Git().config('user.name', get=True, stdout_as_string=six.PY3)
         except git.GitCommandError:
+            logger.warning("Faild to get Git user.name. Return the fallback value '%s'", cls.GIT_USER_NAME)
```
nforro: Ok, doesn't this sound better? `"Failed to get configured git user name, using '%s'"`

junaruga: Your way looks better and more simple. I will fix it.

rebasehelper/tests/test_utils.py
```diff
@@ -42,6 +43,12 @@
 from rebasehelper.utils import LookasideCacheHelper
 
 
+class TestGitHelper(object):
+    def test_get_user_and_email(self):
+        assert GitHelper.get_user()
+        assert GitHelper.get_email()
```
nforro: What's the point of this test? In what situation should it fail? I don't think I get it.

junaruga: Sorry it may be meaningless.
Actually I want to test above cases that I tested.
But I was not sure how much we implement the unit tests in rebase-helper.

Do you think we need additional unit tests for that in this case?

nforro: If you want to test whether `git config --get user.name` works as expected, you should first configure it (e.g. like [here](https://github.com/rebase-helper/rebase-helper/blob/master/rebasehelper/application.py#L467)) and then check if `GitHelper.get_user()` returns what it should. You could configure it system-wide and then on user level and test both cases.
Where do you personally have it configured btw? As it didn't work with `--global` argument. You should test for such case too.

junaruga: @nforro thanks for the information. I will do it.

My configuration is

`~/.gitconfig` is this file.
> https://github.com/junaruga/my-config-pub/blob/master/home/.gitconfig
> [include]
> path = .gitconfig.local

And

```
$ cat ~/.gitconfig.local
[user]
    email = jaruga@redhat.com
    name = Jun Aruga
```

nforro: Interesting, so if `include.path` is used, `--global` option doesn't work. I wonder whether that's a bug.

nforro: Aha, it works with `--include` agument. So `git config --global --include --get user.name`. But I'm not sure whether it's better to drop `--global` or keep it and add `--include`. I can't think of any downsides of not using `--global` at the moment.

junaruga: I was also success.

```
$ git config --global --includes --get user.name
Jun Aruga

$ echo $?
0
```

I found below explanation.

```
$ man git-config
...
       --[no-]includes   
           Respect include.*  directives in config files when looking up values. Defaults
           to off when a specific file is given (e.g., using --file, --global, etc) and on
           when searching all config files.
```

I prefer `git config --get user.name` than `git config --global --includes --get user.name`.

Because `git config --get user.name` can decide to get the value by following git's default priority. And it covers all the below cases. In my experience, I used to manage Git name and email by local scope (`$(repo_path)/.git/config`).

* `--global` (`~/.gitconfig`)
* `--system` (`$(prefix)/etc/gitconfig`)
* `--local` (`$(repo_path)/.git/config`)

junaruga: I have got a long vacation until today. I would start working again from today. :)

junaruga:
@nforro I updated with the updated unit test code. I fixed the warning messages that you mentioned.
I prefer just removing `--global` ratter than `--global --include` , because it covers all the below case.

- My case: user name and email are in included file by `~/.gitconfig` (`--global`).
- User name and email are in `$(prefix)/etc/gitconfig` (`--system`)
- User name and email are in `$(repo_path)/.git/config` (`--local`).

I implemented the unit test with testing git repository.
But I am not confident the way is this project's way.

How do you think?

rebasehelper/tests/test_utils.py
```diff
@@ -42,6 +44,32 @@
...
+        current_dir = os.getcwd()
+        try:
+            with TemporaryEnvironment() as temp:
+                temp_repo_path = temp.path()
+                os.chdir(temp_repo_path)
+                yield
+        finally:
+            os.chdir(current_dir)
+
+    @pytest.mark.usefixtures('temp_dir')
```
nforro: You can use existing [`workdir` fixture](https://github.com/rebase-helper/rebase-helper/blob/master/rebasehelper/tests/conftest.py#L33-L40).

rebasehelper/tests/test_utils.py
```diff
@@ -42,6 +44,32 @@
...
+    def test_get_user_and_email(self):
+        temp_repo_path = os.getcwd()
+        repo = git.Repo.init(temp_repo_path)
+        name = 'Foo Bar'
+        email = 'foo@bar.com'
+        repo.git.config('user.name', name, local=True)
+        repo.git.config('user.email', email, local=True)
```
nforro: This only tests local-level configuration. It would be good to test also global-level and user-level, and also the case with include path. But I'm not sure how to do that, we certainly don't want tests to overwrite someone's configuration.

nforro: You could for example create a set of config files (include one in the other etc.) and then export `GIT_CONFIG` environment variable so that those files are read by `GitHelper.get_user()` and `GitHelper.get_email()`.

junaruga: @nforro Using `GIT_CONFIG` is same with `-f config-file` option. It is not `--global`'s flow.
In case of `--global`, using `$XDG_CONFIG_HOME/git/config` is possible.

- `--system`: I have no idea.
- `--global`: possible with `$XDG_CONFIG_HOME/git/config`
- `--local`: possible with `$repo_url/.git/config` (current logic)
- `-f config-file`: possible with `GIT_CONFIG`.

rebasehelper/tests/test_utils.py
junaruga: ok I will use it.

junaruga:
@nforro, I rebased with `global`, `local` and `GIT_CONFIG` cases. How?

rebasehelper/tests/test_utils.py
```diff
@@ -42,6 +44,61 @@
...
+
+    @pytest.mark.parametrize('config', [
+        # Get from $XDG_CONFIG_HOME/git/config
+        'global',
+        # Get from $repo_path/.git/config
+        'local',
+        # Get from GIT_CONFIG
+        'env',
+    ])
+    @pytest.mark.usefixtures('workdir')
```
nforro: You can just pass `workdir` parameter to `test_get_user_and_email` method and you don't have to use `@pytest.mark.usefixtures('workdir')` nor call `os.getcwd()`.

nforro:
The latest changes look good, but I'd still like to see the include path case tested. I think maybe you can just write configuration to files directly, without using `repo.git.config()`.

junaruga:
OK, I used `workdir` as just the parameter.
I added the case to get from included file in `$XDG_CONFIG_HOME/git/config` (`global`).

I changed to write configuration file directly, with `git.cmd.Git().config('user.name', name)`.
`repo.git.config('user.name', name, local=True)` is only used in the case of `local`.

nforro:
> I changed to write configuration file directly, with `git.cmd.Git().config('user.name', name)`.
`repo.git.config('user.name', name, local=True)` is only used in the case of `local`.

By *directly* I meant writing string constant to a file, same way as you are writing the `[include]` section.
I think it would be more readable. You could also use the same string constant for all config files, just changing the path.

junaruga:
> By directly I meant writing string constant to a file, same way as you are writing the [include] section.
I think it would be more readable. You could also use the same string constant for all config files, just changing the path.

OK, rebased. I created the same string constant as a method. How?

rebasehelper/tests/test_utils.py
```diff
@@ -42,6 +44,77 @@
...
 
+class TestGitHelper(object):
+
+    def write_config_file(self, config_file, name, email):
+        with open(config_file, 'w') as f:
+            f.write('''
+[user]
+    name = {0}
+    email = {1}
+'''.format(name, email))
```
nforro: How about this? So it doesn't break indentation.
```python
                    f.write('[user]\n'
                            '    name = {0}\n'
                            '    email = {1}\n'.format(name, email))
```

Or do it your way but define it as constants at the top of the file, where indentation is not needed.

junaruga: Sure, I am fine for your suggestion. Just moment. I will rebase again. Same for `[include]` part.

Issue #361, opened 2017-10-18
Builder-options do not store entered value


skisela:
`builder-options` do not have any value at the time of calling `*_tool._build_rpm()`.

PR #362, opened 2017-10-18
Create SRPM builder tools

skisela:
Pass builder_options values to `build_dict{}`

Implement `SRPMBuilderTools` to allow users to use mock for source rpms creation.

Implement `SRPM options` to allow users to set configuration for building source rpms.

Resolves: #366 #361

### DIFF
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -520,6 +520,9 @@ class Application(object):
                 build_dict['builds_nowait'] = self.conf.builds_nowait
                 build_dict['build_tasks'] = self.conf.build_tasks
                 build_dict['builder_options'] = self.conf.builder_options
+                build_dict['srpm_builder_options'] = self.conf.srpm_builder_options
+                build_dict['srpm_buildtool'] = self.conf.srpm_buildtool
+
                 patches = [x.get_path() for x in spec_object.get_patches()]
                 spec = spec_object.get_path()
                 sources = spec_object.get_sources()
@@ -553,7 +556,8 @@ class Application(object):
                     # Build finishes properly. Go out from while cycle
 
                     # Do not store rebase-helper internal data to the results
-                    for opt in ['builds_nowait', 'build_tasks', 'builder_options']:
+                    for opt in ['builds_nowait', 'build_tasks', 'builder_options',
+                                'srpm_builder_options', 'srpm_buildtool']:
                         build_dict.pop(opt)
                     results_store.set_build_data(version, build_dict)
                     break
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -517,6 +517,9 @@ class Application(object):
 
                 build_dict['name'] = pkg_name
                 build_dict['version'] = pkg_version
+                build_dict['builds_nowait'] = self.conf.builds_nowait
+                build_dict['build_tasks'] = self.conf.build_tasks
+                build_dict['builder_options'] = self.conf.builder_options
                 patches = [x.get_path() for x in spec_object.get_patches()]
                 spec = spec_object.get_path()
                 sources = spec_object.get_sources()
@@ -548,6 +551,10 @@ class Application(object):
                             if build_dict['rpm'] is None:
                                 return False
                     # Build finishes properly. Go out from while cycle
+
+                    # Do not store rebase-helper internal data to the results
+                    for opt in ['builds_nowait', 'build_tasks', 'builder_options']:
+                        build_dict.pop(opt)
                     results_store.set_build_data(version, build_dict)
                     break
 
--- a/rebasehelper/build_helper.py
+++ b/rebasehelper/build_helper.py
@@ -25,7 +25,6 @@ import os
 import six
 import pkg_resources
 
-from rebasehelper.utils import ProcessHelper
 from rebasehelper.utils import PathHelper
 from rebasehelper.utils import TemporaryEnvironment
 from rebasehelper.logger import logger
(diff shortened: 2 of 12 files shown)


nforro:
@skisela: FYI: it was removed in commit 368d2678d42c87a3553e3ba09b053975116f85f7, any idea why?

skisela:
@nforro yes, you are right, it was to solve #222 but it introduced even worse problems. I will fix it. 

This certainly needs at least a basic test for the content of build_dict{}.

skisela:
Thinking of testing that, testing in functional tests comes to my mind. But it would require expanding the app.run() call I guess and I do not think that it is worth it.

rebasehelper/application.py
```diff
@@ -543,6 +546,10 @@ def build_packages(self):
                             if build_dict['rpm'] is None:
                                 return False
                     # Build finishes properly. Go out from while cycle
+
+                    # Do not store rebase-helper internal data to the results
+                    for opt in ['builds_nowait', 'build_tasks', 'builder_options']:
+                        build_dict.pop(opt)
```
nforro: Just an idea, but wouldn't it be better to separate results from inputs?

rebasehelper/cli.py
```diff
@@ -127,6 +127,12 @@ def build_parser():
             help="build tool to use, defaults to %(default)s"
         )
         parser.add_argument(
+            "--srpm-buildtool",
+            choices=SRPMBuilder.get_supported_tools(),
+            default=SRPMBuilder.get_default_tool(),
+            help="build tool to use, defaults to %(default)s"
```
nforro: This should be `SRPM build tool to use` or `build tool to use for SRPM build`.

rebasehelper/srpm_build_tools/srpm_mock_tool.py
```diff
@@ -0,0 +1,80 @@
...
+        return cls.CMD
+
+    @classmethod
+    def is_default(cls):
+        return cls.DEFAULT
+
+    @classmethod
+    def do_build_srpm(cls, spec, sources, workdir, results_dir, srpm_options=None):
+        """
+        Build SRPM using rpmbuild.
```
nforro: This doesn't seem right.

rebasehelper/srpm_build_tools/srpm_mock_tool.py
```diff
@@ -0,0 +1,80 @@
...
+import os
+
+from rebasehelper.logger import logger
+from rebasehelper.build_helper import SourcePackageBuildError
+from rebasehelper.build_helper import RpmbuildTemporaryEnvironment
+from rebasehelper.utils import PathHelper
+from rebasehelper.utils import ProcessHelper
+
+
+class SRPMMockBuildTool(object):
```
nforro: `MockSRPMBuildTool` makes more sense to me.

rebasehelper/cli.py
```diff
@@ -207,6 +213,13 @@ def build_parser():
                  "and note that = before it is mandatory"
         )
         parser.add_argument(
+            "--srpm-options",
```
nforro: Definitely use `--srpm-builder-options`. The alternative is shorter but confusing.

rebasehelper/build_helper.py
```diff
@@ -244,35 +243,27 @@ def get_detached_task(cls, task_id, results_dir):
...
-                                                   output=output)
-
-        if ret != 0:
-            return None
-        else:
-            return PathHelper.find_first_file(workdir, '*.src.rpm')
+
+    @staticmethod
+    def get_builder_options(**kwargs):
+        builder_options = kwargs.get('builder_options', None)
```
nforro: Using `None` as second argument here is redundant, see https://docs.python.org/3/library/stdtypes.html#dict.get.

rebasehelper/application.py
skisela: Maybe it is just a matter of a preference. Seeing output is tightly connected with the input parameters used, which can be useful. But I can hide the keys with the shown way.

nforro: I agree with @thozza, IMO it shouldn't be visible in the output. But that doesn't mean it can't be part of `build_dict`.

thozza:
@skisela @nforro I have one comment. In the past, we used to use mock for building SRPMs. However, I'm not aware of any benefit of doing it, there are only downsides. Mock provides "clean" environment for package build. However when SRPM is created, there is NO build. You can compare the SRPM creation to creating a tarball. Since mock build for SRPM took too long without any benefit. That's why it was always built using rpmbuild directly. Passing some SRPM options should be IMHO possible even without using mock.

nforro:
@thozza: There can be a difference between SRPM created using rpmbuild and SRPM created using mock, because RPM macros can influence the process. And local RPM macros can be defined differently than those in a clean mock environment.
But I agree, this can be also resolved by passing suitable options to rpmbuild.

skisela:
@thozza In case a clean(clean of macros) environment is wanted even during the "build" of SRPM, isn't it better to at least provide the option to use mock, while having the default "SRPM build tool" set to rpmbuild?

rebasehelper/cli.py
```diff
@@ -207,6 +213,13 @@ def build_parser():
                  "and note that = before it is mandatory"
         )
         parser.add_argument(
+            "--srpm-builder-options",
+            default=None,
+            metavar="SRPM_OPTIONS",
```
nforro: This should also be `SRPM_BUILDER_OPTIONS`, and the same in the help text below.

setup.py
```diff
@@ -90,6 +91,10 @@ def get_requirements():
             'koji = rebasehelper.build_tools.koji_tool:KojiBuildTool',
             'copr = rebasehelper.build_tools.copr_tool:CoprBuildTool',
         ],
+        'rebasehelper.srpm_build_tools': [
+            'rpmbuild = rebasehelper.srpm_build_tools.srpm_rpmbuild_tool:RpmbuildSRPMBuildTool',
+            'mock = rebasehelper.srpm_build_tools.srpm_mock_tool:MockSRPMBuildTool',
```
nforro: I think the package names should be just `rpmbuild_tool` and `mock_tool`, but if you want to keep `srpm` there, it should  be `rpmbuild_srpm_tool` and `mock_srpm_tool`.

thozza:
@skisela sure, the macros use case is definitely valid, although I'm not sure how critical it is. In other words, is there anyone asking for it? In the end, it is up to you what you'll implement. I just don't want you to do the same mistakes we already did 😄

rebasehelper/build_helper.py
```diff
@@ -244,35 +243,27 @@ def get_detached_task(cls, task_id, results_dir):
...
+
+    @staticmethod
+    def get_srpm_buildtool(**kwargs):
+        return kwargs.get('srpm_buildtool')
 
     @classmethod
-    def _build_srpm(cls, spec, sources, patches, results_dir):
+    def _build_srpm(cls, spec, sources, patches, results_dir, **kwargs):
         """
         Builds the SRPM using rpmbuild
```
nforro: The docstring is no longer valid.

rebasehelper/srpm_build_tools/mock_tool.py
```diff
@@ -0,0 +1,81 @@
...
+    @classmethod
+    def get_build_tool_name(cls):
+        return cls.CMD
+
+    @classmethod
+    def is_default(cls):
+        return cls.DEFAULT
+
+    @classmethod
+    def do_build_srpm(cls, spec, sources, workdir, results_dir, srpm_builder_options):
```
nforro: The `do_` prefix is not necessary, it was used in `BuildToolBase` to differentiate the method from already existing `build_srpm`.

rebasehelper/srpm_build_tools/mock_tool.py
```diff
@@ -0,0 +1,81 @@
...
+import os
+
+from rebasehelper.logger import logger
+from rebasehelper.build_helper import SourcePackageBuildError
+from rebasehelper.build_helper import RpmbuildTemporaryEnvironment
+from rebasehelper.utils import PathHelper
+from rebasehelper.utils import ProcessHelper
+
+
+class MockSRPMBuildTool(object):
```
nforro: I think it's good practice to derive classes like this from a base class, even though we don't plan to introduce more subclasses in the future.

rebasehelper/srpm_build_tools/mock_tool.py
```diff
@@ -0,0 +1,80 @@
...
+
+import os
+
+from rebasehelper.logger import logger
+from rebasehelper.build_helper import BuildToolBase
+from rebasehelper.utils import PathHelper
+from rebasehelper.utils import ProcessHelper
+
+
+class MockSRPMBuildTool(BuildToolBase):
```
nforro: This works, but is technically incorrect. If you are deriving from `BuildToolBase`, you should implement all *virtual* methods. But those are not needed with SRPM build tool. So you should create a different base class, or use `BuildToolBase` as a base for newly created `RPMBuildToolBase` and `SRPMBuildToolBase`.

PR #367, opened 2017-10-25
Compare local and the latest upstream version

FrNecas:
If local version is higher than (or equal to) the latest upstream version, rebase-helper doesn't do anything.

For example:

 ```
[root@9af0720655ba rubygem-afm]# python3 ../rebase-helper/rebase-helper.py 
Results directory 'rebase-helper-results' exists, removing it
Workspace directory 'rebase-helper-workspace' exists, removing it
Running 'rubygems' versioneer
Determined latest upstream version '0.2.2'

Your current version is the latest available version.
```

Resolves: #354

### DIFF
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -29,6 +29,8 @@ import rpm
 import git
 import six
 
+from pkg_resources import parse_version
+
 from rebasehelper.archive import Archive
 from rebasehelper.specfile import SpecFile, get_rebase_name, spec_hooks_runner
 from rebasehelper.logger import logger, logger_report, LoggerHelper
@@ -207,6 +209,9 @@ class Application(object):
             self.rebase_spec_file.set_version(version)
             self.rebase_spec_file.set_extra_version_separator(separator)
             self.rebase_spec_file.set_extra_version(extra_version)
+
+        if parse_version(self.rebase_spec_file.get_version()) <= parse_version(self.spec_file.get_version()):
+            raise RebaseHelperError("Current version is equal to or newer than the requested version, nothing to do.")
 
         # run spec hooks
         spec_hooks_runner.run_spec_hooks(self.spec_file, self.rebase_spec_file, **self.kwargs)

### CODE REVIEW

File: rebasehelper/application.py [hunk 1]
```diff
@@ -44,6 +44,7 @@
 from rebasehelper.build_log_analyzer import BuildLogAnalyzerMakeError, BuildLogAnalyzerPatchError
 from rebasehelper.versioneer import versioneers_runner
 from rebasehelper import version
+from pkg_resources import parse_version
```
nforro: This import line should be above local imports (those from `rebasehelper` package). So it should be under `import six` separated by newline.

nforro: See https://www.python.org/dev/peps/pep-0008/#imports.

rebasehelper/application.py
```diff
@@ -194,6 +195,9 @@ def _prepare_spec_objects(self):
                 raise RebaseHelperError('Could not determine latest upstream version '
                                         'and no SOURCES argument specified!')
 
+        if parse_version(self.conf.sources) <= parse_version(self.spec_file.get_version()):
+            raise RebaseHelperError("Your local version is the latest available version.")
```
nforro: You shouldn't expect that the version is the lastest available, it could have been specified on the command line and could be some older version. So I think better message is this: `Current version is equal to or newer than the requested version, nothing to do.`.

rebasehelper/application.py
```diff
@@ -194,6 +195,9 @@ def _prepare_spec_objects(self):
                 raise RebaseHelperError('Could not determine latest upstream version '
                                         'and no SOURCES argument specified!')
 
+        if parse_version(self.conf.sources) <= parse_version(self.spec_file.get_version()):
```
nforro: Beware that `self.conf.sources` doesn't have to be a version string, it can be filename of an source archive. See `SpecFile.set_version_using_archive()` below.

rebasehelper/application.py
nforro: You should also modify the commit message accordingly.

nforro:
If you want to write the commit message in present tense, you should use `now` - i.e. after the change.
So: `If local version is higher than (or equal to) requested version, rebase-helper now doesn't do anything.`.
But I think in this case this simple commit message would suffice: `Abort rebase if requested version is not newer than current`, or even: `Abort rebase if requested version is not newer`.

nforro:
Actually it was meant to be complete commit message, in this case subject only. Commit message should be as short as possible but contain all necessary information at the same time. It's unnecessary to write something in the subject and then the same information in different words in the body.

PR #373, opened 2017-10-30
Add lint tasks to tox.

junaruga:
The purpose is to run lint tasks casually in development.
I added `tox -e lint-py2` and `tox -e lint-py3`.
This modification is also useful to optimize `pylintrc` disable's items by running the lint tasks repeatedly.

This PR's lint tasks are inspired from https://github.com/pypa/pip/blob/master/tox.ini .

How?


tox.ini
```diff
@@ -25,3 +25,22 @@ basepython=python2
 
 [testenv:py3]
 basepython=python3
+
+[lint]
+skip_install=true
+commands=
+    pylint --rcfile=pylintrc rebasehelper
+deps=
+    pylint>=1.4.5
```
nforro: How about creating `lint-requirements.txt` file and reference it here and also include it in `test-requirements.txt`?

junaruga: That is a great idea. The duplication code "pylint>=1.4.5" was my concern.
Let me try it.

test-requirements.txt
```diff
@@ -3,5 +3,5 @@ cov-core
 coverage
 pytest-cov
 pytest
-pylint>=1.4.5
+-r lint-requirements.txt
```
nforro: I think it would look a bit better if this was the last line in the file. Also, maybe it's not really necessary to include it here, it could be put into `deps` in `[testenv]`. It would perhaps make more sense, as linting and testing are different things.

junaruga:
The reason that this PR's travis test has not been finished is because some travis tests for my old PRs of "Fix to get git config user.name and user.email correctly." are running.
You can cancel those, if you like.
https://travis-ci.org/rebase-helper/rebase-helper/pull_requests

Issue #294, opened 2017-08-01
Make arguments more user friendly


nforro:
* add short alternatives to arguments
* add possibility to specify arguments in config file

junaruga:
I want `rebase-helper --version` option to show the version.

PR #377, opened 2017-11-03
Merge CLI arguments with config arguments

FrNecas:
There is now a parameter --conf to set path to a configfile.
If the path is not set, $HOME/.config/rebase-helper is used by default.

Config lines starting with # are ignored.
CLI arguments have priority over config arguments if they are set manually.
If there is not a certain config argument, CLI argument is used.

Resolves: #294

### DIFF
--- a/rebasehelper/cli.py
+++ b/rebasehelper/cli.py
@@ -34,6 +34,7 @@ from rebasehelper.application import Application
 from rebasehelper.logger import logger, LoggerHelper
 from rebasehelper.exceptions import RebaseHelperError
 from rebasehelper.utils import KojiHelper, ConsoleHelper
+from rebasehelper.config import Conf
 
 
 class CustomHelpFormatter(argparse.HelpFormatter):
@@ -163,14 +164,17 @@ class CliHelper(object):
                 raise RebaseHelperError('Wrong format of --builder-options. It must be in the following form:'
                                         ' --builder-options="--desired-builder-option".')
             cli = CLI()
-            if cli.version:
+            if hasattr(cli, 'version'):
                 logger.info(VERSION)
                 sys.exit(0)
-            ConsoleHelper.use_colors = ConsoleHelper.should_use_colors(cli)
-            execution_dir, results_dir, debug_log_file = Application.setup(cli)
-            if not cli.verbose:
+
+            conf = Conf(getattr(cli, 'conf', None))
+            conf.merge(cli)
+            ConsoleHelper.use_colors = ConsoleHelper.should_use_colors(conf)
+            execution_dir, results_dir, debug_log_file = Application.setup(conf)
+            if not conf.verbose:
                 handler.setLevel(logging.INFO)
-            app = Application(cli, execution_dir, results_dir, debug_log_file)
+            app = Application(conf, execution_dir, results_dir, debug_log_file)
             app.run()
         except KeyboardInterrupt:
             logger.info('\nInterrupted by user')
--- a/rebasehelper/cli.py
+++ b/rebasehelper/cli.py
@@ -27,15 +27,12 @@ import sys
 
 import six
 
+from rebasehelper.options import OPTIONS, traverse_options
 from rebasehelper.constants import PROGRAM_DESCRIPTION, NEW_ISSUE_LINK
 from rebasehelper.version import VERSION
 from rebasehelper.application import Application
 from rebasehelper.logger import logger, LoggerHelper
 from rebasehelper.exceptions import RebaseHelperError
-from rebasehelper.build_helper import Builder, SRPMBuilder
-from rebasehelper.checker import checkers_runner
-from rebasehelper.output_tool import BaseOutputTool
-from rebasehelper.versioneer import versioneers_runner
 from rebasehelper.utils import KojiHelper, ConsoleHelper
 
 
@@ -117,163 +114,29 @@ class CLI(object):
     def build_parser():
         parser = CustomArgumentParser(description=PROGRAM_DESCRIPTION,
                                       formatter_class=CustomHelpFormatter)
-        parser.add_argument(
-            "--version",
-            default=False,
-            action="store_true",
-            help="show rebase-helper version and exit"
-        )
-        parser.add_argument(
-            "-v",
-            "--verbose",
-            default=False,
-            action="store_true",
-            help="be more verbose (recommended)"
-        )
+
         group = parser.add_mutually_exclusive_group()
-        group.add_argument(
-            "-b",
-            "--build-only",
-            default=False,
-            action="store_true",
-            help="only build SRPM and RPMs"
-        )
-        group.add_argument(
-            "--comparepkgs-only",
-            default=False,
-            dest="comparepkgs",
-            metavar="COMPAREPKGS_DIR",
-            help="compare already built packages, %(metavar)s must be a directory "
-                 "with the following structure: <dir_name>/{old,new}/RPM"
-        )
-        group.add_argument(
-            "-p",
-            "--patch-only",
-            default=False,
-            action="store_true",
-            help="only apply patches"
-        )
-        parser.add_argument(
-            "--buildtool",
-            choices=Builder.get_supported_tools(),
-            default=Builder.get_default_tool(),
-            help="build tool to use, defaults to %(default)s"
-        )
-        parser.add_argument(
-            "--srpm-buildtool",
-            choices=SRPMBuilder.get_supported_tools(),
-            default=SRPMBuilder.get_default_tool(),
-            help="SRPM build tool to use, defaults to %(default)s"
-        )
-        parser.add_argument(
-            "--pkgcomparetool",
-            choices=checkers_runner.get_supported_tools(),
-            default=checkers_runner.get_default_tools(),
-            type=lambda s: s.split(','),
-            help="set of tools to use for package comparison, defaults to %(default)s"
-        )
-        parser.add_argument(
-            "--outputtool",
-            choices=BaseOutputTool.get_supported_tools(),
-            default=BaseOutputTool.get_default_tool(),
-            help="tool to use for formatting rebase output, defaults to %(default)s"
-        )
-        parser.add_argument(
-            "--versioneer",
-            choices=versioneers_runner.get_available_versioneers(),
-            default=None,
-            help="tool to use for determining latest upstream version"
-        )
-        parser.add_argument(
-            "--not-download-sources",
-            default=False,
-            action="store_true",
-            help="do not download sources"
-        )
-        parser.add_argument(
-            "-w",
-            "--keep-workspace",
-            default=False,
-            action="store_true",
-            help="do not remove workspace directory after finishing"
-        )
-        parser.add_argument(
-            "-c",
-            "--continue",
-            default=False,
-            action="store_true",
-            dest='cont',
-            help="continue previously interrupted rebase"
-        )
-        parser.add_argument(
-            "--color",
-            choices=['always', 'never', 'auto'],
-            default='auto',
-            help="colorize the output, defaults to %(default)s"
-        )
-        parser.add_argument(
-            "--disable-inapplicable-patches",
-            default=False,
-            action="store_true",
-            dest='disable_inapplicable_patches',
-            help="disable inapplicable patches in rebased SPEC file"
-        )
-        parser.add_argument(
-            "--non-interactive",
-            default=False,
-            action="store_true",
-            dest='non_interactive',
-            help="do not interact with user"
-        )
-        parser.add_argument(
-            "--build-tasks",
-            dest="build_tasks",
-            metavar="OLD_TASK,NEW_TASK",
-            type=lambda s: s.split(','),
-            help="comma-separated remote build task ids"
-        )
-        parser.add_argument(
-            "--builds-nowait",
-            default=False,
-            action="store_true",
-            help="do not wait for remote builds to finish"
-        )
-        parser.add_argument(
-            "--builder-options",
-            default=None,
-            metavar="BUILDER_OPTIONS",
-            help="enable arbitrary local builder option(s), enclose %(metavar)s in quotes "
-                 "and note that = before it is mandatory"
-        )
-        parser.add_argument(
-            "--srpm-builder-options",
-            default=None,
-            metavar="SRPM_BUILDER_OPTIONS",
-            help="enable arbitrary local srpm builder option(s), enclose %(metavar)s in quotes "
-                 "and note that = before it is mandatory"
-        )
-        parser.add_argument(
-            "--results-dir",
-            help="directory where rebase-helper output will be stored"
-        )
-        parser.add_argument(
-            "--get-old-build-from-koji",
-            default=False,
-            action="store_true",
-            help="do not build old sources, download latest build from Koji instead"
-        )
-        parser.add_argument(
-            "sources",
-            metavar='SOURCES',
-            nargs='?',
-            default=None,
-            help="new upstream sources"
-        )
-        parser.add_argument(
-            "--changelog-entry",
-            default="- New upstream release %{version}",
-            help="text to use as changelog entry, can contain RPM macros, which will be expanded"
-        )
+        current_group = 0
+        for option in traverse_options(OPTIONS):
+            option_kwargs = dict(option)
+            for key in ("group", "default", "name"):
+                if key in option_kwargs:
+                    del option_kwargs[key]
+
+            if "group" in option:
+                if current_group != option["group"]:
+                    current_group = option["group"]
+                    group = parser.add_mutually_exclusive_group()
+                actions_container = group
+            else:
+                actions_container = parser
+
+            # default is set to SUPPRESS to prevent arguments which were not specified on command line from being
+            # added to namespace. This allows rebase-helper to determine which arguments are used with their
+            # default value. This is later used for merging CLI arguments with config.
+            actions_container.add_argument(*option["name"], action=CustomAction, default=argparse.SUPPRESS,
+                                           actual_default=option.get("default"), **option_kwargs)
+
         return parser
 
     def __init__(self, args=None):
--- a/rebasehelper/cli.py
+++ b/rebasehelper/cli.py
@@ -50,10 +50,42 @@ class CustomHelpFormatter(argparse.HelpFormatter):
         return re.sub(r' ((SRPM_)?BUILDER_OPTIONS)', r'=\1', text)
 
     def _expand_help(self, action):
+        action.default = getattr(action, 'actual_default', None)
         if isinstance(action.default, list):
             default_str = ','.join([str(c) for c in action.default])
             action.default = default_str
         return super(CustomHelpFormatter, self)._expand_help(action)
+
+
+class CustomAction(argparse.Action):
+    def __init__(self, option_strings,
+                 switch=False,
+                 actual_default=None,
+                 dest=None,
+                 default=None,
+                 nargs=None,
+                 required=False,
+                 type=None,
+                 metavar=None,
+                 help=None,
+                 choices=None):
+
+        super(CustomAction, self).__init__(
+            option_strings=option_strings,
+            dest=dest,
+            default=default,
+            required=required,
+            metavar=metavar,
+            type=type,
+            help=help,
+            choices=choices)
+
+        self.switch = switch
+        self.nargs = 0 if self.switch else nargs
+        self.actual_default = actual_default
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        setattr(namespace, self.dest, True if self.switch else values)
 
 
 class CustomArgumentParser(argparse.ArgumentParser):
--- a/rebasehelper/config.py
+++ b/rebasehelper/config.py
@@ -1,0 +1,66 @@
+# -*- coding: utf-8 -*-
+#
+# This tool helps you to rebase package to the latest version
+# Copyright (C) 2013-2014 Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# he Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Authors: Petr Hracek <phracek@redhat.com>
+#          Tomas Hozza <thozza@redhat.com>
+
+import os.path
+
+from six.moves import configparser
+
+from rebasehelper.options import OPTIONS, traverse_options
+from rebasehelper.constants import CONFIG_FILE
+
+
+class Conf:
+    def __init__(self, config_file):
+        self.path_to_config = self.get_conf_path(config_file)
+        self.config = self.get_config()
+
+    @staticmethod
+    def get_conf_path(config_file):
+        if not config_file:
+            path = os.environ.get('XDG_CONFIG_HOME', os.path.expandvars('$HOME/.config'))
+            config_file = os.path.join(path, CONFIG_FILE)
+            return config_file
+        return os.path.abspath(config_file)
+
+    def get_config(self):
+        conf = {}
+        if os.path.isfile(self.path_to_config):
+            config = configparser.ConfigParser()
+            config.read(self.path_to_config)
+            for section in config.sections():
+                conf.update({k.replace('-', '_'): v for k, v in config.items(section)})
+
+        return conf
+
+    def merge(self, cli):
+        self.config.update(vars(cli.args))
+
+        for option in traverse_options(OPTIONS):
+            args = [n.lstrip('-').replace('-', '_') for n in option['name'] if n.startswith('--')]
+            if args:
+                dest = args[0]
+
+            if dest and dest not in self.config:
+                self.config[dest] = option.get('default')
+
+    def __getattr__(self, name):
+        return self.config.get(name)
--- a/rebasehelper/constants.py
+++ b/rebasehelper/constants.py
@@ -22,3 +22,4 @@
 
 PROGRAM_DESCRIPTION = "Tool to help package maintainers rebase their packages to the latest upstream version"
 NEW_ISSUE_LINK = 'https://github.com/rebase-helper/rebase-helper/issues/new'
+CONFIG_FILE = "rebase-helper.cfg"
--- a/rebasehelper/options.py
+++ b/rebasehelper/options.py
@@ -1,0 +1,199 @@
+# -*- coding: utf-8 -*-
+#
+# This tool helps you to rebase package to the latest version
+# Copyright (C) 2013-2014 Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# he Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Authors: Petr Hracek <phracek@redhat.com>
+#          Tomas Hozza <thozza@redhat.com>
+
+from rebasehelper.build_helper import Builder, SRPMBuilder
+from rebasehelper.checker import checkers_runner
+from rebasehelper.output_tool import BaseOutputTool
+from rebasehelper.versioneer import versioneers_runner
+
+OPTIONS = [
+    {
+        "name": ("--version",),
+        "default": False,
+        "switch": True,
+        "help": "show rebase-helper version and exit",
+    },
+    {
+        "name": ("-v", "--verbose"),
+        "default": False,
+        "switch": True,
+        "help": "be more verbose (recommended)",
+    },
+    [
+        {
+            "name": ("-b", "--build-only"),
+            "default": False,
+            "switch": True,
+            "help": "only build SRPM and RPMs",
+        },
+        {
+            "name": ("--comparepkgs-only",),
+            "default": False,
+            "dest": "comparepkgs",
+            "metavar": "COMPAREPKGS_DIR",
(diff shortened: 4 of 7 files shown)


rebasehelper/cli.py
```diff
@@ -49,12 +49,51 @@ def _format_action_invocation(self, action):
         return text.replace(' BUILDER_OPTIONS', '=BUILDER_OPTIONS')
 
     def _expand_help(self, action):
+        if hasattr(action, "actual_default"):
+            action.default = action.actual_default
```
nforro: You could just use:
```python
action.default = getattr(action, 'actual_default', None)
```

Also, it would be nice to honor quoting style, i.e. use single quotes like everywhere else in this class.

rebasehelper/cli.py
```diff
@@ -49,12 +49,51 @@ def _format_action_invocation(self, action):
...
+            metavar=metavar,
+            type=type,
+            help=help,
+            choices=choices)
+
+        self.switch = switch
+        if self.switch:
+            self.nargs = 0
+        else:
+            self.nargs = nargs
```
nforro: This could be shortened to:
```python
self.nargs = nargs if not self.switch else 0
```
(or the other way around).

rebasehelper/cli.py
```diff
@@ -49,12 +49,51 @@ def _format_action_invocation(self, action):
...
+        self.actual_default = actual_default
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        if self.switch:
+            setattr(namespace, self.dest, True)
+        else:
+            setattr(namespace, self.dest, values)
```
nforro: Again, this could be written also as:
```python
setattr(namespace, self.dest, values if not self.switch else True)
```
But it's a matter of preference, if you think your way is more readable, keep it that way.

rebasehelper/cli.py
```diff
@@ -26,17 +26,16 @@
...
 from rebasehelper.application import Application
 from rebasehelper.logger import logger, LoggerHelper
 from rebasehelper.exceptions import RebaseHelperError
-from rebasehelper.build_helper import Builder, SRPMBuilder
-from rebasehelper.checker import checkers_runner
-from rebasehelper.output_tool import BaseOutputTool
-from rebasehelper.versioneer import versioneers_runner
 from rebasehelper.utils import KojiHelper, ConsoleHelper
 
+SUPPRESS = argparse.SUPPRESS
```
nforro: What's the point of this?

rebasehelper/cli.py
```diff
@@ -123,168 +122,37 @@ class CLI(object):
...
+        for node in OPTIONS:
+            if type(node) is list:
+                group = parser.add_mutually_exclusive_group()
+                for option in node:
+                    option_args = dict(option)
+                    for key in ("name", "default"):
+                        if key in option_args.keys():
+                            del option_args[key]
+                    group.add_argument(*option["name"], **option_args, action=CustomAction, default=SUPPRESS,
+                                       actual_default=option["default"] if "default" in option.keys() else None)
```
nforro: This block (inside the `for` loop) is the same as below, if you want to keep it like this, you should define a function/method to avoid duplicate code. But it might be better to traverse options recursively.

rebasehelper/cli.py
```diff
@@ -123,168 +122,37 @@ class CLI(object):
...
-            default="- New upstream release %{version}",
-            help="text to use as changelog entry, can contain RPM macros, which will be expanded"
-        )
+        for node in OPTIONS:
+            if type(node) is list:
+                group = parser.add_mutually_exclusive_group()
+                for option in node:
+                    option_args = dict(option)
+                    for key in ("name", "default"):
+                        if key in option_args.keys():
```
nforro: This is the same as `if key in option_args`.

rebasehelper/cli.py
nforro: And this:
```python
actual_default=option["default"] if "default" in option.keys() else None
```
could be written like this:
```python
actual_default=option.get('default')
```

rebasehelper/cli.py
```diff
@@ -123,168 +122,37 @@ class CLI(object):
...
+                option_args = dict(node)
+                for key in ("name", "default"):
+                    if key in option_args.keys():
+                        del option_args[key]
+                parser.add_argument(*node["name"], **option_args, action=CustomAction, default=SUPPRESS,
+                                    actual_default=node["default"] if "default" in node.keys() else None)
         return parser
 
+    @staticmethod
+    def remove_groups():
```
nforro: What's the point of this method?

rebasehelper/options.py
```diff
@@ -0,0 +1,136 @@
...
+#
+# Authors: Petr Hracek <phracek@redhat.com>
+#          Tomas Hozza <thozza@redhat.com>
+
+from rebasehelper.build_helper import Builder, SRPMBuilder
+from rebasehelper.checker import checkers_runner
+from rebasehelper.output_tool import BaseOutputTool
+from rebasehelper.versioneer import versioneers_runner
+
+OPTIONS = [
```
nforro: It would be probably better to indent a bit more to improve readability, e.g. like this:
```python
OPTIONS = [
    {
        "name": ("--version",),
        "default": False,
        "switch": True,
        "help": "show rebase-helper version and exit"
    },
    {
        "name": ("-v", "--verbose"),
        "default": False,
        "switch": True,
        "help": "be more verbose (recommended)"
    },
    [
        {
            "name": ("-b", "--build-only"),
            "default": False,
            "switch": True,
            "help": "only build SRPM and RPMs"
        },
        ...
    ],
    ...
]
```

rebasehelper/cli.py
```diff
@@ -177,11 +178,14 @@ def run():
             if cli.version:
                 logger.info(VERSION)
                 sys.exit(0)
-            ConsoleHelper.use_colors = ConsoleHelper.should_use_colors(cli)
-            execution_dir, results_dir, debug_log_file = Application.setup(cli)
-            if not cli.verbose:
+
+            conf = Conf(cli)
+            conf.merge()
```
nforro: It would make more sense to me to use this kind of workflow:
```python
conf = Conf(cli.conf)
conf.merge(cli)
```

rebasehelper/config.py
```diff
@@ -0,0 +1,83 @@
...
+                if "name" in option.keys():
+                    for item in option["name"]:
+                        if item == "--conf":
+                            return option
+
+        if "conf" in vars(self.cli.args).keys():
+            path_to_conf = os.path.abspath(self.cli.args.conf)
+        else:
+            conf_option = get_conf_option()
+            path_to_conf = conf_option["default"]
```
nforro: I don't think default path should be defined in the option. Of course, it works as well, but I think extracting the value is too complicated.

rebasehelper/config.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def get_config(self):
+        conf = {}
+        if os.path.isfile(self.path_to_config):
+            config = configparser.ConfigParser()
+            config.read(self.path_to_config)
+            for section in config.sections():
+                for key, val in config.items(section):
+                    dest = key.replace("-", "_")
+                    conf[dest] = val
```
nforro: This could be written as dict comprehension:
```python
conf.update({k.replace('-', '_'): v for k, v in config.items(section)})
```

rebasehelper/config.py
```diff
@@ -0,0 +1,83 @@
...
+        else:
+            logger.info('Could not find any set or default config file')
+
+        return conf
+
+    def merge(self):
+        for key, value in vars(self.cli.args).items():
```
nforro: You shoud use `six.iteritems()` (it returns iterator in python2, as opposed to `items()` method).

rebasehelper/config.py
```diff
@@ -0,0 +1,83 @@
...
+            self.config[key] = value
```
nforro: Anyway, the loop could be replaced with:
```python
self.config.update(vars(self.cli.args))
```

rebasehelper/config.py
```diff
@@ -0,0 +1,83 @@
...
+
+        for option in OPTIONS:
+            if "dest" in option.keys():
+                dest = option["dest"]
+            else:
+                dest = max(option["name"], key=len)
```
nforro: You should rather find first option string starting with `--`.

rebasehelper/config.py
```diff
@@ -0,0 +1,83 @@
...
+
+        return conf
+
+    def merge(self):
+        for key, value in vars(self.cli.args).items():
+            self.config[key] = value
+
+        for option in OPTIONS:
+            if "dest" in option.keys():
+                dest = option["dest"]
```
nforro: This could be written as:
```python
dest = option.get('dest')
if not dest:
    ...
```

rebasehelper/config.py
```diff
@@ -0,0 +1,83 @@
...
+            else:
+                dest = max(option["name"], key=len)
+                dest = dest.lstrip("-").replace("-", "_")
+
+            if dest not in self.config.keys():
```
nforro: This is the same as `if dest not in self.config`.

rebasehelper/config.py
```diff
@@ -0,0 +1,83 @@
...
+                self.config[dest] = option.get("default")
+
+    def __getattr__(self, name):
+        try:
+            self.config[name]
```
nforro: There is no `return` here.

nforro: And you don't have to catch `KeyError`, `self.config.get(name)` will handle it for you.

nforro:
Also the commit messages deserve a bit more love. At least the first one contains multiple typos.

rebasehelper/cli.py
```diff
@@ -84,168 +123,37 @@ class CLI(object):
...
+        for node in OPTIONS:
+            if type(node) is list:
+                group = parser.add_mutually_exclusive_group()
+                for option in node:
+                    option_args = dict(option)
+                    for key in ("name", "default"):
+                        if key in option_args.keys():
+                            del option_args[key]
+                    group.add_argument(*option["name"], **option_args, action=CustomAction, default=SUPPRESS,
+                                       actual_default=option["default"] if "default" in option.keys() else None)
```
nforro: Also, you have to put `**option_args` at the end, after other keyword arguments, as this raises `SyntaxError` in python2 (see https://readthedocs.org/projects/rebase-helper/builds/6258197/).

rebasehelper/cli.py
```diff
@@ -116,163 +114,26 @@ class CLI(object):
...
+            option_kwargs = dict(option)
+            for key in ("group", "default", "name"):
+                if key in option_kwargs:
+                    del option_kwargs[key]
+
+            if "group" in option:
+                if previous_group != option["group"]:
+                    group = parser.add_mutually_exclusive_group()
+                actions_container = group
+                previous_group = option["group"]
```
nforro: This could be moved two lines up, it is more logical that way IMHO.

nforro: Also the variable could be named `current_group`, it makes more sense since you are counting the groups from 0.

rebasehelper/cli.py
```diff
@@ -116,163 +114,26 @@ class CLI(object):
...
+            else:
+                actions_container = parser
+
+            actions_container.add_argument(*option["name"], action=CustomAction, default=argparse.SUPPRESS,
```
nforro: It would be nice to add a comment here explaining why `SUPPRESS` is used.

rebasehelper/options.py
```diff
@@ -0,0 +1,199 @@
...
+        "metavar": "SOURCES",
+        "nargs": "?",
+        "default": None,
+        "help": "new upstream sources"
+    },
+    {
+        "name": ("--changelog-entry",),
+        "default": "- New upstream release %{version}",
+        "help": "text to use as changelog entry, can contain RPM macros, which will be expanded"
+    }
```
nforro: There should always be a comma after last item of a list. It has no disadvantages and enables you to add more items without caring whether there is a comma or not. Also there are no unnecessary comma adding lines in diff.
The same applies of course for dict items.

rebasehelper/options.py
```diff
@@ -0,0 +1,199 @@
...
+]
+
+
+def traverse(options):
```
nforro: The name is too generic, something like `traverse_options` would be better. Or you can keep it like this, but you shouldn't import it directly to a namespace and rather use it as `options.traverse()`.

rebasehelper/options.py
```diff
@@ -0,0 +1,199 @@
...
+    group_index = 0
+    for opt in options:
+        if isinstance(opt, list):
+            for inner_opt in traverse(opt):
```
nforro: The recursion is unnecessary, it wouldn't work anyway since you are not passing the group index.

rebasehelper/cli.py
```diff
@@ -160,14 +160,17 @@ def run():
                 raise RebaseHelperError('Wrong format of --builder-options. It must be in the following form:'
                                         ' --builder-options="--desired-builder-option".')
             cli = CLI()
-            if cli.version:
+            if hasattr(cli, "version"):
```
nforro: You obviously prefer double quotes, that's fine, but you should keep the quoting style consistent among code blocks.
With the exception of `"--builder-options"` (which should be fixed) every string in this method uses single quotes.

rebasehelper/cli.py
```diff
@@ -160,14 +160,17 @@ def run():
...
                 logger.info(VERSION)
                 sys.exit(0)
-            ConsoleHelper.use_colors = ConsoleHelper.should_use_colors(cli)
-            execution_dir, results_dir, debug_log_file = Application.setup(cli)
-            if not cli.verbose:
+
+            conf = Conf(getattr(cli, "default", None))
```
nforro: Why `default`? Shouldn't it be `conf`?

rebasehelper/config.py
```diff
@@ -0,0 +1,81 @@
...
+class Conf:
+    def __init__(self, path_to_conf):
+        self.path_to_config = self.get_conf_path(path_to_conf)
+        self.config = self.get_config()
+
+    def get_conf_path(self, path_to_conf):
+        default_options = ("$XDG_CONFIG_HOME/rebase-helper.cfg", "$HOME/.config/rebase-helper.cfg")
+
+        if path_to_conf is None:
+            if os.path.isfile(default_options[0]):
```
nforro: This won't work, you have to expand the environment variables first. You could use [`os.path.expandvars`](https://docs.python.org/2/library/os.path.html#os.path.expandvars) for that, but I think better approach would be to expand `XDG_CONFIG_HOME` first, defaulting to `$HOME/.config`, and then use the expanded value.

Try this for example:
```python
path = os.environ.get('XDG_CONFIG_HOME', os.path.expandvars('$HOME/.config'))
config_file = os.path.join(path, 'rebase-helper.cfg')
```

rebasehelper/config.py
```diff
@@ -0,0 +1,81 @@
...
+                relative_path = default_options[0]
+
+            else:
+                relative_path = default_options[1]
+
+        else:
+            relative_path = os.path.abspath(self.cli.args.conf)
```
nforro: Why are you totally ignoring the path passed to the constructor and instead are reading it directly from `self.cli` (which doesn't even exist)?

rebasehelper/config.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def merge(self, cli):
+        self.config.update(vars(cli.args))
+
+        for option in traverse(OPTIONS):
+            dest = option.get("dest")
+            if not dest:
+                for name in option["name"]:
+                    if name.startswith("--"):
+                        dest = name
```
nforro: You should break here, otherwise you will be assigning the last name, not the first. Also, `"name"` is probably not the best expression, if you are iterating through it. It should be plural.

rebasehelper/config.py
```diff
@@ -0,0 +1,81 @@
...
+
+            if dest:
+                dest = dest.lstrip("-").replace("-", "_")
+
+            if dest not in self.config and dest:
```
nforro: This should be swapped, first you want to know whether `dest` is not `None`, and only then (if it isn't) you want to find it in `self.config`.

rebasehelper/options.py
```diff
@@ -184,6 +184,10 @@
         "name": ("--changelog-entry",),
         "default": "- New upstream release %{version}",
         "help": "text to use as changelog entry, can contain RPM macros, which will be expanded"
+    },
+    {
+        "name": ("--conf",),
+        "help": "path to config with preset arguments"
```
nforro: What is `config with preset arguments`? Did you mean `predefined`? Anyway, I think `custom path to configuration file` or `load specified configuration file instead of a default` would be better.

rebasehelper/options.py
nforro: Another option is to create a class and embed both `OPTIONS` and `traverse` into it.

rebasehelper/config.py
nforro: Also, the name of the config file should be defined as a constant somewhere, e.g. in `rebasehelper/constants.py` or `rebasehelper/settings.py` (BTW these two files should be probably merged together, but that's not in scope of this PR).

rebasehelper/options.py
```diff
@@ -0,0 +1,199 @@
...
+        "choices": ["always", "never", "auto"],
+        "default": "auto",
+        "help": "colorize the output, defaults to %(default)s"
+    },
+    {
+        "name": ("--disable-inapplicable-patches",),
+        "default": False,
+        "switch": True,
+        "dest": "disable_inapplicable_patches",
+         "help": "disable inapplicable patches in rebased SPEC file"
```
nforro: You have wrong indentation here.

nforro:
To nitpick the commit messages:

>**Modify Action and HelpFormatter classes**

While technically not incorrect, this subject is not very informative nor accurate. I think someting like **Add CustomAction class** would be better.
>
>CustomAction has 2 new arguments, actual_default and switch.

You probably mean `attributes`, not `arguments`.

>Default is set to SUPPRESS to determine which options were set by default.
>CustomHelpFormatter then expands help using actual_default value.

You shoud use `default`, not `Default`, since you are referring to literal attribute. And I think the following text describes the change better:
```
"default" attribute of the action is set to SUPPRESS, to prevent the action
appearing in a namespace if not explicitly set. CustomHelpFormatter._expand_help
uses "actual_default" instead.
```

>
>If switch is set to True, CustomAction behaves as StoreTrueAction of
>argparse.

I would write `behaves as argparse._StoreTrueAction`.

>**Rewrite CLI options using tree structure***

This sound better: **Move definitions of program options to a separate file**.
>
>Parser options are now defined in separate file using tree
>structure. Options in a list are put into mutually exclusive group.
>

I don't think it is necessary to write anything more about it.
>Options are using custom Action class, see commit
>5f0f439

You don't have to reference previous commit, especially using hash which can (and will) eventually change.

>**Merge CLI arguments with config arguments**
>

This doesn't describe the nature of the change. I would write something like **Introduce Conf class for reading program options from config file**.
>There is now a parameter --conf to set path to a config file.
>If the path is not set, $XDG_CONFIG_HOME/rebase-helper.cfg is
>checked first then $HOME/.config/rebase-helper is checked.

My version would be:
```
Configuration file is searched in $XDG_CONFIG_HOME, but can be set
explicitly with --conf option.
```
>
>Config lines starting with # are ignored.
>CLI arguments have priority over config arguments if they are set manually.
>If there is not a certain config argument, CLI argument is used.

I really don't understand this last line. I'd write simply something like this:
```
Options specified on the command line override those from config file.
```

rebasehelper/cli.py
```diff
@@ -117,163 +114,28 @@ class CLI(object):
...
+
+            if "group" in option:
+                if current_group != option["group"]:
+                    current_group = option["group"]
+                    group = parser.add_mutually_exclusive_group()
+                actions_container = group
+            else:
+                actions_container = parser
+
+            # default is set to SUPPRESS to prevent arguments which were not added explicitly from being
```
nforro: What does `were not added explicitly` mean? Added where?

rebasehelper/cli.py
```diff
@@ -117,163 +114,28 @@ class CLI(object):
...
+            # added to namespace. This allows rebase-helper to determine which arguments were used by default.
```
nforro: > which arguments were used by default

Again, what does `used` mean? Used by what, why?

rebasehelper/cli.py
```diff
@@ -162,14 +163,18 @@ def run():
...
-            if not cli.verbose:
+
+            conf = Conf(getattr(cli, 'conf', None))
+            conf.merge(cli)
+            ConsoleHelper.use_colors = ConsoleHelper.should_use_colors(conf)
+            execution_dir, results_dir, debug_log_file = Application.setup(conf)
+            if not conf.verbose:
                 handler.setLevel(logging.INFO)
-            app = Application(cli, execution_dir, results_dir, debug_log_file)
+            print(conf.config)
```
nforro: I guess this was supposed to be removed?

rebasehelper/config.py
```diff
@@ -0,0 +1,75 @@
...
+
+class Conf:
+    def __init__(self, config_file):
+        self.path_to_config = self.get_conf_path(config_file)
+        self.config = self.get_config()
+
+    @staticmethod
+    def get_conf_path(config_file):
+        if not config_file:
+            path = os.environ.get(DEFAULT_CONFIG_LOCATIONS[0], os.path.expandvars(DEFAULT_CONFIG_LOCATIONS[1]))
```
nforro: This is not what I had in mind. Defining this as constants is unnecessary and actually much less readable. Also those two constants do not have the same kind of content, the first one is name of an environment variable while the second one is a path.

rebasehelper/config.py
```diff
@@ -0,0 +1,75 @@
...
+            config_file = os.path.join(path, 'rebase-helper.cfg')
```
nforro: >'rebase-helper.cfg'

This is what should be defined as a constant.

rebasehelper/config.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def get_config(self):
+        conf = {}
+        if os.path.isfile(self.path_to_config):
+            config = configparser.ConfigParser()
+            config.read(self.path_to_config)
+            for section in config.sections():
+                conf.update({k.replace('-', '_'): v for k, v in config.items(section)})
+        else:
+            logger.info('Could not find any set or default config file')
```
nforro: I don't think it's necessary to mention this. Some message could be printed in case an user specifies path to a config file using `--conf` and it is not readable/parseable, but that's not the case here.

rebasehelper/config.py
```diff
@@ -0,0 +1,75 @@
...
+    def merge(self, cli):
+        self.config.update(vars(cli.args))
+
+        for option in traverse_options(OPTIONS):
+            dest = option.get('dest')
+            if not dest:
+                for name in option['name']:
+                    if name.startswith('--'):
+                        dest = name
+                        break
```
nforro: This loop could be also replaced with something like this:
```python
args = [n for n in option['name'] if n.startswith('--')]
if args:
    dest = args[0]
```

You could even do the stripping and dash-replacing inside the list comprehension.

rebasehelper/constants.py
```diff
@@ -22,3 +22,4 @@
 
 PROGRAM_DESCRIPTION = "Tool to help package maintainers rebase their packages to the latest upstream version"
 NEW_ISSUE_LINK = 'https://github.com/rebase-helper/rebase-helper/issues/new'
+DEFAULT_CONFIG_LOCATIONS = ("$XDG_CONFIG_HOME", "$HOME/.config")
```
nforro: As written above, this is not a good idea. Besides you have it wrong, it should be `"XDG_CONFIG_HOME"`, not `"$XDG_CONFIG_HOME"`.

You could define a constant for the resulting path (and move the expanding logic here).

rebasehelper/options.py
```diff
@@ -185,14 +185,18 @@
...
+    },
 ]
 
 
 def traverse_options(options):
     group_index = 0
     for opt in options:
         if isinstance(opt, list):
-            for inner_opt in traverse(opt):
+            for inner_opt in opt:
```
nforro: This change should be part of the previous commit.

nforro:
Again some notes about the commit messages:

>**Add CustomAction class**
>
>CustomAction has 2 new arguments, actual_default and switch.

I can still see the word `arguments` there. It would perhaps made sense if you write it as `Constructor of CustomAction has 2 new arguments`, but that's not accurate anyway and definitely not worth mentioning.
>default attribute of the action is set to SUPPRESS to prevent the action

`default` definitely needs to be quoted, otherwise it's really confusing. And if you quote `default` attribute, you have to quote all other attributes as well for the sake of consistency.
>from appearing in namespace if not set explicitly.
>CustomHelpFormatter._expand help uses "actual_default" instead.

What is `CustomHelpFormatter._expand help`? There should be an underscore instead of space. Also adding `()` at the end would make it obvious that it's actually a method.

>**Add Conf class for reading program options from config file**
>
>Configuartion file is searched in $XDG_CONFIG_HOME but can be set
>explicitly with --conf option

There is missing punctuation, why? Also, there is a typo in `Configuartion`. And `--conf` could/should be quoted.
>
>Options specified from command line override those from config file.

This doesn't make sense, options can't be *specified from* the command line. They could perhaps be *gathered* or *collected from* the command line, but they can definitely be *specified on* the command line.

nforro:
So the changes look good to me now, only thing that is missing are some tests for the config part.

rebasehelper/tests/test_conf.py
```diff
@@ -0,0 +1,86 @@
...
+        cli = CLI(cli_args)
+        conf = Conf(self.path_to_conf)
+        conf.merge(cli)
+        for key, value in six.iteritems(expected_output):
+            assert conf.config[key] == value
+
+        os.remove(self.path_to_conf)
+
+    @staticmethod
+    def create_config(args, path_to_conf):
```
nforro: This would be better to implement as a fixture, see this for inspiration: https://github.com/rebase-helper/rebase-helper/blob/master/rebasehelper/tests/test_archive.py#L51-L56

Also, you don't have to remove the file, as it exists in temporary directory.

rebasehelper/tests/test_conf.py
```diff
@@ -0,0 +1,86 @@
...
+import os
+import pytest
+import six
+
+from rebasehelper.cli import CLI
+from rebasehelper.config import Conf
+
+
+class TestConfig(object):
+    path_to_conf = os.path.abspath('test_conf.cfg')
```
nforro: Why `os.path.abspath`?

rebasehelper/tests/test_conf.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @pytest.mark.parametrize('args', [
+        {
+            'changelog-entry': 'Updated to',
+            'versioneer': 'pypi',
+         },
+        {},
+    ], ids=[
+        'filled',
```
nforro: `configured` (or `preconfigured`) sounds better to me.

rebasehelper/tests/test_conf.py
```diff
@@ -0,0 +1,86 @@
...
+         },
+         {
+             'changelog-entry': 'Version set to',
+             'versioneer': 'pypi',
+             'buildtool': 'rpmbuild',
+             'color': 'auto',
+         }), ], ids=[
+            'CLI with config',
+        ]
+    )
```
nforro: The indentation is a bit inconsistent, it should look like this:
```python
@pytest.mark.parametrize('cli_args, conf_args, merged', [
    (
        [
            '--changelog-entry', 'Version set to',
            '--buildtool', 'rpmbuild',
        ],
        {
            'changelog-entry': 'Updated to ',
            'versioneer': 'pypi',
        },
        {
            'changelog-entry': 'Version set to',
            'versioneer': 'pypi',
            'buildtool': 'rpmbuild',
            'color': 'auto',
        },
    ),
], ids=[
    'CLI with config',
])
```

Do not put `,` after last item in a dict or list unless the items are on their own lines.

rebasehelper/tests/test_conf.py
```diff
@@ -0,0 +1,86 @@
...
+#          Tomas Hozza <thozza@redhat.com>
+
+import os
+import pytest
+import six
+
+from rebasehelper.cli import CLI
+from rebasehelper.config import Conf
+
+CONF_FILE = 'test_conf.cfg'
```
nforro: This should be made class attribute, to stay consistent with other tests.

rebasehelper/tests/test_conf.py
nforro: The indentation is still wrong, and inconsistent even with what you have just a few lines up.

rebasehelper/tests/test_conf.py
```diff
@@ -0,0 +1,86 @@
...
+    def test_merge(self, cli_args, merged, config_file):
+        expected_output = {x.replace('-', '_'): merged[x] for x in merged}
+        cli = CLI(cli_args)
+        conf = Conf(config_file)
+        conf.merge(cli)
+        for key, value in six.iteritems(expected_output):
+            assert conf.config[key] == value
+
+    @pytest.fixture
+    def config_file(self, conf_args):
```
nforro: The fixture should be the first method in the class. It's not a rule, but I find it more logical as it is a prerequisite for the tests.

rebasehelper/tests/test_conf.py
```diff
@@ -0,0 +1,86 @@
...
+        config = six.moves.configparser.ConfigParser()
+        config['SECTION1'] = {}
+        for key, value in six.iteritems(conf_args):
+            config['SECTION1'][key] = value
```
nforro: Why iterate through `conf_args` when you can just assign it directly?
```python
config['SECTION1'] = conf_args
```

nforro: Actually, scratch that. `ConfigParser` in python2 doesn't support direct assignment, you have to use [legacy API](https://docs.python.org/3/library/configparser.html#legacy-api-examples).

rebasehelper/tests/test_conf.py
```diff
@@ -0,0 +1,86 @@
...
+            'versioneer': 'pypi',
+        },
+        {},
+    ], ids=[
+        'configured',
+        'empty',
+    ])
+    def test_get_config(self, conf_args, config_file):
+        conf = Conf(config_file)
+        expected_result = {key.replace('-', '_'): value for key, value in six.iteritems(conf_args)}
```
nforro: You could use just `k` and `v` for variable names in the dict comprehension, it is obvious what those mean and the line would be shorter.

rebasehelper/tests/test_conf.py
```diff
@@ -0,0 +1,86 @@
...
+                'buildtool': 'rpmbuild',
+                'color': 'auto',
+            }
+        ),
+    ], ids=[
+            'CLI with config',
+        ]
+    )
+    def test_merge(self, cli_args, merged, config_file):
+        expected_output = {x.replace('-', '_'): merged[x] for x in merged}
```
nforro: Why are you using different approach here from what you have in the other test? While this may be a bit faster, the point is that you should keep the code consistent.

rebasehelper/tests/test_conf.py
```diff
@@ -0,0 +1,86 @@
...
+        cli = CLI(cli_args)
+        conf = Conf(config_file)
+        conf.merge(cli)
+        for key, value in six.iteritems(expected_output):
+            assert conf.config[key] == value
```
nforro: Again, why are you iterating here and not comparing directly like in the other test?

PR #384, opened 2017-11-15
Add parentheses to format error message properly

FrNecas:
When the exception was raised, it resulted in an error: "not enough arguments for format string" because of the missing parentheses


nforro:
@FrNecas: there is one more occurence of the same problem, see commit c3f08998cce0a2e82e552d84b0791b1d27f1098c.

Issue #376, opened 2017-11-01
Modify rebase helper output


skisela:
```
$ rebase-helper --buildtool koji
...
...
rebase-helper failed due to an unexpected error. Please report this problem
'/home/skisela/fedora/rubygem-coffee-script-source/rebase-helper-results/rebase-helper-debug.log'
file in the report.
Thank you!
```
Maybe better wording would be to advise the user to check the debug log first and then create an issue if still desperate? Here the reason of the fail is quite obvious.

#### rebase-helper-results/rebase-helper-debug.log:
```
2017-11-01 09:40:30,950 DEBUG cli.py:293 run: 
Traceback (most recent call last):
  File "/home/skisela/packages/venv-rebase-helper/lib/python2.7/site-packages/rebasehelper/cli.py", line 267, in run
    app.run()
  File "/home/skisela/packages/venv-rebase-helper/lib/python2.7/site-packages/rebasehelper/application.py", line 805, in run
    build = self.build_packages()
  File "/home/skisela/packages/venv-rebase-helper/lib/python2.7/site-packages/rebasehelper/application.py", line 532, in build_packages
    build_dict.update(builder.build(spec, sources, patches, results_dir, **build_dict))
  File "/home/skisela/packages/venv-rebase-helper/lib/python2.7/site-packages/rebasehelper/build_helper.py", line 379, in build
    return self._tool.build(*args, **kwargs)
  File "/home/skisela/packages/venv-rebase-helper/lib/python2.7/site-packages/rebasehelper/build_tools/koji_tool.py", line 165, in build
    rpms, rpm_logs, koji_task_id = cls._scratch_build(srpm, **kwargs)
  File "/home/skisela/packages/venv-rebase-helper/lib/python2.7/site-packages/rebasehelper/build_tools/koji_tool.py", line 82, in _scratch_build
    session = cls.koji_helper.session_maker()
  File "/home/skisela/packages/venv-rebase-helper/lib/python2.7/site-packages/rebasehelper/utils.py", line 870, in session_maker
    koji_session.ssl_login(cls.cert, cls.ca_cert, cls.ca_cert)
  File "/usr/lib/python2.7/site-packages/koji/__init__.py", line 2273, in ssl_login
    sinfo = self.callMethod('sslLogin', proxyuser)
  File "/usr/lib/python2.7/site-packages/koji/__init__.py", line 2321, in callMethod
    return self._callMethod(name, args, opts)
[...]

PR #386, opened 2017-11-22
Handle SSL Exception when kerberos ticket is missing

skisela:
Base Exceptions are very non-informative when building packages. We should consider logging the errors into debug log.

Fixes: #376

### DIFF
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -584,9 +584,10 @@ class Application(object):
                     msg = 'Building {} RPM packages failed; see {} for more information'.format(version, e.logfile)
                 raise RebaseHelperError(msg, logfiles=builder.get_logs().get('logs'))
 
-            except Exception:
+            except Exception as e:
                 raise RebaseHelperError('Building package failed with unknown reason. '
-                                        'Check all available log files.')
+                                        'The error reported is "{}". '
+                                        'Check all available log files.'.format(six.text_type(e)))
 
         if self.conf.builds_nowait and not self.conf.build_tasks:
             if builder.creates_tasks():
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -558,6 +558,10 @@ class Application(object):
                     build_dict.pop(opt)
                 results_store.set_build_data(version, build_dict)
 
+            except RebaseHelperError:
+                # Proper RebaseHelperError instance was created already. Re-raise it.
+                raise
+
             except SourcePackageBuildError as e:
                 build_dict.update(builder.get_logs())
                 build_dict['source_package_build_error'] = six.text_type(e)
--- a/rebasehelper/utils.py
+++ b/rebasehelper/utils.py
@@ -901,9 +901,8 @@ class KojiHelper(object):
             return koji_session
         try:
             koji_session.krb_login()
-        except koji.krbV.Krb5Error:
-            # fall back to login using certificate
-            koji_session.ssl_login(cls.cert, cls.ca_cert, cls.ca_cert)
+        except koji.krbV.Krb5Error as e:
+            raise RebaseHelperError('Kerberos login failed. The error reported is: %s' % (six.text_type(e)))
         return koji_session
 
     @classmethod


rebasehelper/application.py
```diff
@@ -581,7 +581,8 @@ def build_packages(self):
                     msg = 'Building {} RPM packages failed; see {} for more information'.format(version, e.logfile)
                 raise RebaseHelperError(msg, logfiles=builder.get_logs().get('logs'))
 
-            except Exception:
+            except Exception as e:
+                logger.error(e)
```
nforro: Wouldn't it be better to append the exception message to the RebaseHelperError which is being raised anyway?

skisela: Probably yes. But isn't it better to have the errors in logs too for consistency?

skisela: Oh I see it already. You're right, your approach is sufficient.

nforro: Yes, of course. But that's already taken care of: https://github.com/rebase-helper/rebase-helper/blob/master/rebasehelper/cli.py#L282-L286

rebasehelper/application.py
```diff
@@ -581,9 +581,10 @@ def build_packages(self):
                     msg = 'Building {} RPM packages failed; see {} for more information'.format(version, e.logfile)
                 raise RebaseHelperError(msg, logfiles=builder.get_logs().get('logs'))
 
-            except Exception:
+            except Exception as e:
+                error_msg = '\n' + str(e)
                 raise RebaseHelperError('Building package failed with unknown reason. '
-                                        'Check all available log files.')
+                                        'Check all available log files.' + error_msg)
```
nforro: Hm, how about something like this?
```python
'Build failed with unknown reason. The error reported is "{}". Check all available log files.'.format(six.text_type(e))
```

rebasehelper/application.py
```diff
@@ -581,9 +581,10 @@ def build_packages(self):
                     msg = 'Building {} RPM packages failed; see {} for more information'.format(version, e.logfile)
                 raise RebaseHelperError(msg, logfiles=builder.get_logs().get('logs'))
 
-            except Exception:
-                raise RebaseHelperError('Building package failed with unknown reason. '
-                                        'Check all available log files.')
+            except Exception as e:
+                raise RebaseHelperError('Building package failed with unknown reason.'
+                                        'The error reported is "{}".'
+                                        'Check all available log files.'.format(str(e)))
```
nforro: Just a nitpick, it's probably safer to use `six.text_type` instead of `str`, like [here](https://github.com/rebase-helper/rebase-helper/blob/master/rebasehelper/cli.py#L185).

nforro:
I wonder how this change can fix issue #376. You only extended the error message, but the issue is about unhandled `SSLError` exception.

rebasehelper/build_tools/koji_tool.py
```diff
@@ -79,7 +81,11 @@ def creates_tasks(cls):
 
     @classmethod
     def _scratch_build(cls, source, **kwargs):
-        session = cls.koji_helper.session_maker()
+        try:
+            session = cls.koji_helper.session_maker()
+        except requests.exceptions.SSLError as e:
+            raise RebaseHelperError('{}\n{}'.format(six.text_type(e),
+                                                    'You probably do not own a correct active kerberos ticket.'))
```
nforro: This is not right. What actually happens is that [`koji_session.krb_login()`](https://github.com/rebase-helper/rebase-helper/blob/51755dd20b8240f7abe28676953d38315795eb51/rebasehelper/utils.py#L903) raises `koji.krbV.Krb5Error` and then falls back to authenticate using [`koji_session.ssl_login()`](https://github.com/rebase-helper/rebase-helper/blob/51755dd20b8240f7abe28676953d38315795eb51/rebasehelper/utils.py#L906), but that is long unsupported and you probably can't even get valid certificate to authenticate through it now.
It might be best to drop the fallback entirely and instead raise `RebaseHelperError` with an explanation.

nforro: Also, there is currently plan to deprecate `python-krbV` in Fedora, which will probably affect `koji` and the `koji.krbV.Krb5Error` exception:
https://fedoraproject.org/wiki/Changes/kerberos-in-python-modernization

skisela: Alright, so I guess raising `RebaseHelperError` with a message that koji session was not created possibly because of missing kerberos ticket, will be better.

nforro: There should be some error message associated with `koji.krbV.Krb5Error`, so I would just use that.

rebasehelper/utils.py
```diff
@@ -867,9 +867,12 @@ def session_maker(cls, baseurl=None):
...
             koji_session.krb_login()
-        except koji.krbV.Krb5Error:
-            # fall back to login using certificate
-            koji_session.ssl_login(cls.cert, cls.ca_cert, cls.ca_cert)
+        except koji.krbV.Krb5Error as e:
+            try:
+                # fall back to login using certificate
+                koji_session.ssl_login(cls.cert, cls.ca_cert, cls.ca_cert)
+            except Exception:
+                raise RebaseHelperError('Kerberos login failed. The reason is: {}' % (six.text_type(e)))
```
nforro: This is not correct either. It discards the error raised from `koji_session.ssl_login()`. Proper way to handle this would be to first try to login using kerberos, print an error if it fails, fall back to login using certificate and print an error and stop execution if that fails either.
But as I wrote in my previous comment, I think logging in using certificate can be dropped, because it doesn't work anymore (actually since [2016-12-12](https://fedoraproject.org/wiki/Infrastructure/Kerberos)).

PR #392, opened 2017-11-29
Do not require = in front of builder options

nforro:


skisela:
I am afraid, that this implementation breaks options, where spaces are part of the option:
```
rebase-helper --buildtool=mock --builder-options="-D 'macro1 val'"
```
outputs
`ERROR: Bad option for '--define' ('macro1).  Use --define 'macro expr'`

as the command is converted to 
```
['mock', '--old-chroot', '--rebuild', '/home/skisela/fedora/rubygem-bundler/rebase-helper-results/old-build/SRPM/rubygem-bundler-1.13.7-3.fc27.src.rpm', '--resultdir', '/tmp/rebase-helper-RyJ18G/RESULTS', '-D', "'macro1", "val'"]
```
internally also by `get_builder_options()`

skisela:
My bad, the above is broken also in master branch.

nforro:
Please create a new issue for that.

PR #401, opened 2017-12-03
Disallow plugins

FrNecas:


rebasehelper/options.py
```diff
@@ -116,6 +117,13 @@
         "default": None,
         "help": "tool to use for determining latest upstream version",
     },
+    {
+        "name": ["--disable-spec-hook"],
```
nforro: I think better name for this would be `--spec-hook-blacklist`. It would make it obvious that it is a list and one can specify multiple items. What do you think?

rebasehelper/options.py
```diff
@@ -116,6 +117,13 @@
...
+        "choices": spec_hooks_runner.get_available_spec_hooks(),
+        "default": [],
+        "type": lambda s: s.split(","),
+        "help": "set which spec hooks can't be run",
```
nforro: This sound a bit better to me: `prevent specified spec hooks from being run`.

rebasehelper/options.py
```diff
@@ -116,6 +117,13 @@
         "default": None,
         "help": "tool to use for determining latest upstream version",
     },
+    {
```
nforro: I would add a comment here, something like `# blacklists`.

rebasehelper/specfile.py
```diff
@@ -1528,10 +1532,11 @@ def run_spec_hooks(self, spec_file, rebase_spec_file, **kwargs):
         :param kwargs: Keyword arguments from Application instance
         """
         for name, spec_hook in six.iteritems(self.spec_hooks):
-            categories = spec_hook.get_categories()
-            if not categories or spec_file.category in categories:
-                logger.info("Running '%s' spec hook", name)
-                spec_hook.run(spec_file, rebase_spec_file, **kwargs)
+            if spec_hook.__name__ not in kwargs["disable_spec_hook"]:
```
nforro: You could reverse the condition and just `continue` if it passes. It would prevent unnecessary indentation.

rebasehelper/versioneer.py
```diff
@@ -70,20 +70,22 @@ def get_available_versioneers(self):
...
+        :param disabled_versioneers: Versioneers disabled on the command line or in a config file
         :return: Latest upstream version of a package
         """
         if versioneer:
             logger.info("Running '%s' versioneer", versioneer)
             return self.versioneers[versioneer].run(package_name)
-        # run all versioneers, categorized first
-        for versioneer in sorted(self.versioneers.values(), key=lambda v: not v.get_categories()):
+        # run all versioneers, except those disabled in config, categorized first
+        for versioneer in sorted([v for k, v in six.iteritems(self.versioneers) if k not in disabled_versioneers],
```
nforro: It would be better to move the filtering step to a separate line.

rebasehelper/versioneer.py
```diff
@@ -70,20 +70,22 @@ def get_available_versioneers(self):
...
 
-    def run(self, versioneer, package_name, category):
+    def run(self, versioneer, package_name, category, disabled_versioneers):
         """
         Runs specified versioneer or all versioneers subsequently until one of them succeeds.
 
         :param versioneer: Name of a versioneer
         :param package_name: Name of a package
         :param category: Package category
+        :param disabled_versioneers: Versioneers disabled on the command line or in a config file
```
nforro: This is not accurate, what if at some point this method will be run from some other location and with a set of disabled versioneers that is acquired from a totally different source (or just always empty for instance)?
You should state what is the meaning of this argument in context of this method, so for example: `List of versioneers that will be skipped`.

rebasehelper/tests/test_versioneer.py
```diff
@@ -42,7 +42,7 @@ class TestVersioneer(object):
     ])
     def test_anitya_versioneer(self, package, min_version):
         assert AnityaVersioneer.get_name() in versioneers_runner.versioneers
-        version = versioneers_runner.run(AnityaVersioneer.get_name(), package, None)
+        version = versioneers_runner.run(AnityaVersioneer.get_name(), package, None, [])
```
nforro: The last argument could default to empty list, that way you wouldn't have to specify it at all.

rebasehelper/tests/test_specfile.py
```diff
@@ -438,14 +438,14 @@ def test_typo_fix_spec_hook(self, spec_object):
         assert TypoFixHook.get_name() in spec_hooks_runner.spec_hooks
         assert '- This is chnagelog entry with some indentional typos\n' in spec_object.spec_content
         spec_object.category = 'sample'
-        spec_hooks_runner.run_spec_hooks(spec_object, spec_object)
+        spec_hooks_runner.run_spec_hooks(spec_object, spec_object, **{'disable_spec_hook': []})
```
nforro: Why don't you simply use `disable_spec_hook=[]`?

Anyway, I think it would make sense not to pass empty list here, but default to empty list in `run_spec_hooks()` in case `disable_spec_hook` kwarg doesn't exist.

rebasehelper/options.py
FrNecas: Yes, blacklist describes it better than disable. I will also use "--versioneer-blacklist" for disabling versionners.

rebasehelper/specfile.py
```diff
@@ -1529,7 +1533,12 @@ def run_spec_hooks(self, spec_file, rebase_spec_file, **kwargs):
         :param rebase_spec_file: Rebased spec file object
         :param kwargs: Keyword arguments from Application instance
         """
+        if not kwargs.get("spec_hook_blacklist"):
+            kwargs["spec_hook_blacklist"] = []
```
nforro: What about `blacklist = kwargs.get("spec_hook_blacklist", [])`?

rebasehelper/versioneer.py
```diff
@@ -70,20 +70,22 @@ def get_available_versioneers(self):
...
+        :param versioneer_blacklist: List of versioneers that will be skipped
         :return: Latest upstream version of a package
         """
         if versioneer:
             logger.info("Running '%s' versioneer", versioneer)
             return self.versioneers[versioneer].run(package_name)
-        # run all versioneers, categorized first
-        for versioneer in sorted(self.versioneers.values(), key=lambda v: not v.get_categories()):
+        # run all versioneers, except those disabled in config, categorized first
+        allowed_versionners = [v for k, v in six.iteritems(self.versioneers) if k not in versioneer_blacklist]
```
nforro: There is a typo in the variable name.

Issue #378, opened 2017-11-03
RFE: Check {rpm,abipkg,pkg}diff output and inform user in case it might contain something important


skisela:
Eg. license consistency from rpmdiff.

skisela:
Resolved by #404

PR #404, opened 2017-12-06
Improve text output, check diffs output

skisela:
- Check output of ABIPkgDiff
- Implement format() method for each checker for better text output
- Add info about untouched patches to results

Resolves: #426, #378

### DIFF
--- a/rebasehelper/checker.py
+++ b/rebasehelper/checker.py
@@ -20,17 +20,22 @@
 # Authors: Petr Hracek <phracek@redhat.com>
 #          Tomas Hozza <thozza@redhat.com>
 
-import six
 import pkg_resources
+import os
+
+import six
 
 from rebasehelper.logger import logger
+from rebasehelper.settings import REBASE_HELPER_RESULTS_DIR
 
 
 class BaseChecker(object):
     """ Base class used for testing tool run on final pkgs. """
 
+    NAME = None
     DEFAULT = False
     category = None
+    results_dir = None
 
     @classmethod
     def match(cls, cmd):
@@ -41,9 +46,14 @@ class BaseChecker(object):
         raise NotImplementedError()
 
     @classmethod
+    def get_checker_output_dir_short(cls):
+        """Return short version of checker output directory"""
+        return os.path.join(REBASE_HELPER_RESULTS_DIR, 'checkers', cls.NAME)
+
+    @classmethod
     def get_checker_name(cls):
         """Returns a name of the checker"""
-        raise NotImplementedError()
+        return cls.NAME
 
     @classmethod
     def is_default(cls):
@@ -58,6 +68,10 @@ class BaseChecker(object):
     @classmethod
     def get_category(cls):
         return cls.category
+
+    @classmethod
+    def get_underlined_title(cls, text, separator='='):
+        return "\n{}\n{}".format(text, separator * len(text))
 
 
 class CheckersRunner(object):
--- a/rebasehelper/checkers/abipkgdiff_tool.py
+++ b/rebasehelper/checkers/abipkgdiff_tool.py
@@ -22,8 +22,10 @@
 
 from __future__ import print_function
 import os
+import re
 
 import rpm
+import six
 
 from rebasehelper.utils import ProcessHelper, RpmHelper
 from rebasehelper.logger import logger
@@ -36,8 +38,9 @@ from rebasehelper.checker import BaseChecker
 class AbiCheckerTool(BaseChecker):
     """abipkgdiff compare tool"""
 
-    CMD = "abipkgdiff"
+    NAME = "abipkgdiff"
     DEFAULT = True
+    abi_changes = None
     results_dir = ''
     log_name = 'abipkgdiff.log'
     category = 'RPM'
@@ -48,14 +51,10 @@ class AbiCheckerTool(BaseChecker):
     # dbus-glib-0.80-3.fc12.x86_64.rpm dbus-glib-0.104-3.fc23.x86_64.rpm
     @classmethod
     def match(cls, cmd=None):
-        if cmd == cls.CMD:
+        if cmd == cls.NAME:
             return True
         else:
             return False
-
-    @classmethod
-    def get_checker_name(cls):
-        return cls.CMD
 
     @classmethod
     def is_default(cls):
@@ -89,9 +88,13 @@ class AbiCheckerTool(BaseChecker):
     @classmethod
     def run_check(cls, results_dir, **kwargs):
         """Compares old and new RPMs using abipkgdiff"""
+        # Check if ABI changes occured
+        cls.abi_changes = None
+        cls.results_dir = os.path.join(results_dir, cls.NAME)
+        os.makedirs(cls.results_dir)
         debug_old, rest_pkgs_old = cls._get_packages_for_abipkgdiff(results_store.get_build('old'))
         debug_new, rest_pkgs_new = cls._get_packages_for_abipkgdiff(results_store.get_build('new'))
-        cmd = [cls.CMD]
+        cmd = [cls.NAME]
         reports = {}
         for pkg in rest_pkgs_old:
             command = list(cmd)
@@ -112,17 +115,99 @@ class AbiCheckerTool(BaseChecker):
             command.append(pkg)
             command.append(new_pkg)
             logger.debug('Package name for ABI comparison %s', old_name)
-            output = os.path.join(cls.results_dir, results_dir, old_name + '-' + cls.log_name)
+            output = os.path.join(cls.results_dir, old_name + '.txt')
             try:
                 ret_code = ProcessHelper.run_subprocess(command, output_file=output)
             except OSError:
-                raise CheckerNotFoundError("Checker '%s' was not found or installed." % cls.CMD)
+                raise CheckerNotFoundError("Checker '{}' was not found or installed.".format(cls.NAME))
 
             if int(ret_code) & settings.ABIDIFF_ERROR and int(ret_code) & settings.ABIDIFF_USAGE_ERROR:
-                raise RebaseHelperError('Execution of %s failed.\nCommand line is: %s' % (cls.CMD, cmd))
-            if int(ret_code) == 0:
-                text = 'ABI of the compared binaries in package %s are equal.' % old_name
-            else:
-                text = 'ABI of the compared binaries in package %s are not equal.' % old_name
-            reports[output] = text
-        return reports
+                raise RebaseHelperError('Execution of {} failed.\nCommand line is: {}'.format(cls.NAME, cmd))
+            reports[old_name] = int(ret_code)
+        return dict(packages=cls.parse_abi_logs(reports),
+                    abi_changes=cls.abi_changes,
+                    path=cls.get_checker_output_dir_short())
+
+    @classmethod
+    def parse_abi_logs(cls, reports):
+        """
+        Parse abipkgdiff logs
+        :param reports: dictionary with paths to the logs of the produced (sub)packages
+        :return: returns dict of packages names with its abipkgdiff changes
+        """
+        def parse_changes(lines):
+            """
+            Parses each line with abipkgdiff output info
+            :param lines: list of lines of changes
+            :return: dictionary of list of changes
+
+            Example abipkgdiff line:
+            '  Functions changes summary: 3 Removed, 0 Changed, 0 Added functions (4 filtered out)'
+            """
+            summary_re = re.compile(r'''^
+            \s+(?P<kind>[\w\s]+changes\s+summary):\s+
+            (?P<changes>.+)
+            $
+            ''', re.VERBOSE)
+            changes_re = re.compile(r'''
+            (?P<count>\d+)\s+
+            (?P<what>Added|Changed|Removed)(\s+functions|variables)?
+            (\s+\((?P<filtered_out>\d+)\s+filtered\s+out\))?
+            ''', re.VERBOSE)
+            result = {}
+            for line in lines:
+                ms = summary_re.match(line)
+                if ms:
+                    ds = ms.groupdict()
+                    result[ds['kind']] = {}
+                    for mc in changes_re.finditer(ds['changes']):
+                        dc = mc.groupdict()
+                        if int(dc['count']) or dc['filtered_out']:
+                            result[ds['kind']][dc['what']] = dc
+            return result
+
+        pkgs = {}
+        for pkg, ret_code in six.iteritems(reports):
+            # If no abi changes for the package, store empty dictionary
+            cur_pkg = {}
+            if ret_code:
+                with open(os.path.join(cls.results_dir, pkg + '.txt'), 'r') as f:
+                    cur_pkg = parse_changes(f.readlines())
+                    cls.abi_changes = True
+            pkgs[pkg] = cur_pkg
+        return pkgs
+
+    @classmethod
+    def format(cls, data):
+        """
+        Format abipkgdiff output
+        :param data: abipkgdiff dictionary output
+        :return: formated abipkgdiff list of strings
+        """
+        output_lines = [cls.get_underlined_title('abipkgdiff')]
+        if not data['abi_changes']:
+            output_lines.append('No ABI changes occured')
+            return output_lines
+
+        for pkg_name, pkg_changes in sorted(six.iteritems(data['packages'])):
+            if not pkg_changes:
+                continue
+            output_lines.append("ABI changes in {}:".format(pkg_name))
+
+            for sum_title, changes_list in sorted(six.iteritems(pkg_changes)):
+                if not changes_list:
+                    continue
+                output_lines.append("{}".format(sum_title))
+
+                for change_name, change_info in sorted(six.iteritems(changes_list)):
+                    if change_info['filtered_out']:
+                        output_lines.append(" - {} {} (filtered out {})".format(change_name,
+                                                                                change_info['count'],
+                                                                                change_info['filtered_out']))
+                    else:
+                        output_lines.append(" - {} {}".format(change_name, change_info['count']))
+        output_lines.append("Details in {}:".format(data['path']))
+        for pkg_name, pkg_changes in sorted(six.iteritems(data['packages'])):
+            output_lines.append(" - {}.txt".format(pkg_name))
+
+        return output_lines
--- a/rebasehelper/checkers/csmock_tool.py
+++ b/rebasehelper/checkers/csmock_tool.py
@@ -22,7 +22,8 @@
 
 from __future__ import print_function
 import os
-from six import StringIO
+
+import six
 
 from rebasehelper.utils import ProcessHelper
 from rebasehelper.utils import PathHelper
@@ -34,19 +35,19 @@ from rebasehelper.checker import BaseChecker
 class CsmockTool(BaseChecker):
     """ Csmock compare tool."""
 
-    CMD = "csmock"
+    NAME = "csmock"
     category = "SRPM"
 
     @classmethod
     def match(cls, cmd=None):
-        if cmd == cls.CMD:
+        if cmd == cls.NAME:
             return True
         else:
             return False
 
     @classmethod
     def get_checker_name(cls):
-        return cls.CMD
+        return cls.NAME
 
     @classmethod
     def is_default(cls):
@@ -59,21 +60,38 @@ class CsmockTool(BaseChecker):
 
         old_pkgs = results_store.get_old_build().get('srpm', None)
         new_pkgs = results_store.get_new_build().get('srpm', None)
-        csmock_dir = os.path.join(results_dir, cls.CMD)
-        os.makedirs(csmock_dir)
+        results_dir = os.path.join(results_dir, cls.NAME)
+        os.makedirs(results_dir)
         arguments = ['--force', '-a', '-r', 'fedora-rawhide-x86_64', '--base-srpm']
         if old_pkgs and new_pkgs:
-            cmd = [cls.CMD]
+            cmd = [cls.NAME]
             cmd.extend(arguments)
             cmd.append(old_pkgs)
             cmd.append(new_pkgs)
-            cmd.extend(['-o', csmock_dir])
-            output = StringIO()
+            cmd.extend(['-o', results_dir])
+            output = six.StringIO()
             try:
                 ProcessHelper.run_subprocess(cmd, output_file=output)
             except OSError:
-                raise CheckerNotFoundError("Checker '%s' was not found or installed." % cls.CMD)
-        csmock_report['error'] = PathHelper.find_all_files_current_dir(csmock_dir, '*.err')
-        csmock_report['txt'] = PathHelper.find_all_files_current_dir(csmock_dir, '*.txt')
-        csmock_report['log'] = PathHelper.find_all_files_current_dir(csmock_dir, '*.log')
+                raise CheckerNotFoundError("Checker '{}' was not found or installed.".format(cls.NAME))
+        csmock_report['error'] = PathHelper.find_all_files_current_dir(results_dir, '*.err')
+        csmock_report['txt'] = PathHelper.find_all_files_current_dir(results_dir, '*.txt')
+        csmock_report['log'] = PathHelper.find_all_files_current_dir(results_dir, '*.log')
+        csmock_report['path'] = cls.get_checker_output_dir_short()
         return csmock_report
+
+    @classmethod
+    def format(cls, data):
+        """
+        Formats csmock data to string
+        :param data: csmock data dictionary
+        :return: string formated output
+        """
+        output_lines = [cls.get_underlined_title("csmock")]
+        output_lines.append("Details in {}:".format(data['path']))
+        for key, files_list in six.iteritems(data):
+            if key in ['error', 'txt', 'log'] and files_list:
+                for f in files_list:
+                    output_lines.append(" - {}".format(os.path.basename(f)))
+
+        return output_lines
--- a/rebasehelper/checkers/pkgdiff_tool.py
+++ b/rebasehelper/checkers/pkgdiff_tool.py
@@ -36,9 +36,9 @@ from xml.etree import ElementTree
 class PkgDiffTool(BaseChecker):
     """ Pkgdiff compare tool. """
 
-    CMD = "pkgdiff"
+    NAME = "pkgdiff"
     DEFAULT = True
-    pkgdiff_results_filename = 'pkgdiff_reports.html'
+    pkgdiff_results_filename = 'report'
     files_xml = "files.xml"
     results_dir = ''
     results_dict = {}
@@ -46,14 +46,10 @@ class PkgDiffTool(BaseChecker):
 
     @classmethod
     def match(cls, cmd=None):
-        if cmd == cls.CMD:
+        if cmd == cls.NAME:
             return True
         else:
             return False
-
-    @classmethod
-    def get_checker_name(cls):
-        return cls.CMD
 
     @classmethod
     def is_default(cls):
@@ -93,7 +89,7 @@ class PkgDiffTool(BaseChecker):
             with open(file_name, 'w') as f:
                 f.writelines(lines)
         except IOError:
-            raise RebaseHelperError("Unable to create XML file for pkgdiff tool '%s'" % file_name)
+            raise RebaseHelperError("Unable to create XML file for pkgdiff tool '{}'".format(file_name))
 
         return file_name
 
@@ -207,10 +203,11 @@ class PkgDiffTool(BaseChecker):
         Compares old and new RPMs using pkgdiff
         :param results_dir result dir where are stored results
         """
-        cls.results_dir = results_dir
-        cls.pkgdiff_results_full_path = os.path.join(cls.results_dir, cls.pkgdiff_results_filename)
+        cls.results_dir = os.path.join(results_dir, cls.NAME)
+        os.makedirs(cls.results_dir)
+        cls.pkgdiff_results_full_path_html = os.path.join(cls.results_dir, cls.pkgdiff_results_filename + '.html')
 
-        cmd = [cls.CMD]
+        cmd = [cls.NAME]
         cmd.append('-hide-unchanged')
         for version in ['old', 'new']:
             old = results_store.get_build(version)
@@ -220,18 +217,18 @@ class PkgDiffTool(BaseChecker):
         cmd.append('-extra-info')
         cmd.append(cls.results_dir)
         cmd.append('-report-path')
-        cmd.append(cls.pkgdiff_results_full_path)
+        cmd.append(cls.pkgdiff_results_full_path_html)
         try:
             ret_code = ProcessHelper.run_subprocess(cmd, output_file=ProcessHelper.DEV_NULL)
         except OSError:
-            raise CheckerNotFoundError("Checker '%s' was not found or installed." % cls.CMD)
+            raise CheckerNotFoundError("Checker '{}' was not found or installed.".format(cls.NAME))
 
         # From pkgdiff source code:
         # ret_code 0 means unchanged
         # ret_code 1 means Changed
         # other return codes means error
         if int(ret_code) != 0 and int(ret_code) != 1:
-            raise RebaseHelperError('Execution of %s failed.\nCommand line is: %s' % (cls.CMD, cmd))
+            raise RebaseHelperError('Execution of {} failed.\nCommand line is: {}'.format(cls.NAME, cmd))
         results_dict = cls.process_xml_results(cls.results_dir)
         lines = []
 
@@ -239,14 +236,27 @@ class PkgDiffTool(BaseChecker):
             if val:
                 if lines:
                     lines.append('')
-                lines.append('Following files were %s:' % key)
+                lines.append('Following files were {}:'.format(key))
                 lines.extend(val)
 
-        pkgdiff_report = os.path.join(cls.results_dir, 'report-' + cls.pkgdiff_results_filename + '.log')
+        pkgdiff_report = os.path.join(cls.results_dir, cls.pkgdiff_results_filename + '.txt')
         try:
             with open(pkgdiff_report, "w") as f:
                 f.write('\n'.join(lines))
         except IOError:
-            raise RebaseHelperError("Unable to write result from %s to '%s'" % (cls.CMD, pkgdiff_report))
+            raise RebaseHelperError("Unable to write result from {} to '{}'".format(cls.NAME, pkgdiff_report))
 
-        return {pkgdiff_report: None}
+        return dict(path=cls.get_checker_output_dir_short())
+
+    @classmethod
+    def format(cls, data):
+        """
+        Formats pkgdiff data to string
+        :param data: pkgdiff data dictionary
+        :return: formated pkgdiff list of strings
+        """
+        output_lines = [cls.get_underlined_title("pkgdiff")]
+        output_lines.append("Details in {}:".format(data['path']))
+        output_lines.append(" - {}.html".format(cls.pkgdiff_results_filename))
+        output_lines.append(" - {}.txt".format(cls.pkgdiff_results_filename))
+        return output_lines
--- a/rebasehelper/checkers/rpmdiff_tool.py
+++ b/rebasehelper/checkers/rpmdiff_tool.py
@@ -37,20 +37,16 @@ from rebasehelper.checker import BaseChecker
 class RpmDiffTool(BaseChecker):
     """rpmdiff compare tool"""
 
-    CMD = "rpmdiff"
+    NAME = "rpmdiff"
     DEFAULT = True
     category = "RPM"
 
     @classmethod
     def match(cls, cmd=None):
-        if cmd == cls.CMD:
+        if cmd == cls.NAME:
             return True
         else:
             return False
-
-    @classmethod
-    def get_checker_name(cls):
-        return cls.CMD
 
     @classmethod
     def is_default(cls):
@@ -112,8 +108,9 @@ class RpmDiffTool(BaseChecker):
 
         for tag in settings.CHECKER_TAGS:
             results_dict[tag] = []
-        cls.results_dir = results_dir
-        cls.workspace_dir = kwargs['workspace_dir']
+
+        cls.results_dir = os.path.join(results_dir, cls.NAME)
+        os.makedirs(cls.results_dir)
 
         # Only S (size), M(mode) and 5 (checksum) are now important
         not_catched_flags = ['T', 'F', 'G', 'U', 'V', 'L', 'D', 'N']
@@ -123,7 +120,7 @@ class RpmDiffTool(BaseChecker):
             if 'debuginfo' in key or 'debugsource' in key:
                 # skip debug{info,source} packages
                 continue
-            cmd = [cls.CMD]
+            cmd = [cls.NAME]
             # TODO modify to online command
             for x in not_catched_flags:
                 cmd.extend(['-i', x])
@@ -138,9 +135,8 @@ class RpmDiffTool(BaseChecker):
             try:
                 ProcessHelper.run_subprocess(cmd, output_file=output)
             except OSError:
-                raise CheckerNotFoundError("Checker '%s' was not found or installed." % cls.CMD)
+                raise CheckerNotFoundError("Checker '{}' was not found or installed.".format(cls.NAME))
             results_dict = cls._analyze_logs(output, results_dict)
-
         results_dict = cls.update_added_removed(results_dict)
         cls.results_dict = {k: v for k, v in six.iteritems(results_dict) if v}
         lines = []
@@ -148,14 +144,35 @@ class RpmDiffTool(BaseChecker):
             if val:
                 if lines:
                     lines.append('')
-                lines.append('Following files were %s:' % key)
+                lines.append('Following files were {}:'.format(key))
                 lines.extend(val)
 
-        rpmdiff_report = os.path.join(cls.results_dir, 'report-' + cls.CMD + '.log')
+        rpmdiff_report = os.path.join(cls.results_dir, 'report.txt')
+
+        counts = {k: len(v) for k, v in six.iteritems(results_dict)}
+
         try:
             with open(rpmdiff_report, "w") as f:
                 f.write('\n'.join(lines))
         except IOError:
-            raise RebaseHelperError("Unable to write result from %s to '%s'" % (cls.CMD, rpmdiff_report))
+            raise RebaseHelperError("Unable to write result from {} to '{}'".format(cls.NAME, rpmdiff_report))
 
-        return {rpmdiff_report: None}
+        return {'path': cls.get_checker_output_dir_short(), 'files_changes': counts}
+
+    @classmethod
+    def format(cls, data):
+        """
+        Format rpmdiff output for outputtool
+        :param data: rpmdiff output data dictionary
+        :return: formated rpmdiff output list of strings
+        """
+        output_lines = [cls.get_underlined_title("rpmdiff")]
+
+        output_lines.append(" - {} added files".format(data['files_changes']['added']))
+        output_lines.append(" - {} changed files".format(data['files_changes']['changed']))
+        output_lines.append(" - {} removed files".format(data['files_changes']['removed']))
+
+        output_lines.append("Details in {}:".format(data['path']))
+        output_lines.append(" - report.txt")
+
+        return output_lines
--- a/rebasehelper/output_tool.py
+++ b/rebasehelper/output_tool.py
@@ -26,6 +26,8 @@ import pkg_resources
 
 from rebasehelper.logger import logger, logger_output
 from rebasehelper.results_store import results_store
+from rebasehelper.checkers.abipkgdiff_tool import AbiCheckerTool
+from rebasehelper.settings import REBASE_HELPER_RESULTS_DIR
 
 
 class BaseOutputTool(object):
@@ -47,6 +49,10 @@ class BaseOutputTool(object):
         raise NotImplementedError
 
     @classmethod
+    def prepend_results_dir_name(cls, *path_members):
+        return os.path.join(REBASE_HELPER_RESULTS_DIR, *path_members)
+
+    @classmethod
     def print_cli_summary(cls, app):
         """
         Print report of the rebase
@@ -55,24 +61,31 @@ class BaseOutputTool(object):
         """
         cls.app = app
         cls.print_patches_cli()
+        result = results_store.get_result_message()
+        checkers = results_store.get_checkers()
+
+        if 'abipkgdiff' in checkers:
+            if AbiCheckerTool.abi_changes:
+                logger_output.warning('\nABI changes occured. Check abipkgdiff output')
 
         logger_output.heading('\nAvailable logs:')
-        logger_output.info('%s:\n%s', 'Debug log', app.debug_log_file)
+        logger_output.info('%s:\n%s', 'Debug log', cls.prepend_results_dir_name(os.path.relpath(app.debug_log_file,
+                                                                                app.results_dir)))
         if results_store.get_old_build() is not None:
-            logger_output.info('%s:\n%s', 'Old build logs and (S)RPMs', os.path.join(app.results_dir, 'old-build'))
+            logger_output.info('%s:\n%s', 'Old build logs and (S)RPMs', cls.prepend_results_dir_name('old-build'))
         if results_store.get_new_build() is not None:
-            logger_output.info('%s:\n%s', 'New build logs and (S)RPMs', os.path.join(app.results_dir, 'new-build'))
+            logger_output.info('%s:\n%s', 'New build logs and (S)RPMs', cls.prepend_results_dir_name('old-build'))
         logger_output.info('')
 
         logger_output.heading('%s:', 'Rebased sources')
-        logger_output.info("%s", app.rebased_sources_dir)
+        logger_output.info("%s", cls.prepend_results_dir_name(os.path.relpath(app.debug_log_file, app.results_dir)))
 
-        logger_output.heading('%s:', 'Generated patch')
-        logger_output.info("%s\n", os.path.join(app.results_dir, 'changes.patch'))
+        patch = results_store.get_changes_patch()
+        if 'changes_patch' in patch:
+            logger_output.heading('%s:', 'Generated patch')
+            logger_output.info("%s\n", cls.prepend_results_dir_name(os.path.basename(patch['changes_patch'])))
 
         cls.print_report_file_path()
-
-        result = results_store.get_result_message()
 
         if not app.conf.patch_only:
             if 'success' in result:
@@ -88,7 +101,7 @@ class BaseOutputTool(object):
     def print_report_file_path(cls):
         """Print path to the report file"""
         logger_output.heading('%s report:' % cls.NAME)
-        logger_output.info('%s', os.path.join(cls.app.results_dir, 'report.' + cls.get_extension()))
+        logger_output.info('%s', cls.prepend_results_dir_name('report.' + cls.get_extension()))
 
     @classmethod
     def print_patches_cli(cls):
@@ -116,7 +129,7 @@ class BaseOutputTool(object):
 
         if patch_type in patches:
             logger_output.info('\n%s patches:', patch_type)
-            for patch in patches[patch_type]:
+            for patch in sorted(patches[patch_type]):
                 logger_method(patch)
 
     @classmethod
--- a/rebasehelper/output_tools/text_output_tool.py
+++ b/rebasehelper/output_tools/text_output_tool.py
@@ -132,7 +132,7 @@ class TextOutputTool(BaseOutputTool):
         """Function is used for printing rpm build logs"""
         if rpms.get('logs', None) is None:
             return
-        for logs in sorted(rpms.get('logs', None)):
+        for logs in sorted(rpms.get('logs', []) + rpms.get('srpm_logs', [])):
             if dirpath not in logs:
                 # Skip logs that do not belong to curent rpms(and version)
                 continue
--- a/rebasehelper/output_tools/text_output_tool.py
+++ b/rebasehelper/output_tools/text_output_tool.py
@@ -5,6 +5,7 @@ from rebasehelper.output_tool import BaseOutputTool
 from rebasehelper.exceptions import RebaseHelperError
 from rebasehelper.logger import LoggerHelper, logger, logger_report
 from rebasehelper.results_store import results_store
+from rebasehelper.checker import checkers_runner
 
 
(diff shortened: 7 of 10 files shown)


skisela:
This needs to be modified, as it does not count with adding/deleting files between versions.

rebasehelper/checkers/abipkgdiff_tool.py
```diff
@@ -40,6 +40,7 @@ class AbiCheckerTool(BaseChecker):
     DEFAULT = True
     results_dir = ''
     log_name = 'abipkgdiff.log'
+    abi_changes = False
```
nforro: Just a heads up, this won't work. This attribute is only set to `False` when the module is imported for the first time, so if you execute `checkers_runner.run_checker()` multiple times and there are abi changes detected in one of the runs, they will be reported also for all consecutive runs.

skisela: Well is not it enough, that ABI change appears in either one of all the (sub)packages, to warn the user to pay attention to it?

nforro: Sure, but I'm talking about something different. It's wrong to assume there is only one rebase per process. Just try to execute `Application.run()` more than once in one script (with different packages of course) and see for yourself. With your approach we could end up having reports for packages containing information from totally different and unrelated packages (yes, it already happened).

skisela: Oh sure I remember, then you're right, that needs fix as well.

skisela:
Current output of implemented `rpmdiff` and `abipkgdiff`:
Rebase from fcitx-qt5-1.1.1 to fcitx-qt5-1.2.1 completed without an error
```
Rebase from fcitx-qt5-1.1.1 to fcitx-qt5-1.2.1 completed without an error

rpmdiff
======
No license changes detected
* 1 added files
* 0 removed files
* 12 changed files

Details in /home/skisela/fedora/fcitx-qt5/rebase-helper-results/checkers/rpmdiff:
 - rpmdiff.txt
 - licensecheck.txt

abipkgdiff
======
ABI changes in fcitx-qt5:
 * 3 function(s) removed
 * 10 variable symbol(s) removed
 * 2 function symbol(s) added
Details in /home/skisela/fedora/fcitx-qt5/rebase-helper-results/checkers/abipkgdiff
 - fcitx-qt5-devel.txt
 - fcitx-qt5.txt
```

rebasehelper/checkers/rpmdiff_tool.py
```diff
@@ -24,9 +24,11 @@
 import os
 import six
 import re
+import fileinput
```
nforro: This import is unused.

rebasehelper/checkers/rpmdiff_tool.py
```diff
@@ -55,6 +57,51 @@ def get_checker_name(cls):
...
+        Finds out if there is a possibility of license change in new sources
+
+        :return: true if license changes probably happened
+        """
+        diffs = []
+        for srcs in ['new_sources', 'old_sources']:
+            srcs = os.path.join(cls.workspace_dir, srcs)
+
+            # Get all files
+            files = PathHelper.find_all_files(srcs, '*')
```
nforro: What is this for? The list of changed files is in `cls.results_dict['changed']`, right?

rebasehelper/checkers/rpmdiff_tool.py
```diff
@@ -55,6 +57,51 @@ def get_checker_name(cls):
...
+                    continue
+
+                # Do not analyse .git directory files
+                if r'.git' in f:
+                    continue
+
+                out = subprocess.Popen(["/bin/licensecheck", f, "--machine"],
+                                       stdout=subprocess.PIPE)
+                # Get licensecheck output
+                out = out.stdout.read()
```
nforro: Why don't you use `ProcerssHelper` for this?

nforro: Also, any reason not to run `licensecheck` for all files at once?

nforro: Also why `/bin`? `licensecheck` package owns `/usr/bin/licensecheck`, and in Fedora `/bin` is just a symlink to `/usr/bin` anyway.

rebasehelper/checkers/rpmdiff_tool.py
```diff
@@ -142,7 +192,7 @@ def run_check(cls, results_dir):
             results_dict = cls._analyze_logs(output, results_dict)
 
         results_dict = cls.update_added_removed(results_dict)
-        results_dict = dict((k, v) for k, v in six.iteritems(results_dict) if v)
+        cls.results_dict = dict((k, v) for k, v in six.iteritems(results_dict) if v)
```
nforro: This could be replaced with [dict comprehension](https://www.python.org/dev/peps/pep-0274/#semantics).

rebasehelper/checkers/rpmdiff_tool.py
skisela: I think it is a bit safer to run it one by one file, in case the list of files becomes too big.

nforro: It is also slower. But you are probably right that in case of really long list of arguments it could be problematic.

rebasehelper/checkers/abipkgdiff_tool.py
```diff
@@ -22,8 +22,9 @@
 
 from __future__ import print_function
 import os
-
 import rpm
+import six
+import re
```
nforro: `re` is part of standard library, so it should be grouped with `import os`, `six` is third party library, so it should stay grouped with `import rpm`.

rebasehelper/checkers/abipkgdiff_tool.py
```diff
@@ -113,19 +116,90 @@ def run_check(cls, result_dir, **kwargs):
...
+                with open(os.path.join(cls.results_dir, pkg + '.txt'), 'r') as f:
+                    lines = f.readlines()
+                    cur_pkg['functions_changes'] = cls.parse_line("Functions changes .* (\d+) .* (\d+) .* (\d+)",
+                                                                 lines[1])
+                    cur_pkg['variables_changes'] = cls.parse_line("Variables changes .* (\d+) .* (\d+) .* (\d+)",
+                                                                 lines[2])
+                    cur_pkg['functions_symbols_changes'] = cls.parse_line("Function symbols .* (\d+) .* (\d+)",
+                                                                         lines[3])
+                    cur_pkg['variables_symbols_changes'] = cls.parse_line("Variable symbols .* (\d+) .* (\d+)",
+                                                                          lines[4])
```
nforro: The regexps should be definitely more specific. This doesn't have any better chance to match correctly if something changes IMHO. On the other hand, I don't see any reason for matching specific patterns on specific lines. If the order changes it could still work.

Also, you are using single backslashes in non-raw string literals, which is wrong.

rebasehelper/checkers/abipkgdiff_tool.py
```diff
@@ -113,19 +116,90 @@ def run_check(cls, result_dir, **kwargs):
...
+            elif 'variable' in line:
+                return ['%s variable symbol(s) removed' % found.group(1),
+                        '%s variable symbol(s) added' % found.group(2)]
+        elif len(found.groups()) == 3:
+            if 'function' in line:
+                return ['%s function(s) removed' % found.group(1), '%s function(s) changed' % found.group(2),
+                        '%s function(s) added' % found.group(3)]
+            elif 'variable' in line:
+                return ['%s variable(s) removed' % found.group(1), '%s variable(s) changed' % found.group(2),
+                        '%s variable(s) added' % found.group(3)]
```
nforro: I don't like this method. You already know which of the regexps matched, so why test count of groups and presence of some string?
Also, what if regexp doesn't match at all?

rebasehelper/checkers/abipkgdiff_tool.py
skisela:
```
Functions changes summary: 0 Removed, 1 Changed (2 filtered out), 0 Added function
```
We could look for the `changes summary` string and display the whole line, if any changes were found.

rebasehelper/checkers/rpmdiff_tool.py
skisela: You are right, but `cls.results_dict['changed']` contains system paths and not `rebase-helper-workspace` paths.

nforro: Right, because rpmdiff compares binary rpms. That means licensecheck should be run against contents of those binary rpms as well. You can't assume that a source file named X will have the same name (and more importantly, same content) after build.

rebasehelper/checkers/abipkgdiff_tool.py
```diff
@@ -22,8 +22,9 @@
 
 from __future__ import print_function
 import os
-
+import re
```
nforro: You are missing empty line here to separate standard library modules from third party ones.

rebasehelper/checkers/abipkgdiff_tool.py
```diff
@@ -112,17 +118,78 @@ def run_check(cls, results_dir, **kwargs):
...
             if int(ret_code) & settings.ABIDIFF_ERROR and int(ret_code) & settings.ABIDIFF_USAGE_ERROR:
                 raise RebaseHelperError('Execution of %s failed.\nCommand line is: %s' % (cls.CMD, cmd))
-            if int(ret_code) == 0:
-                text = 'ABI of the compared binaries in package %s are equal.' % old_name
-            else:
-                text = 'ABI of the compared binaries in package %s are not equal.' % old_name
-            reports[output] = text
-        return reports
+            ret_code = int(ret_code)
+            reports[old_name] = ret_code
```
nforro: Why don't you merge this into single line?

rebasehelper/checkers/abipkgdiff_tool.py
```diff
@@ -112,17 +118,78 @@ def run_check(cls, results_dir, **kwargs):
...
+        def parse_changes(chgs):
+            """
+            Parses each line with abipkgdiff output info
+            :param chgs: lines of changes
+            :return: dictionary of list of changes
+            """
+            ret_changes = {}
+            for line in chgs:
+                # Find all numbers of added/changed/deleted and get rid of '0' changes
+                match = re.findall(r'\d+ \w+', line)
```
nforro: This is too generic, you should search for specific patterns (*Added*, *Removed*, *Changed*). Take this for example:
```
Functions changes summary: 0 Removed, 2 Changed (144 filtered out), 2 Added functions
Variables changes summary: 0 Removed, 0 Changed (1 filtered out), 0 Added variable
```

You have to consider also filtered-out symbols.

rebasehelper/checkers/abipkgdiff_tool.py
```diff
@@ -112,17 +118,78 @@ def run_check(cls, results_dir, **kwargs):
...
+                counts = [x for x in match if not x.startswith(r'0')]
+                if counts:
+                    # If any useful info found, store it
+                    change_name = re.search(r'(\w+\ ?)+:', line).group(0)
```
nforro: What's with the extra backslash in the regexp? Also, I don't get why are you using such a weird way to extract the substring, when you can use something like `re.match(r'^([\w\s]+):', line).group(1)`. Of course both `re.search` and `re.match` can return `None`, so you should count with that.
Anyway, all of this (getting both titles and counts) could be done with a single regexp (perhaps a verbose one, to make it more readable).

rebasehelper/checkers/abipkgdiff_tool.py
```diff
@@ -112,17 +118,78 @@ def run_check(cls, results_dir, **kwargs):
...
+            return ret_changes
+
+        pkgs = {}
+        for pkg, ret_code in six.iteritems(reports):
+            cur_pkg = {}
+            if ret_code:
+                with open(os.path.join(cls.results_dir, pkg + '.txt'), 'r') as f:
+                    lines = f.readlines()
+                    # Get lines that contain the needed info
+                    changes = [line for line in lines if re.search(r'changes summary', line)]
```
nforro: Using `re.search` here is totally unnecessary (simple `if 'changes summary' in line` would suffice), and as I wrote above, this all could be replaced with a single regexp.

skisela: Fixed, but if you meant merging this to https://github.com/rebase-helper/rebase-helper/pull/404/files#diff-e58646825031faad76b12d43acd60896R149
then I disagree, as it is not worth having such a long line IMHO.

skisela:
Adding libtiff text rebase helper output

```
Rebase from libtiff-4.0.7 to libtiff-4.0.9 completed without an error


rpmdiff
======
* 4 added files
* 6 removed files
* 32 changed files
Details in /home/skisela/fedora/libtiff/rebase-helper-results/checkers/rpmdiff:
 - rpmdiff.txt

abipkgdiff
======
ABI changes in libtiff:
  Functions changes summary:
 * 0 Removed
 * 0 Changed
 * 9 Added
  Variables changes summary:
 * 0 Removed
 * 0 Changed
 * 0 Added
Details in /home/skisela/fedora/libtiff/rebase-helper-results/checkers/abipkgdiff
 - libtiff-devel.txt
 - libtiff.txt
 - libtiff-tools.txt
 - libtiff-static.txt

pkgdiff
======
Details in /home/skisela/fedora/libtiff/rebase-helper-results/checkers/pkgdiff

Patch with differences between old and new version source files:
/home/skisela/fedora/libtiff/rebase-helper-results/changes.patch

Downstream Patches
===================
Legend:
[-] = already applied
[*] = merged, patch modified
[!] = conflicting or inapplicable, patch skipped
[ ] = patch untouched
 * libtiff-hylafax-fix.patch                [-]
 * libtiff-CVE-2016-10266.patch             [-]
 * libtiff-CVE-2016-10267.patch             [-]
 * libtiff-CVE-2016-10268.patch             [-]
 * libtiff-CVE-2016-10269.patch             [-]
 * libtiff-CVE-2016-10270.patch             [-]
 * libtiff-CVE-2016-10271_10272.patch       [-]
 * libtiff-CVE-2017-7592.patch              [-]
 * libtiff-CVE-2017-7594.patch              [-]
 * libtiff-CVE-2017-7595.patch              [-]
 * libtiff-CVE-2017-7596_7597_7599_7600.patch [-]
 * libtiff-CVE-2017-7598.patch              [-]
 * libtiff-CVE-2017-7601.patch              [-]
 * libtiff-CVE-2017-7602.patch              [-]

RPMS
=====

Old packages:
--------------

Source packages and logs are in directory /home/skisela/fedora/libtiff/rebase-helper-results/old-build/SRPM :
 - libtiff-4.0.7-5.fc27.src.rpm
 - build.log

Binary packages and logs are in directory /home/skisela/fedora/libtiff/rebase-helper-results/old-build/RPM :
 - libtiff-debuginfo-4.0.7-5.fc25.x86_64.rpm
 - libtiff-tools-4.0.7-5.fc25.x86_64.rpm
 - libtiff-static-4.0.7-5.fc25.x86_64.rpm
 - libtiff-4.0.7-5.fc25.x86_64.rpm
 - libtiff-devel-4.0.7-5.fc25.x86_64.rpm
 - hw_info.log
 - state.log
 - build.log
 - installed_pkgs.log
 - root.log
 - mock_output.log

New packages:
--------------

Source packages and logs are in directory /home/skisela/fedora/libtiff/rebase-helper-results/new-build/SRPM :
 - libtiff-4.0.9-1.fc27.src.rpm
 - build.log

Binary packages and logs are in directory /home/skisela/fedora/libtiff/rebase-helper-results/new-build/RPM :
 - libtiff-devel-4.0.9-1.fc25.x86_64.rpm
 - libtiff-tools-4.0.9-1.fc25.x86_64.rpm
 - libtiff-static-4.0.9-1.fc25.x86_64.rpm
 - libtiff-debuginfo-4.0.9-1.fc25.x86_64.rpm
 - libtiff-4.0.9-1.fc25.x86_64.rpm
 - hw_info.log
 - state.log
 - build.log
 - installed_pkgs.log
 - root.log
 - mock_output.log
```

rebasehelper/checkers/abipkgdiff_tool.py
```diff
@@ -112,17 +119,80 @@ def run_check(cls, results_dir, **kwargs):
...
+            :return: dictionary of list of changes
+            """
+            ret_changes = {}
+            for line in chgs:
+                # Find all numbers of added/changed/deleted and get rid of '0' changes
+                m = re.search(r'(^[\w\s]+:)'
+                              r'.{1}([\d]+ Removed[\s\w\d()]*)'
+                              r'.{2}([\d]+ Changed[\s\w\d()]*)'
+                              r'.{2}([\d]+ Added[\s\w\d()]*)',
+                              line)
```
nforro: The regexp is still far too generic, but for no good reason - if something changes substantially in the output of abipkgdiff, it will still fail to parse it.

nforro: And you are ignoring filtered-out changes, which is IMHO quite important information.

rebasehelper/checkers/abipkgdiff_tool.py
```diff
@@ -112,17 +119,80 @@ def run_check(cls, results_dir, **kwargs):
...
+                # If any useful info found, store it
+                ret_changes[m.group(1)] = [change for change in m.groups()[1:]]
+            return ret_changes
+
+        pkgs = {}
+        for pkg, ret_code in six.iteritems(reports):
+            cur_pkg = {}
+            if ret_code:
+                with open(os.path.join(cls.results_dir, pkg + '.txt'), 'r') as f:
+                    lines = f.readlines()
```
nforro: How is this better or more readable than for example this?

```python
def parse(lines):
    summary_re = re.compile(r'''^
        (?P<kind>\w+\s+changes\s+summary):\s+
        (?P<removed>\d+)\s+Removed(\s+\((?P<removed_filtered_out>\d+)\s+filtered\s+out\))?,\s+
        (?P<changed>\d+)\s+Changed(\s+\((?P<changed_filtered_out>\d+)\s+filtered\s+out\))?,\s+
        (?P<added>\d+)\s+Added(\s+\((?P<added_filtered_out>\d+)\s+filtered\s+out\))?\s+
        .*$
        ''', re.VERBOSE)
    result = {}
    for line in lines:
        m = summary_re.match(line)
        if m:
            d = m.groupdict()
            result[d['kind']] = d
    return result

...

    changes = parse(f.readlines())
```

Of course, this expects that order of `Removed`, `Changed`, `Added` is fixed (just like your implementation). More robust solution would be this:

```python
def parse(lines):
    summary_re = re.compile(r'''^
        (?P<kind>\w+\s+changes\s+summary):\s+
        (?P<changes>.+)
        $
        ''', re.VERBOSE)
    change_re = re.compile(r'''
        (?P<count>\d+)\s+
        (?P<what>Added|Changed|Removed)
        (\s+\((?P<filtered_out>\d+)\s+filtered\s+out\))?
        ''', re.VERBOSE)
    result = {}
    for line in lines:
        ms = summary_re.match(line)
        if ms:
            ds = ms.groupdict()
            result[ds['kind']] = {}
            for mc in changes_re.finditer(ds['changes']):
                dc = mc.groupdict()
                result[ds['kind']][dc['what']] = dc
    return result

...

    changes = parse(f.readlines())
```

rebasehelper/checkers/abipkgdiff_tool.py
nforro: And it would be nice to add a comment with what such abipkgdiff output looks like, so we have some reference if it changes in the future.

skisela: You are right, except the second comment, as the regex would match anything except `,` leading to storing the "filtered" counts to the same category(Removed/Changed/Added)

rebasehelper/checkers/abipkgdiff_tool.py
skisela: Thanks, this is way more readable and maintainable!

rebasehelper/checkers/abipkgdiff_tool.py
```diff
@@ -112,17 +119,97 @@ def run_check(cls, results_dir, **kwargs):
...
+            '  Functions changes summary: 3 Removed, 0 Changed, 0 Added functions (4 filtered out)'
+            """
+            summary_re = re.compile(r'''^
+            \s+(?P<kind>[\w\s]+changes\s+summary):\s+
+            (?P<changes>.+)
+            $
+            ''', re.VERBOSE)
+            changes_re = re.compile(r'''
+            (?P<count>\d+)\s+
+            (?P<what>Added|Changed|Removed)(\s+functions)?
```
nforro: I think the last group should look like this: `(\s+functions|variables)?`. Otherwise the regexp won't match `Variables changes summary: ...`.

rebasehelper/checkers/abipkgdiff_tool.py
```diff
@@ -112,17 +119,97 @@ def run_check(cls, results_dir, **kwargs):
...
+            return result
+
+        pkgs = {}
+        for pkg, ret_code in six.iteritems(reports):
+            cur_pkg = {}
+            if ret_code:
+                with open(os.path.join(cls.results_dir, pkg + '.txt'), 'r') as f:
+                    cur_pkg = parse_changes(f.readlines())
+                    cls.abi_changes = True
+            # If no abi changes for the package, store empty dictionary
```
nforro: Just a nitpick, but this comment would probably make more sense if it was moved 5 lines up.

nforro:
You should change base to `master` and remove duplicate commits.

rebasehelper/checkers/abipkgdiff_tool.py
```diff
@@ -112,17 +119,97 @@ def run_check(cls, results_dir, **kwargs):
...
+                    for mc in changes_re.finditer(ds['changes']):
+                        dc = mc.groupdict()
+                        if int(dc['count']) or dc['filtered_out']:
+                            result[ds['kind']][dc['what']] = {'count': dc['count'], 'filtered_out': dc['filtered_out']}
+            return result
+
+        pkgs = {}
+        for pkg, ret_code in six.iteritems(reports):
+            cur_pkg = {}
+            # If no abi changes for the package, store empty dictionary
```
nforro: One more line up.

rebasehelper/patch_helper.py
```diff
@@ -239,10 +239,34 @@ def _git_rebase(cls):
             patch_dictionary['modified'] = modified_patches
         if inapplicable_patches:
             patch_dictionary['inapplicable'] = inapplicable_patches
+        untouched_patches = cls.get_untouched_patches(deleted_patches, modified_patches, inapplicable_patches)
```
nforro: Why extra method for this? It could be rewritten for example this way:

```python
patches = [os.path.basename(p.get_path()) for p in cls.patches]
untouched_patches = [p for p in patches if p not in deleted_patches + modified_patches + inapplicable_patches]
```

nforro:
Some notes to the resulting report:


```
Rebase from libtiff-4.0.7 to libtiff-4.0.9 completed without an error

```
^^ Extra empty line here.
```

rpmdiff
======
```
^^ The number of `=` should be equal to the length of the title.
```
* 4 added files
* 6 removed files
* 34 changed files
Details in /home/nforro/tmp/libtiff/rebase-helper-results/checkers/rpmdiff:
```
^^ The paths really need to be relative, this is simply too long.
```
 - rpmdiff.txt

abipkgdiff
======
ABI changes in libtiff:
Functions changes summary
 * Added 8
Details in /home/nforro/tmp/libtiff/rebase-helper-results/checkers/abipkgdiff
 - libtiff-devel.txt
 - libtiff.txt
 - libtiff-tools.txt
 - libtiff-static.txt
```
^^ This is a bit inconsistent with the `rpmdiff` section.
```

pkgdiff
======
Details in /home/nforro/tmp/libtiff/rebase-helper-results/checkers/pkgdiff

Patch with differences between old and new version source files:
/home/nforro/tmp/libtiff/rebase-helper-results/changes.patch

Downstream Patches
===================
Legend:
[-] = already applied
```
^^ Why is `, patch removed` missing?
```
[*] = merged, patch modified
[!] = conflicting or inapplicable, patch skipped
[ ] = patch untouched
 * libtiff-hylafax-fix.patch                [-]
 * libtiff-CVE-2016-10266.patch             [-]
 * libtiff-CVE-2016-10267.patch             [-]
 * libtiff-CVE-2016-10268.patch             [-]
 * libtiff-CVE-2016-10269.patch             [-]
 * libtiff-CVE-2016-10270.patch             [-]
 * libtiff-CVE-2016-10271_10272.patch       [-]
 * libtiff-CVE-2017-7592.patch              [-]
 * libtiff-CVE-2017-7593.patch              [-]
 * libtiff-CVE-2017-7594.patch              [-]
 * libtiff-CVE-2017-7595.patch              [-]
 * libtiff-CVE-2017-7598.patch              [-]
 * libtiff-CVE-2017-7601.patch              [-]
 * libtiff-am-version.patch                 [ ]
 * libtiff-make-check.patch                 [ ]
 * libtiff-CVE-2017-7602.patch              [ ]
```
^^ This patch shouldn't be listed as untouched, it is modified.
```
 * libtiff-CVE-2017-7602.patch              [*]
```
^^ The order of patches should be made fixed and consistent across all outputs.
```

RPMS
=====

Old packages:
--------------

Source packages and logs are in directory /home/nforro/tmp/libtiff/rebase-helper-results/old-build/SRPM :
 - libtiff-4.0.7-5.fc26.src.rpm

Binary packages and logs are in directory /home/nforro/tmp/libtiff/rebase-helper-results/old-build/RPM :
 - libtiff-debuginfo-4.0.7-5.fc26.x86_64.rpm
 - libtiff-tools-4.0.7-5.fc26.x86_64.rpm
 - libtiff-static-4.0.7-5.fc26.x86_64.rpm
 - libtiff-devel-4.0.7-5.fc26.x86_64.rpm
 - libtiff-4.0.7-5.fc26.x86_64.rpm
 - build.log

New packages:
--------------

Source packages and logs are in directory /home/nforro/tmp/libtiff/rebase-helper-results/new-build/SRPM :
 - libtiff-4.0.9-1.fc26.src.rpm

Binary packages and logs are in directory /home/nforro/tmp/libtiff/rebase-helper-results/new-build/RPM :
 - libtiff-debuginfo-4.0.9-1.fc26.x86_64.rpm
 - libtiff-tools-4.0.9-1.fc26.x86_64.rpm
 - libtiff-static-4.0.9-1.fc26.x86_64.rpm
 - libtiff-devel-4.0.9-1.fc26.x86_64.rpm
 - libtiff-4.0.9-1.fc26.x86_64.rpm
 - build.log
```

nforro:
Again, some notes on the report:

First, I assume the report is markdown - so it looks nice in plain text view but can be also rendered as HTML or whatever.

The headings should be underlined with `=` or `-` characters, the underlines should not overlap. Also, the headings should not end with `:`.

Each list should be sorted alphabetically if possible.

All the paths should be relative, not just some of them.

Bullet lists should be consistent - there is a difference between rpmdiff section and the rest of the report. I think it's fine to use `-` for some lists and `*` for other, but there should be some rule where to use which one.

nforro:
One more thing, pkgdiff lists no files, there is just:

```
Details in rebase-helper-results/checkers/pkgdiff:
```

rebasehelper/patch_helper.py
```diff
@@ -179,7 +179,7 @@ def _git_rebase(cls):
                     with open(base_name, 'wb') as f:
                         f.write(diff)
                         f.write(b'\n')
-                    modified_patches.append(base_name)
+                    modified_patches.append(patch_name)
```
nforro: This is not good. Sure, it fixes the issue, but it doesn't belong to this commit and it's not explained anywhere.

rebasehelper/patch_helper.py
```diff
@@ -107,14 +107,13 @@ def apply_patch(repo, patch_object):
             repo.index.commit('Patch: {0}'.format(os.path.basename(patch_name)), skip_hooks=True)
 
     @classmethod
-    def _update_deleted_patches(cls, deleted_patches, inapplicable_patches):
+    def _update_deleted_patches(cls, inapplicable_patches):
```
nforro: How does this commit fix anything? It simply removes obviously unnecessary code.

nforro: And the commit message definitely doesn't help. I think it would be best to merge this with those `base_name` => `patch_name` fixes and describe the reason behind that in commit message.

nforro:
This is not what I had in mind. Commit 86e897f2255aafa82c9866ebc8bbfbce2c2c8318 should be split into two, one containing `base_name` to `patch_name` substitutions plus the removed redundant code, and the other untouched patches. And the commit messages should briefly explain why those changes were made.

nforro:
The paths are now consistent, but absolute. They shoud be relative, first component being `rebase-helper-results`. Location of that directory could be mentioned as well for convenience.

skisela:
Yes I get it, the problem is I pushed it after testing on a relative path `--results-dir` which sets app.results_dir to a relative path, while the default is set to an absolute path. 
The one-line conversion from user given relative path to absolute will be part of this PR.

rebasehelper/output_tools/text_output_tool.py
```diff
@@ -62,54 +62,82 @@ def print_changes_patch(cls):
...
+            if patch_list:
+                for patch in patch_list:
+                    if patch_type == 'deleted':
+                        logger_report.info(' * {0:40} [-]'.format(os.path.basename(patch)))
+                    elif patch_type == 'modified':
+                        logger_report.info(' * {0:40} [*]'.format(os.path.basename(patch)))
+                    elif patch_type == 'inapplicable':
+                        logger_report.info(' * {0:40} [!]'.format(os.path.basename(patch)))
+                    else:
+                        logger_report.info(' * {0:40} [ ]'.format(os.path.basename(patch)))
```
nforro: You could use mapping dict, something like:

```python
symbols = dict(deleted='-', modified='*', inapplicable='!')
```

and then reference it like this:

```python
symbols.get(patch_type, ' ')
```

skisela: Awesome, thanks!

skisela:
I will also modify the CLI output at least to a level of shortening the paths, with the full path information to the rebase-helper-results directory.

nforro:
Commit message of 18bc323c90644c1588e83bfe84b875e5d0b6f818 is still insufficient. `Save patch names instead of paths` - save where? Why? Especially the **why** part is missing.
As for the other part, you should at least start it with `Also`, so it's clear that it doesn't directly relate with the subject.

This applies to other commit messages as well.

nforro:
More report notes:

Every list is indented by a space, except first part of  `rpmdiff` and `pkgdiff`. Why the inconsistency?

`pkgdiff` lists `pkgdiff.txt` and `diffs/`. Why `diffs/`? Why is HTML report not listed and why is it named just `pkgdiff` without any file extension?

Path to `changes.patch` is still absolute.

There are no SRPM logs listed, and RPMs are not sorted alphabetically (neither are patches).

rebasehelper/output_tools/text_output_tool.py
```diff
@@ -125,35 +151,29 @@ def print_summary(cls, path, results):
...
-                            logger_report.info("Log is available here: %s\n", checker)
-                        else:
-                            if isinstance(output, list):
-                                logger_report.info("%s See for more details %s", ','.join(output), checker)
-                            else:
-                                logger_report.info("%s See for more details %s", output, checker)
+            for check_tool in six.itervalues(checkers_runner.plugin_classes):
+                for check, data in sorted(six.iteritems(checkers_results)):
+                    if check == check_tool.get_checker_name():
+                        logger_report.info(check_tool.format(data).rstrip())
```
skisela: @nforro `rstrip` added here because of the extra line that logger adds anyway, which would cause inconsistency, as every line from checkers contains newline characters at the end and every headline from checker is preceeded with a newline character to logically divide the output sections.

rebasehelper/checkers/rpmdiff_tool.py
```diff
@@ -151,11 +152,34 @@ def run_check(cls, results_dir, **kwargs):
...
+    @classmethod
+    def format(cls, data):
+        """
+        Format rpmdiff output for outputtool
+        :param data: rpmdiff output data dictionary
+        :return: string formated output
+        """
+        output_string = cls.get_underlined_title("rpmdiff")
+
+        output_string += " - %s added files\n" % data['files_changes']['added']
```
skisela: Newline character moved to the end of string in each of checkers `format` method, as that is more human readable I suggest.

rebasehelper/checkers/pkgdiff_tool.py
```diff
@@ -242,11 +243,24 @@ def run_check(cls, results_dir, **kwargs):
...
+    @classmethod
+    def format(cls, data):
+        """
+        Formats pkgdiff data to string
+        :param data: pkgdiff data dictionary
+        :return: string formated output
+        """
+        output_string = cls.get_underlined_title("pkgdiff")
+        output_string += "Details in %s:\n" % data['path']
+        output_string += " - pkgdiff_reports.html\n"
```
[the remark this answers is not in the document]
skisela: Maybe even rename it to just pkgdiff.html and pkgdiff.txt.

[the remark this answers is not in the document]
nforro: Yes, `pkgdiff.{txt,html}` sounds best. I don't think any other files should be listed, as all of them should be linked from the main HTML file.

rebasehelper/output_tools/text_output_tool.py
nforro: Well, maybe you could just return a list of lines (without additional whitespacing).

rebasehelper/patch_helper.py
```diff
@@ -232,7 +229,7 @@ def _git_rebase(cls):
                     ret_code = 0
             else:
                 break
-        deleted_patches = cls._update_deleted_patches(deleted_patches, inapplicable_patches)
+        deleted_patches = cls._update_deleted_patches(inapplicable_patches)
```
nforro: Previous occurence of `deleted_patches` is unused after this change, you should remove it.

nforro:
Commit message of 8fc422d68ab02d1ab5fef102c4ca4fa5fed2bc69 could look like this:

>**Return only filenames of patches, not full paths**
>
>Returning full paths from GitPatchTool._git_rebase() breaks patch processing,
>because the results are matched against Patch tags, that contain
>filenames only.
>
>Also gather deleted_patches only after rebase is complete, no point doing it in process.

rebasehelper/output_tools/text_output_tool.py
skisela: Is this what you meant? 
https://github.com/rebase-helper/rebase-helper/pull/404/files#diff-3858a44859ef63796d6b6d43a35b9529R175

nforro: Yes, but you should probably rename `output_string` variable to something more appropriate.

rebasehelper/checker.py
```diff
@@ -40,6 +43,11 @@ def match(cls, cmd):
         """
         raise NotImplementedError()
 
+    @classmethod
+    def get_checker_output_dir_short(cls):
+        """Return short version of checker output directory"""
+        return os.path.join('rebase-helper-results/checkers', cls.CMD)
```
nforro: Since you are already using `os.path.join()`, you shouldn't use `/` directly. Also there is `REBASE_HELPER_RESULTS_DIR` constant defined in `settings.py`.

```python
return os.path.join(REBASE_HELPER_RESULTS_DIR, 'checkers', cls.CMD)
```

rebasehelper/checkers/pkgdiff_tool.py
```diff
@@ -207,8 +207,9 @@ def run_check(cls, results_dir, **kwargs):
         Compares old and new RPMs using pkgdiff
         :param results_dir result dir where are stored results
         """
-        cls.results_dir = results_dir
-        cls.pkgdiff_results_full_path = os.path.join(cls.results_dir, cls.pkgdiff_results_filename)
+        cls.results_dir = os.path.join(results_dir, 'pkgdiff')
```
nforro: You should probably use `cls.CMD` instead of `pkgdiff`.

rebasehelper/checkers/pkgdiff_tool.py
```diff
@@ -242,11 +243,24 @@ def run_check(cls, results_dir, **kwargs):
...
+    def format(cls, data):
+        """
+        Formats pkgdiff data to string
+        :param data: pkgdiff data dictionary
+        :return: formated pkgdiff list of strings
+        """
+        output_lines = [cls.get_underlined_title("pkgdiff")]
+        output_lines.append("Details in %s:" % data['path'])
+        output_lines.append(" - pkgdiff_reports.html")
+        output_lines.append(" - pkgdiff_reports.txt")
```
nforro: Here you should definitely make use of `cls.pkgdiff_results_filename`.

rebasehelper/checkers/pkgdiff_tool.py
```diff
@@ -38,7 +38,7 @@ class PkgDiffTool(BaseChecker):
 
     CMD = "pkgdiff"
     DEFAULT = True
-    pkgdiff_results_filename = 'pkgdiff_reports.html'
+    pkgdiff_results_filename = 'pkgdiff_reports'
```
nforro: I think this name doesn't make much sense. `pkgdiff` is redundant, and `reports` shouldn't be plural, as it's just one report. Simple `report` sound best to me.

rebasehelper/checkers/abipkgdiff_tool.py
```diff
@@ -89,6 +92,10 @@ def _find_debuginfo(cls, debug, pkg):
     @classmethod
     def run_check(cls, results_dir, **kwargs):
         """Compares old and new RPMs using abipkgdiff"""
+        # Check if ABI changes occured
+        cls.abi_changes = None
+        cls.results_dir = os.path.join(results_dir, 'abipkgdiff')
```
nforro: Again, you should use `cls.CMD`. Thinking about it, it would probably make more sense to rename it to `cls.NAME`.

rebasehelper/checkers/abipkgdiff_tool.py
```diff
@@ -112,17 +119,98 @@ def run_check(cls, results_dir, **kwargs):
...
                 raise RebaseHelperError('Execution of %s failed.\nCommand line is: %s' % (cls.CMD, cmd))
-            if int(ret_code) == 0:
-                text = 'ABI of the compared binaries in package %s are equal.' % old_name
-            else:
-                text = 'ABI of the compared binaries in package %s are not equal.' % old_name
-            reports[output] = text
-        return reports
+            reports[old_name] = int(ret_code)
+        return {'packages': cls.parse_abi_logs(reports), 'abi_changes': cls.abi_changes,
+                'path': cls.get_checker_output_dir_short()}
```
nforro: This is just my opinion, but I prefer to use `dict()` function to construct dictionaries, it's slightly more readable:

```python
return dict(packages=cls.parse_abi_logs(reports),
            abi_changes=cls.abi_changes,
            path=cls.get_checker_output_dir_short())
```

rebasehelper/checkers/abipkgdiff_tool.py
```diff
@@ -112,17 +119,98 @@ def run_check(cls, results_dir, **kwargs):
...
+            result = {}
+            for line in lines:
+                ms = summary_re.match(line)
+                if ms:
+                    ds = ms.groupdict()
+                    result[ds['kind']] = {}
+                    for mc in changes_re.finditer(ds['changes']):
+                        dc = mc.groupdict()
+                        if int(dc['count']) or dc['filtered_out']:
+                            result[ds['kind']][dc['what']] = {'count': dc['count'], 'filtered_out': dc['filtered_out']}
```
nforro: Is there a reason not to assign `dc` directly?

rebasehelper/checkers/abipkgdiff_tool.py
```diff
@@ -112,17 +119,98 @@ def run_check(cls, results_dir, **kwargs):
...
+                output_lines.append("%s" % sum_title)
+
+                for change_name, change_info in sorted(six.iteritems(changes_list)):
+                    if change_info['filtered_out']:
+                        output_lines.append(" - {} {} (filtered out {})".format(change_name,
+                                                                                 change_info['count'],
+                                                                                 change_info['filtered_out']))
+                    else:
+                        output_lines.append(" - {} {}".format(change_name, change_info['count']))
+        output_lines.append("Details in %s:" % data['path'])
```
nforro: You shouldn't mix `%` and `format()`. I actually think we should start to enforce using only `format()` in rebase-helper, `%` is going to be [removed](https://docs.python.org/release/3.1.5/library/stdtypes.html#old-string-formatting-operations) from Python in the future anyway.

rebasehelper/checkers/rpmdiff_tool.py
```diff
@@ -151,11 +151,34 @@ def run_check(cls, results_dir, **kwargs):
                 lines.append('Following files were %s:' % key)
                 lines.extend(val)
 
-        rpmdiff_report = os.path.join(cls.results_dir, 'report-' + cls.CMD + '.log')
+        rpmdiff_report = os.path.join(cls.results_dir, cls.CMD + '.txt')
```
nforro: I think `checkers/rpmdiff/report.txt` is better than `checkers/rpmdiff/rpmdiff.txt`.

rebasehelper/checkers/rpmdiff_tool.py
```diff
@@ -151,11 +151,34 @@ def run_check(cls, results_dir, **kwargs):
...
+
+        counts = {'added': len(results_dict['added']),
+                  'removed': len(results_dict['removed']),
+                  'changed': len(results_dict['changed'])}
```
nforro: How about this?

```python
counts = {k: len(v) for k, v in six.iteritems(results_dict)}
```

rebasehelper/output_tool.py
```diff
@@ -57,6 +58,13 @@ def print_cli_summary(cls, app):
         cls.app = app
         cls.print_patches_cli()
 
+        result = results_store.get_result_message()
+
+        if 'success' in result:
+            # checkers provide output only after successfully built packages
```
nforro: This comment is not accurate anymore.

rebasehelper/checkers/csmock_tool.py
```diff
@@ -22,7 +22,7 @@
 
 from __future__ import print_function
 import os
-from six import StringIO
+import six
```
nforro: You should also add a blank line before the import (`six` is third-party library).

rebasehelper/tests/test_output_tool.py
```diff
@@ -68,39 +68,51 @@ def results_store(self, workdir):
...
 - test-devel-1.2.2-1.x86_64.rpm
-Available New logs:
 - logfile3.log
-- logfile4.log
-=== Checker Results from checker(s) results ===
-Following files were moved
-/usr/sbin/test
-/usr/sbin/test2
-See for more details pkgdiff"""
+- logfile4.log""".format(workdir)
```
nforro: Since the string is really long, I would prefer to use keyword argument(s) in `format()`, so `.format(workdir=workdir)` and `Rebase helper results are located in {workdir}`.

rebasehelper/output_tool.py
```diff
@@ -70,21 +70,25 @@ def print_cli_summary(cls, app):
                 ConsoleHelper.cprint('\nABI changes occured. Check abipkgdiff output', color='red')
 
         ConsoleHelper.cprint('\nAvailable logs:', color='yellow')
-        print('{0}:\n{1}'.format('Debug log', app.debug_log_file))
+        print('{0}:\n{1}'.format('Debug log', cls.get_relative_results_path(os.path.relpath(app.debug_log_file,
+                                                                                            app.results_dir))))
```
nforro: Wouldn't it make more sense to incorporate `os.path.relpath(x, app.results_dir)` into `get_relative_results_path()`?

rebasehelper/output_tools/text_output_tool.py
```diff
@@ -62,54 +62,79 @@ def print_changes_patch(cls):
     @classmethod
     def print_message_and_separator(cls, message="", separator='='):
         logger_report.info(message)
-        logger_report.info(separator * len(message))
+        logger_report.info(separator * (len(message)-1))
```
nforro: `-` should be surrounded by spaces.

rebasehelper/output_tool.py
skisela: Well the paths used for `os.path.relpath` are supposed to be similiar, but often times only relative path is given. Currently I do not see how we could benefit from using it in `get_relative_results_path`

rebasehelper/output_tools/text_output_tool.py
```diff
@@ -46,114 +47,132 @@ def print_success_message(cls):
...
+                message = "\nSource packages and logs are in directory %s:"
+            else:
+                message = "\nBinary packages and logs are in directory %s:"
+
             if isinstance(srpm, str):
-                logger_report.info(message, type_rpm.upper(), os.path.dirname(srpm))
-                logger_report.info("- %s", os.path.basename(srpm))
+                # Print SRPM path
+                dirname = os.path.dirname(srpm)
+                logger_report.info(message, cls.get_relative_results_path(version.lower() + '-build/SRPM'))
```
nforro: You should use `os.path.join()` here instead of `/` (or `os.path.sep`). But wouldn't it be better to make `get_relative_results_path()` method to accept multiple arguments which would be joined together?

nforro: Also, you switched to using `format()` in logger methods, but here you are not using it. To be honest I'm not sure if we should get rid of lazy logging, it's true that it makes the code less consistent, but it isn't really the same thing as using the `%` operator.

rebasehelper/output_tool.py
```diff
@@ -68,21 +68,25 @@ def print_cli_summary(cls, app):
...
-            print('{0}:\n{1}'.format('Old build logs and (S)RPMs', os.path.join(app.results_dir, 'old-build')))
+            print('{0}:\n{1}'.format('Old build logs and (S)RPMs', cls.get_relative_results_path('old-build')))
         if results_store.get_new_build() is not None:
-            print('{0}:\n{1}'.format('New build logs and (S)RPMs', os.path.join(app.results_dir, 'new-build')))
+            print('{0}:\n{1}'.format('New build logs and (S)RPMs', cls.get_relative_results_path('new-build')))
         print('')
 
         ConsoleHelper.cprint('%s:' % 'Rebased sources', color='yellow')
-        print("%s" % app.rebased_sources_dir)
+        print("%s" % cls.get_relative_results_path('rebased-sources'))
```
nforro: You are still using the `%` operator here.

rebasehelper/output_tool.py
nforro: In that case I think the method should be renamed. `get_relative_results_path()` implies that is converts its arguments to a path relative to something, which is not exactly true.

rebasehelper/output_tool.py
```diff
@@ -48,6 +48,10 @@ def get_report_path(cls, app):
     def get_extension(cls):
         raise NotImplementedError
 
+    @classmethod
+    def get_relative_results_path(cls, append_path):
+        return os.path.join('rebase-helper-results', append_path)
```
nforro: You should use `REBASE_HELPER_RESULTS_DIR` constant here.

nforro:
Commit cf7610bf7ec0aa923d34b5a4698249b7ee9732fd is not supposed to be in the PR.

rebasehelper/checker.py
```diff
@@ -20,17 +20,22 @@
...
+from rebasehelper.settings import REBASE_HELPER_RESULTS_DIR
 
 
 class BaseChecker(object):
     """ Base class used for testing tool run on final pkgs. """
 
+    NAME = None
     DEFAULT = False
     category = None
+    results_dir = 'rebase-helper-results'
```
nforro: You should use `REBASE_HELPER_RESULTS_DIR` constant here.

nforro:
Patches in report are still not sorted alphabetically (as opposed to console output).

skisela:
Actually the CLI patches list is not alphabetically sorted:
```
deleted patches:
libtiff-hylafax-fix.patch
libtiff-CVE-2016-10266.patch
libtiff-CVE-2016-10267.patch
```

While the report patches list is sorted, but sorted according to the patch group type, which has more priority. 

Do you think it is better to ignore the patch group totally?

nforro:
Yes, that would make more sense. But the console output should be sorted as well.

rebasehelper/output_tool.py
```diff
@@ -47,6 +49,10 @@ def get_report_path(cls, app):
     def get_extension(cls):
         raise NotImplementedError
 
+    @classmethod
+    def prepend_results_dir_short(cls, *path_members):
```
nforro: What does `_short` mean?

rebasehelper/output_tools/text_output_tool.py
```diff
@@ -56,59 +57,86 @@ def print_changes_patch(cls):
         patch = cls.results_store.get_changes_patch()
         if patch is not None:
             logger_report.info('\nPatch with differences between old and new version source files:')
-            logger_report.info(patch['changes_patch'])
+            logger_report.info(cls.prepend_results_dir_short('changes.patch'))
```
nforro: Why are you discarding `patch['changes_patch']` and using hardcoded filename instead?

rebasehelper/output_tools/text_output_tool.py
```diff
@@ -56,59 +57,86 @@ def print_changes_patch(cls):
...
+                message = "\nSource packages and logs are in directory %s:"
+            else:
+                message = "\nBinary packages and logs are in directory %s:"
+
             if isinstance(srpm, str):
-                logger_report.info(message, type_rpm.upper(), os.path.dirname(srpm))
-                logger_report.info("- %s", os.path.basename(srpm))
+                # Print SRPM path
+                dirname = os.path.dirname(srpm)
+                logger_report.info(message, cls.prepend_results_dir_short(version.lower() + '-build/SRPM'))
```
nforro: No, this should look like this:

```python
logger_report.info(message, cls.prepend_results_dir_short(version.lower() + '-build', 'SRPM'))
```

rebasehelper/output_tools/text_output_tool.py
skisela: I thought it's more convenient, than extracting file base name from the full path stored in `results_store`.

rebasehelper/output_tool.py
skisela: I wanted to highlight the fact, that it is not the full results dir path.

rebasehelper/output_tools/text_output_tool.py
skisela: Yes you are right, thanks, I missed this one.

rebasehelper/output_tool.py
nforro: By full you mean absolute? Anyway, if you must append anything, `prepend_results_dir_name` sounds better to me.

rebasehelper/output_tools/text_output_tool.py
nforro: Then why are you storing the value at all? It doesn't make sense.

nforro:
I don't know if you are aware, but the paths in checker reports are absolute again.

rebasehelper/output_tool.py
```diff
@@ -55,25 +61,32 @@ def print_cli_summary(cls, app):
...
         logger_output.heading('\nAvailable logs:')
-        logger_output.info('%s:\n%s', 'Debug log', app.debug_log_file)
+        logger_output.info('%s:\n%s', 'Debug log', cls.prepend_results_dir_name(os.path.relpath(app.debug_log_file,
+                                                                                app.results_dir)))
         if results_store.get_old_build() is not None:
-            logger_output.info('%s:\n%s', 'Old build logs and (S)RPMs', os.path.join(app.results_dir, 'old-build'))
+            logger_output.info('%s:\n%s', 'Old build logs and (S)RPMs', cls.prepend_results_dir_name('old-build'))
         if results_store.get_new_build() is not None:
-            logger_output.info('%s:\n%s', 'New build logs and (S)RPMs', os.path.join(app.results_dir, 'new-build'))
+            logger_output.info('%s:\n%s', 'New build logs and (S)RPMs', cls.prepend_results_dir_name('old-build'))
```
nforro: This is obviously a copy-paste error - should be `new-build`.

rebasehelper/output_tools/text_output_tool.py
```diff
@@ -120,40 +148,35 @@ def print_summary(cls, path, results):
...
             LoggerHelper.add_file_handler(logger_report, path)
         except (OSError, IOError):
-            raise RebaseHelperError("Can not create results file '%s'" % path)
+            raise RebaseHelperError("Can not create results file '{}'".format(path))
 
         cls.results_store = results
 
         cls.print_success_message()
-        type_pkgs = ['old', 'new']
+        logger_report.info("Rebase helper results are located in %s", os.path.dirname(path))
```
nforro: What is `Rebase helper`? We should refer to it as `rebase-helper`. But I think something like this would be better here:
`All result files are stored in %s`.

nforro:
I don't see a reference to a SRPM build log:

```
RPMS
====

Old packages
------------

Source packages and logs are in directory rebase-helper-results/old-build/SRPM:
 - libtiff-4.0.7-5.fc26.src.rpm

Binary packages and logs are in directory rebase-helper-results/old-build/RPM:
 - libtiff-4.0.7-5.fc26.x86_64.rpm
 - libtiff-debuginfo-4.0.7-5.fc26.x86_64.rpm
 - libtiff-devel-4.0.7-5.fc26.x86_64.rpm
 - libtiff-static-4.0.7-5.fc26.x86_64.rpm
 - libtiff-tools-4.0.7-5.fc26.x86_64.rpm
 - build.log

New packages
------------

Source packages and logs are in directory rebase-helper-results/new-build/SRPM:
 - libtiff-4.0.9-1.fc26.src.rpm

Binary packages and logs are in directory rebase-helper-results/new-build/RPM:
 - libtiff-4.0.9-1.fc26.x86_64.rpm
 - libtiff-debuginfo-4.0.9-1.fc26.x86_64.rpm
 - libtiff-devel-4.0.9-1.fc26.x86_64.rpm
 - libtiff-static-4.0.9-1.fc26.x86_64.rpm
 - libtiff-tools-4.0.9-1.fc26.x86_64.rpm
 - build.log
```

skisela:
The string is now fixed as you suggested and I added a commit: https://github.com/rebase-helper/rebase-helper/pull/404/commits/c2e83ff6e416260db5215e7bc0cf851299266dfc#diff-55e1f3684ba905b1c1f92175ff8187d2R579

I think the srpm build logs might have begun disappearing after https://github.com/rebase-helper/rebase-helper/pull/446 but I might be wrong.

PR #409, opened 2017-12-26
Rewrite DownloadHelper using requests

FrNecas:


rebasehelper/utils.py
```diff
@@ -285,61 +287,56 @@ def format_size(s):
...
         Method for downloading file from HTTP, HTTPS and FTP URL.
 
         :param url: URL from which to download the file
         :param destination_path: path where to store downloaded file
-        :param timeout: timeout in seconds for blocking actions like connecting, etc.
         :param blocksize: size in Bytes of blocks used for downloading the file and reporting progress
         :return: None
         """
+        requests_ftp.monkeypatch_session()
+        session = requests.Session()
```
nforro: There is no reason to monkeypatch `requests.Session`, this should be written as:
```python
session = requests_ftp.FTPSession()
```

rebasehelper/utils.py
```diff
@@ -285,61 +287,56 @@ def format_size(s):
...
+        r = session.get(url)
+        if r.status_code == 404:
```
nforro: What about other HTTP and FTP status codes?

rebasehelper/utils.py
```diff
@@ -285,61 +287,56 @@ def format_size(s):
...
+            raise DownloadError
+
+        lowercase_headers = {k.lower(): v for k, v in r.headers.items()}
```
nforro: `r.headers` is of type `requests.structures.CaseInsensitiveDict`, so this should be unnecessary.

rebasehelper/utils.py
```diff
@@ -1198,7 +1199,7 @@ def download_build(cls, client, build_id, destination):
         logs = []
         for _, url in six.iteritems(result.data['results_by_chroot']):
             url = url if url.endswith('/') else url + '/'
-            d = pyquery.PyQuery(url, opener=lambda x: urllib.request.urlopen(x))
+            d = pyquery.PyQuery(url, opener=lambda x: requests.get(x))
```
nforro: Actually `PyQuery` uses `requests` by default, so removing `opener` argument should be sufficient, see:
https://github.com/gawel/pyquery/blob/master/pyquery/openers.py

rebasehelper/utils.py
```diff
@@ -285,61 +287,54 @@ def format_size(s):
...
         :param url: URL from which to download the file
         :param destination_path: path where to store downloaded file
-        :param timeout: timeout in seconds for blocking actions like connecting, etc.
         :param blocksize: size in Bytes of blocks used for downloading the file and reporting progress
         :return: None
         """
+        session = requests_ftp.ftp.FTPSession()
+        r = session.get(url)
+        if not 200 <= r.status_code < 300:
+            raise DownloadError(r.url, r.status_code, r.reason, r.headers, None)
```
nforro: `DownloadError` should contain reason string only, see https://github.com/rebase-helper/rebase-helper/blob/master/rebasehelper/specfile.py#L163-L165.

PR #414, opened 2018-01-03
Get error from koji and process it

skisela:
As mentioned in https://pagure.io/koji/issue/712 getTaskResult raises exception which contains koji error message, in case the build failed.

### DIFF
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -42,6 +42,7 @@ from rebasehelper.patch_helper import Patcher
 from rebasehelper.exceptions import RebaseHelperError, CheckerNotFoundError
 from rebasehelper.results_store import results_store
 from rebasehelper.versioneer import versioneers_runner
+from rebasehelper.build_tools.mock_tool import MockBuildTool
 from rebasehelper import version
 
 
@@ -580,12 +581,16 @@ class Application(object):
                 build_dict['binary_package_build_error'] = six.text_type(e)
                 results_store.set_build_data(version, build_dict)
 
-                if e.logfile is None:
+                if e.return_code:
+                    # koji was used for build therefore only status code is set
+                    e.logfile = MockBuildTool.get_mock_logfile_path(e.return_code,
+                                                                    os.path.join(results_dir, 'RPM'))
+                if e.logfile:
+                    msg = 'Building {} RPM packages failed; see {} for more information'.format(version, e.logfile)
+                else:
                     msg = 'Building {} RPM packages failed; see logs in {} for more information'.format(
                         version, os.path.join(results_dir, 'RPM')
                     )
-                else:
-                    msg = 'Building {} RPM packages failed; see {} for more information'.format(version, e.logfile)
                 raise RebaseHelperError(msg, logfiles=builder.get_logs().get('logs'))
 
             except Exception as e:
--- a/rebasehelper/build_tools/koji_tool.py
+++ b/rebasehelper/build_tools/koji_tool.py
@@ -23,6 +23,7 @@
 import os
 import six
 import koji  # pylint: disable=import-error
+import re
 
 # unused import needed to prevent loading koji buildtool with Koji < 1.13
 import koji_cli.lib  # pylint: disable=import-error
@@ -90,8 +91,14 @@ class KojiBuildTool(BuildToolBase):
         task_dict = cls.koji_helper.watch_koji_tasks(session, [task_id])
         task_list = []
         package_failed = False
+        build_error = None
         for key in six.iterkeys(task_dict):
             if task_dict[key] == koji.TASK_STATES['FAILED']:
+                try:
+                    # call getTaskResult, as it raises an exception containing desired error message
+                    session.getTaskResult(key)
+                except koji.BuildError as e:
+                    build_error = six.text_type(e)
                 package_failed = True
             task_list.append(key)
         rpms, logs = cls.koji_helper.download_scratch_build(session,
@@ -102,9 +109,25 @@ class KojiBuildTool(BuildToolBase):
             logger.info('RPM build failed %s', weburl)
             logs.append(weburl)
             cls.logs.append(weburl)
-            raise BinaryPackageBuildError
+            cls.logs.extend(logs)
+            raise BinaryPackageBuildError(return_code=cls.get_mock_return_code(build_error))
         logs.append(weburl)
         return rpms, logs, task_id
+
+    @classmethod
+    def get_mock_return_code(cls, msg):
+        """
+        Parses koji status message for mock build return code
+
+        :param msg: mock status message
+        :return: returns mock build return code
+        """
+        match = re.search(r'mock exited with status (\d+)', msg)
+        if match:
+            # Example error message that can be parsed:
+            # BuildError: error building package (arch noarch),
+            # mock exited with status 1; see build.log for more information
+            return int(match.group(1))
 
     @classmethod
     def get_logs(cls):
--- a/rebasehelper/build_helper.py
+++ b/rebasehelper/build_helper.py
@@ -57,6 +57,8 @@ class BinaryPackageBuildError(RuntimeError):
         :param kwargs: dictionary containing path to the logfile that contains main errors
         """
         self.args = args
+        # Return code obtained from koji only at this time
+        self.return_code = kwargs.get('return_code')
         self.logfile = kwargs.get('logfile')
 
 
--- a/rebasehelper/build_tools/mock_tool.py
+++ b/rebasehelper/build_tools/mock_tool.py
@@ -80,15 +80,36 @@ class MockBuildTool(BuildToolBase):
 
         ret = ProcessHelper.run_subprocess(cmd, output=output)
 
+        if ret == 0:
+            return [f for f in PathHelper.find_all_files(results_dir, '*.rpm') if not f.endswith('.src.rpm')]
+        else:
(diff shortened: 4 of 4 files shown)


rebasehelper/build_tools/koji_tool.py
```diff
@@ -90,8 +90,13 @@ def _scratch_build(cls, source, **kwargs):
         task_dict = cls.koji_helper.watch_koji_tasks(session, [task_id])
         task_list = []
         package_failed = False
+        build_error = ''
```
nforro: Wouldn't it make more sense to set the variable to `None`?

rebasehelper/build_tools/koji_tool.py
```diff
@@ -90,8 +90,13 @@ def _scratch_build(cls, source, **kwargs):
...
         for key in six.iterkeys(task_dict):
             if task_dict[key] == koji.TASK_STATES['FAILED']:
+                try:
+                    info = session.getTaskResult(key)
```
nforro: There is no reason to assign result of `getTaskResult()` method to a variable, it's not used anyway.

rebasehelper/build_tools/koji_tool.py
```diff
@@ -90,8 +90,13 @@ def _scratch_build(cls, source, **kwargs):
...
+                except Exception as result:
```
nforro: Why `result`? Why not `error` or simply `e`?

nforro: Also you should catch specific exception type, not general `Exception`.

rebasehelper/build_tools/koji_tool.py
skisela: You are right, but I consider it better for readability, to be more obvious, that it actually gets the results. I can get rid of the assignment.

rebasehelper/build_tools/koji_tool.py
```diff
@@ -90,8 +90,13 @@ def _scratch_build(cls, source, **kwargs):
...
+                    build_error = result
```
nforro: I think it would be better to store error string to `build_error`, not `Exception` instance.

rebasehelper/build_tools/koji_tool.py
nforro: If you think it would be confusing, add a comment with explanation.

rebasehelper/build_tools/koji_tool.py
skisela: You are correct, I will fix it, as well as the comment, as the exception raised is BuildError, not the one stated.

rebasehelper/build_tools/koji_tool.py
```diff
@@ -90,8 +90,14 @@ def _scratch_build(cls, source, **kwargs):
...
         task_list = []
         package_failed = False
+        build_error = None
         for key in six.iterkeys(task_dict):
             if task_dict[key] == koji.TASK_STATES['FAILED']:
+                try:
+                    # Gets the task results but the value is not used in this function currently
+                    info = session.getTaskResult(key)
+                except koji.BuildError as e:
+                    build_error = e.__str__()
```
nforro: You shoud use `str(e)`, or rather safer `six.text_type(e)`.

rebasehelper/build_tools/koji_tool.py
```diff
@@ -90,8 +90,14 @@ def _scratch_build(cls, source, **kwargs):
         task_dict = cls.koji_helper.watch_koji_tasks(session, [task_id])
         task_list = []
         package_failed = False
+        build_error = None
         for key in six.iterkeys(task_dict):
             if task_dict[key] == koji.TASK_STATES['FAILED']:
+                try:
+                    # Gets the task results but the value is not used in this function currently
+                    info = session.getTaskResult(key)
```
nforro: The unused variable is still there ;-)
And the comment doesn't really explain anything, what I had in mind is something like this:
```python
# call getTaskResult, as it raises an exception containing desired error message
```

skisela: Alright, I thought you meant commenting, why the assignment is there. I get it now.

rebasehelper/application.py
```diff
@@ -576,7 +576,10 @@ def build_packages(self):
                 build_dict['binary_package_build_error'] = six.text_type(e)
                 results_store.set_build_data(version, build_dict)
 
-                if e.logfile is None:
+                if e.exact_error:
+                    # value acquired from koji.getTaskResult by raising koji.BuildError exception
+                    msg = e.exact_error + ' in ' + results_dir
```
nforro: What exactly does this print? Do you have an example?

skisela: koji error message + location of where the log is located.
```error building package (arch noarch), mock exited with status 1; see build.log for more information in /home/skisela/fedora/rubygem-bundler/rebase-helper-results/new-build```

Comment (skisela) [reply rebasehelper/application.py]: Maybe there should also be code checking if there is some *.log at all, even though majority of build errors contains it.

Comment (nforro) [reply rebasehelper/application.py]: Right, so it relies on the format of the error message. That is not the best thing to do, as it could change anytime, but since you do it anyway, you could as well parse the mock error code out of it and use it to determine the logfile.

File: rebasehelper/build_tools/koji_tool.py [hunk 8]
```diff
@@ -102,10 +109,28 @@ def _scratch_build(cls, source, **kwargs):
             logger.info('RPM build failed %s', weburl)
             logs.append(weburl)
             cls.logs.append(weburl)
-            raise BinaryPackageBuildError
+            return_code = cls.get_mock_return_code(build_error)
+            if return_code:
+                raise BinaryPackageBuildError(return_code=return_code)
+            else:
+                # Unknown error message obtained from koji
+                raise BinaryPackageBuildError()
```
nforro: Why not to write it just like this?
```python
raise BinaryPackageBuildError(return_code=cls.get_mock_return_code(build_error))
```

rebasehelper/build_tools/koji_tool.py
```diff
@@ -102,10 +109,28 @@ def _scratch_build(cls, source, **kwargs):
...
+    @classmethod
+    def get_mock_return_code(cls, msg):
+        """
+        Parses mock status message for mock build return code
+
+        :param msg: mock status message
+        :return: returns mock build return code
+        """
+        if re.search(r'status [0-9]+', msg):
+            found = re.search(r'[0-9]+', msg)
```
nforro: This could definitely be done with just one `re.search` call. I would also rather match for more specific string, like `mock exited with status \d+`.
Also it would be nice to include a comment with typical error message for which the regexp is valid, to be able to update it in case the message changes in the future.

rebasehelper/build_tools/koji_tool.py
```diff
@@ -102,10 +109,28 @@ def _scratch_build(cls, source, **kwargs):
...
+            else:
+                # Unknown error message obtained from koji
+                raise BinaryPackageBuildError()
         logs.append(weburl)
         return rpms, logs, task_id
 
+    @classmethod
+    def get_mock_return_code(cls, msg):
+        """
+        Parses mock status message for mock build return code
```
nforro: Shouldn't this be rather `Parses koji status message for mock build return code`?

rebasehelper/application.py
```diff
@@ -576,11 +577,16 @@ def build_packages(self):
...
+                if e.logfile is None and e.return_code is None:
                     msg = 'Building {} RPM packages failed; see logs in {} for more information'.format(
                         version, os.path.join(results_dir, 'RPM')
                     )
                 else:
+                    if e.logfile is None:
+                        # koji was used for build therefore only status code is set
+                        e.logfile = MockBuildTool.get_mock_logfile_path(e.return_code,
+                                                                        os.path.join(results_dir, 'RPM'))
+
```
nforro: I think it would make more sense to write this part of code like this:
```python
if e.return_code:
    # set e.logfile
if e.logfile:
    ...
else:
    ...
```

rebasehelper/build_tools/koji_tool.py
```diff
@@ -102,10 +109,27 @@ def _scratch_build(cls, source, **kwargs):
...
+
+        :param msg: mock status message
+        :return: returns mock build return code
+        """
+        found = re.search(r'mock exited with status \d+', msg)
+        if found:
+            # Example error message that can be parsed:
+            # BuildError: error building package (arch noarch),
+            # mock exited with status 1; see build.log for more information
+            return int(found.group(0).split()[-1])
```
nforro: I have to admit this is quite original solution ;-)
But I would prefer something like this instead:
```python
match = re.search(r'mock exited with status (\d+)', msg)
if match:
    return int(match.group(1))
```

skisela: Is it safe attempting to access a group, that is potentially non-existent?

skisela: Yes you're right. Thanks!

rebasehelper/build_tools/koji_tool.py
```diff
@@ -102,10 +109,26 @@ def _scratch_build(cls, source, **kwargs):
...
+
+        :param msg: mock status message
+        :return: returns mock build return code
+        """
+        match = re.search(r'mock exited with status \d+', msg)
+        if match:
+            # Example error message that can be parsed:
+            # BuildError: error building package (arch noarch),
+            # mock exited with status 1; see build.log for more information
+            return int(match.group(1))
```
nforro: This will fail, as there is no group in the regexp.

PR #427, opened 2018-01-15
Add option to make changes to specfile between builds

skisela:
Sometimes the change to specfile is as easy as adding one line only.
The problem below can be fixed by adding the following to `%files` section:
```
%{_libdir}/fcitx/qt/libfcitx-quickphrase-editor5.so
```
Add option to pause between builds and modify the specfile used for new build in `rebased_sources`

Helps fixing this kind of errors with ease:
```
error: Installed (but unpackaged) file(s) found:
    /usr/lib64/fcitx/qt/libfcitx-quickphrase-editor5.so
    Installed (but unpackaged) file(s) found:
    /usr/lib64/fcitx/qt/libfcitx-quickphrase-editor5.so
RPM build errors:
Child return code was: 1
```

### DIFF
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -35,7 +35,7 @@ from rebasehelper.specfile import SpecFile, get_rebase_name, spec_hooks_runner
 from rebasehelper.logger import logger, logger_report, LoggerHelper
 from rebasehelper import settings
 from rebasehelper.output_tool import output_tools_runner
-from rebasehelper.utils import PathHelper, GitHelper, KojiHelper, FileHelper, MacroHelper
+from rebasehelper.utils import PathHelper, GitHelper, KojiHelper, FileHelper, MacroHelper, ConsoleHelper
 from rebasehelper.checker import checkers_runner
 from rebasehelper.build_helper import Builder, SourcePackageBuildError, BinaryPackageBuildError
 from rebasehelper.patch_helper import Patcher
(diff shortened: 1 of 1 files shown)


nforro:
Not a bad idea, but the option should be mutually exclusive with `--non-interactive`.

If I understand it correctly, this counts with an user doing a rebase twice. First time to get the build log with an error message, and second time to modify the spec file. Wouldn't it be better to offer an user a possibility to both modify the spec file and retry the new build after it fails?

skisela:
Yes you are right with `--non-interactive`.

Well we already got rid of the option, which started the rebase the second time.

You mean asking the user after a failed build, if he wants to try it again with the pause between builds? Which would be automatically disabled with --non-interactive.

nforro:
Yes, something like:
```
Build failed; see rebase-helper-results/.../build.log for more information.
Do you want to try it one more time? (Feel free to modify rebase-helper-results/.../package.spec in the meantime)
```

skisela:
@nforro Do you think, that it is not too annoying, to have an infinite while loop, which runs either until successful build is created, or interrupted by the user?

nforro:
No, that should be fine IMHO.

rebasehelper/application.py
```diff
@@ -490,113 +490,143 @@ def build_packages(self):
...
+                        msg = 'Building {} RPM packages failed; see logs in {} for more information'.format(
+                            version, os.path.join(results_dir, 'RPM')
+                        )
+                    else:
+                        msg = 'Building {} RPM packages failed; see {} for more information'.format(version, e.logfile)
+
+                    logger.info(msg)
+                    if not self.conf.non_interactive and \
+                            ConsoleHelper.get_message('Do you want to try it one more time?'
+                                                      ' (You will be asked to change %s)' % rebased_spec_path):
```
nforro: > You will be asked to change ...

This doesn't seem right, because user is not asked that actually. I'd rather see something like this (the wording could definitely be improved):

```
...
Do you want to try it one more time? [Y/n] y
Now it's time to modify rebase-helper-results/rebased-sources/foo.spec if necessary.
Do you want to continue with the rebuild now? [Y/n] y
...
```

skisela: I was thinking about the wording, and this seems most suitable. Thanks!

rebasehelper/application.py
```diff
@@ -490,113 +490,139 @@ def build_packages(self):
...
-                                               'differs from version in SPEC file (%s)!',
-                                               koji_version, koji_build_id, pkg_version)
-                            pkg_version = pkg_full_version = koji_version
+            successful_builds = 0
+            try_build_again = False
+
+            while successful_builds < 1:
+
+                spec_object = self.spec_file if version == 'old' else self.rebase_spec_file
+                rebased_spec_path = PathHelper.find_first_file(self.rebased_sources_dir, '*.spec', 0)
```
nforro: This seems unnecessary, can't you use `self.rebase_spec_file.path`?

PR #433, opened 2018-01-26
Do not use empty list as a default argument

FrNecas:


rebasehelper/specfile.py
```diff
@@ -360,14 +360,19 @@ def get_not_used_patches(self):
...
-    def _process_patches(self, comment_out=[], remove_patches=[], disable_inapplicable_patches=None):
+    def _process_patches(self, comment_out=None, remove_patches=None, disable_inapplicable_patches=None):
         """
         Comment out and delete patches from SPEC file
 
         :var comment_out: list with patch numbers to comment out
         :var remove_patches: list with patch numbers to delete
         :var disable_inapplicable_patches: boolean value deciding if the inapplicable patches should be commented out
         """
+        if not comment_out:
```
nforro: I think `if comment_out is None:` would be more appropriate here.

rebasehelper/specfile.py
```diff
@@ -360,14 +360,19 @@ def get_not_used_patches(self):
...
-    def _process_patches(self, comment_out=[], remove_patches=[], disable_inapplicable_patches=None):
+    def _process_patches(self, comment_out=None, remove_patches=None, disable_inapplicable_patches=None):
         """
         Comment out and delete patches from SPEC file
 
         :var comment_out: list with patch numbers to comment out
         :var remove_patches: list with patch numbers to delete
         :var disable_inapplicable_patches: boolean value deciding if the inapplicable patches should be commented out
         """
+        if comment_out is None:
```
nforro: > I think `if comment_out is None:` would be more appropriate here.

Of course this is the case also for the other two appearances.

PR #437, opened 2018-01-30
Update data of both spec_file and rebase_spec_file

nforro:


nforro:
Actually, this bug would have been caught by pylint if warnings were enabled:
`Using possibly undefined loop variable 'spec_file' (undefined-loop-variable)`

But we would need to fix every warning first, otherwise no PR would pass ;-) Maybe we can figure out a way how to do only differential checks, so that it fails only when a new warning appears.

rebasehelper/application.py
```diff
@@ -224,7 +224,8 @@ def _prepare_spec_objects(self):
                 spec_file.download_remote_sources()
 
         # parse spec again with sources downloaded to properly expand %prep section
-        spec_file._update_data()
+        for spec_file in [self.spec_file, self.rebase_spec_file]:
+            spec_file._update_data()
```
skisela: Can't this be actually moved into the loop just above?

nforro: Yes, that would be probably better.

PR #439, opened 2018-01-31
Add license checker

skisela:


rebasehelper/checkers/license_tool.py
```diff
@@ -0,0 +1,131 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Authors: Petr Hracek <phracek@redhat.com>
+#          Tomas Hozza <thozza@redhat.com>
+
+from __future__ import print_function
```
nforro: Why do you need `print_function`?

rebasehelper/checkers/license_tool.py
```diff
@@ -0,0 +1,131 @@
...
+import os
```
nforro: Missing empty lines around this import.

rebasehelper/checkers/license_tool.py
```diff
@@ -0,0 +1,131 @@
...
+        new_sources_path = os.path.join(cls.workspace_dir, 'new_sources')
+
+        # Get paths to directories of sources(diff would not handle it, if directory names were different)
+        # Examples of the paths (old_sources/fcitx-qt5-1.1.1, new_sources/fcitx-qt5-1.2.2)
+        old_dir = os.path.join(old_sources_path, os.listdir(old_sources_path)[0])
+        new_dir = os.path.join(new_sources_path, os.listdir(new_sources_path)[0])
+
+        # Example diff output
+        # 'Files fcitx-qt5-1.1.1/guiwrapper/common.h and fcitx-qt5-1.2.2/guiwrapper/common.h differ'
+        ProcessHelper.run_subprocess(["/usr/bin/diff", '-q', '-r', old_dir, new_dir], output_file=changed_files)
```
nforro: Is calling `diff` really necessary? It's another dependency, and `licensecheck` has `--recursive` option.

setup.py
```diff
@@ -100,6 +100,7 @@ def get_requirements():
             'pkgdiff = rebasehelper.checkers.pkgdiff_tool:PkgDiffTool',
             'abipkgdiff = rebasehelper.checkers.abipkgdiff_tool:AbiCheckerTool',
             'csmock = rebasehelper.checkers.csmock_tool:CsmockTool',
+            'license = rebasehelper.checkers.license_tool:LicenseTool',
```
nforro: I really don't think you should leave out the `check` part from the name.

rebasehelper/checkers/license_tool.py
```diff
@@ -0,0 +1,99 @@
...
+    @classmethod
+    def get_checker_name(cls):
+        return cls.CMD
+
+    @classmethod
+    def is_default(cls):
+        return cls.DEFAULT
+
+    @classmethod
+    def get_license_changes(cls):
```
nforro: This method is not called anywhere?

rebasehelper/checkers/license_tool.py
```diff
@@ -0,0 +1,99 @@
...
+            ProcessHelper.run_subprocess(["/usr/bin/licensecheck", source_dir, "--machine", "--recursive"],
+                                         output_file=out)
+            diff = set()
+            for l in out:
+                diff.add(l.split('\t')[1])
+            diffs.append(diff)
+
+        for new_licenses in diffs[1]:
+            if new_licenses not in diffs[0]:
+                cls.license_changes = True
```
nforro: Ok, this only checks if some specific license newly appeared. I don't think that's enough. While it shouldn't matter when some file changes license to some different but already present license, I think it's worth mentioning to a maintaner.

Actually, this tool should ideally make list of all files which changed license, group it by specific license transitions and save it as a report.

rebasehelper/output_tool.py
```diff
@@ -64,6 +65,8 @@ def print_cli_summary(cls, app):
             # checkers provide output only after successfully built packages
             if AbiCheckerTool.abi_changes:
                 ConsoleHelper.cprint('\nABI changes occured. Check abipkgdiff output', color='red')
+            if LicenseTool.license_changes:
+                ConsoleHelper.cprint('\nLicense changes occured. Check licensecheck output', color='red')
```
nforro: Here you are referring to a licensecheck output, but there isn't any, right?

setup.py
```diff
@@ -100,6 +100,7 @@ def get_requirements():
             'pkgdiff = rebasehelper.checkers.pkgdiff_tool:PkgDiffTool',
             'abipkgdiff = rebasehelper.checkers.abipkgdiff_tool:AbiCheckerTool',
             'csmock = rebasehelper.checkers.csmock_tool:CsmockTool',
+            'licensecheck = rebasehelper.checkers.license_tool:LicenseTool',
```
nforro: Why `license_tool:LicenseTool` and not `licensecheck_tool:LicenseCheckTool`?

rebasehelper/checkers/license_tool.py
skisela: You are right, I will modify it.

skisela: So far the output looks like this on fcitx-qt5 package:
```
licensecheck
======
License changes occured!
Check the following files:
/home/skisela/fedora/fcitx-qt5/rebase-helper-workspace/new_sources/fcitx-qt5-1.2.2/platforminputcontext/qfcitxplatforminputcontext.cpp license change:GPL (v2 or later) => UNKNOWN
/home/skisela/fedora/fcitx-qt5/rebase-helper-workspace/new_sources/fcitx-qt5-1.2.2/platforminputcontext/main.h license change:GPL (v2 or later) => UNKNOWN
/home/skisela/fedora/fcitx-qt5/rebase-helper-workspace/new_sources/fcitx-qt5-1.2.2/guiwrapper/main.cpp license change:GPL (v2 or later) => GPL (v3 or later)
/home/skisela/fedora/fcitx-qt5/rebase-helper-workspace/new_sources/fcitx-qt5-1.2.2/platforminputcontext/main.cpp license change:GPL (v3 or later) => UNKNOWN
/home/skisela/fedora/fcitx-qt5/rebase-helper-workspace/new_sources/fcitx-qt5-1.2.2/platforminputcontext/main.cpp license change:GPL (v2 or later) => UNKNOWN
/home/skisela/fedora/fcitx-qt5/rebase-helper-workspace/new_sources/fcitx-qt5-1.2.2/quickphrase-editor/main.h license change:GPL (v2 or later) => GPL (v3 or later)
/home/skisela/fedora/fcitx-qt5/rebase-helper-workspace/new_sources/fcitx-qt5-1.2.2/quickphrase-editor/main.cpp license change:GPL (v2 or later) => GPL (v3 or later)
/home/skisela/fedora/fcitx-qt5/rebase-helper-workspace/new_sources/fcitx-qt5-1.2.2/platforminputcontext/qfcitxplatforminputcontext.h license change:GPL (v2 or later) => UNKNOWN
/home/skisela/fedora/fcitx-qt5/rebase-helper-workspace/new_sources/fcitx-qt5-1.2.2/COPYING.BSD license change:New file => BSD (3 clause)
```

skisela: Where `New file => ${some_license}` means a new file appeared and `${some_license} => UNKNOWN` means it does not appear in the new sources. I guess substituting it with "deleted" is more suitable.
Also a newline between file_path and license change is appropriate.

skisela: I guess it is worth mentioning only the changed licenses, not the deleted+added(not new ones)

rebasehelper/output_tool.py
```diff
@@ -64,6 +65,8 @@ def print_cli_summary(cls, app):
             # checkers provide output only after successfully built packages
             if AbiCheckerTool.abi_changes:
                 ConsoleHelper.cprint('\nABI changes occured. Check abipkgdiff output', color='red')
+            if LicenseCheckTool.license_changes:
```
skisela: @nforro Do you think it is appropriate to inform about license changes even after a failed build? I suppose it is.

nforro: Definitely, that's actually the main reason for running different checkers at different phases of a rebase.

rebasehelper/checkers/license_tool.py
nforro: No, I think additions and deletions should be included as well, but perhaps in a different way than changes. Maintainer needs to know whether there is a possibility that the `License` tag in a spec file requires adjustment.

And the output should be grouped by license transition, to make it more compact and readable.

rebasehelper/checkers/licensecheck_tool.py
```diff
@@ -0,0 +1,130 @@
...
+        old_lics, new_lics = set(), set()
+        for new_file, new_license in six.iteritems(diffs[1]):
+            new_lics.add(new_license)
+            for old_file, old_license in six.iteritems(diffs[0]):
+                if old_license != new_license == 'UNKNOWN':
+                    continue
+                old_lics.add(old_license)
+                # Get the important file path part strip ".../new_sources/package-{version}/..." part away
+                new_f_path = new_file.split('/')[new_file.split('/').index('new_sources')+2:]
+                old_f_path = old_file.split('/')[old_file.split('/').index('old_sources')+2:]
```
nforro: Surely there is more elegant way how to do this. What about:

```python
new_f_path = os.path.relpath(new_file, new_sources_path).split(os.path.sep)[1:]
old_f_path = os.path.relpath(old_file, old_sources_path).split(os.path.sep)[1:]
```

rebasehelper/checkers/licensecheck_tool.py
```diff
@@ -0,0 +1,130 @@
...
+    @classmethod
+    def get_checker_name(cls):
+        return cls.CMD
+
+    @classmethod
+    def is_default(cls):
+        return cls.DEFAULT
+
+    @classmethod
+    def get_license_changes(cls):
```
nforro: This is how I think the output of this method should look like:

```python
{
    'added': {
        'FDL': [
            'some/file.txt',
        ],
        'GPL': [
            'some/other/file.txt',
        ],
    },
    'changed': {
        'BSD => MIT': [
            'some/file.txt',
            'some/other/file.txt',
        ],
    },
    'removed': {
        'GPL': [
            'some/file.txt',
        ],
    },
}
```

where `added` are newly appearing files or transitions from `UNKNOWN`, and `removed` disappearing files or transitions to `UNKNOWN`.

skisela: And the corresponding text output would be:
```
...
licensecheck
======
License changes occured!
Removed - license removed to unset or file disappeared
Added - license added(previously unset) or new file appeared

removed license(s)
====
* LGPL (v2 or later)
 - platforminputcontext/keyserver_x11.h
* GPL (v2 or later)
 - platforminputcontext/qfcitxplatforminputcontext.cpp
 - platforminputcontext/main.cpp
 - platforminputcontext/qfcitxplatforminputcontext.h
 - platforminputcontext/main.h
 - platforminputcontext/keydata.h
 - platforminputcontext/keyuni.h
 - platforminputcontext/keyuni.cpp
 - platforminputcontext/utils.h
 - platforminputcontext/utils.cpp

added license(s)
====
* BSD (3 clause)
 - COPYING.BSD
* UNKNOWN
 - quickphrase-editor/editor.ui
 - po/ko.po
...
```

nforro: Well, the details should go to a licensecheck report file, only some summary (so maybe just a list of (dis)appearing licenses) should be printed in the general report.

skisela: I am sorry, I meant the `report.txt` text output.

nforro: Yes, that's what I mean by general report. `licensecheck` should have its own report file, just like e.g. `rpmdiff`.

skisela: You actually got me right. Then I would suggest the `changed` licenses not the disappeared onces..?

nforro: From the files point of view you have added, changed and removed licenses. But if you forget about files, and just take all the licenses present in the old sources and the new sources, then you can print only differences between those sets.

rebasehelper/checkers/licensecheck_tool.py
```diff
@@ -0,0 +1,180 @@
...
+            if old_file not in diffs[1]:
+                if old_license not in changes['removed']:
+                    changes['removed'][old_license] = []
+                changes['removed'][old_license].append(old_file)
+
+        new_licenses = []
+        for new in new_lics:
+            if new not in old_lics:
+                cls.license_changes = True
+                new_licenses.append(new)
```
nforro: You should also check whether a license disappeared - that equals to license change too.

skisela: You are right, and it still does not count with packages that are not in a directory after unpacking the tarball.

nforro: I'm not sure I understand, what packages are not in which directory?

rebasehelper/checkers/licensecheck_tool.py
```diff
@@ -0,0 +1,182 @@
...
+        for old_file, old_license in six.iteritems(diffs[0]):
+            if old_file not in diffs[1]:
+                if old_license not in changes['removed']:
+                    changes['removed'][old_license] = []
+                changes['removed'][old_license].append(old_file)
+
+        new_licenses = [lic for lic in new_lics if lic not in old_lics]
+        disappeared_licenses = [lic for lic in old_lics if lic not in new_lics]
+
+        if new_licenses or disappeared_licenses:
```
nforro: Hm, how about taking advantage of sets and using:

```python
if old_lics ^ new_lics:
```
See: https://docs.python.org/3/library/stdtypes.html#frozenset.symmetric_difference

rebasehelper/checkers/licensecheck_tool.py
skisela: While running tests licensecheck fails due to unexpected format of untarred archive `test-1.0.2.tar.xz` which does not contain a subdirectory `test-1.0.2`, just a file `text.txt`.
Licensecheck always tries to cut the first "version" directory off from the file names for more convenient manipulation. 
Most packages are made so, that the tarball contains the version subdirectory.

Do you think we should solve it somehow, or just move the testing file into a versioned subdirectory?

skisela: I will probably solve it by try-catch with appending `../` to the exceptional files.

nforro: Not sure which tests, but this is certainly a bug. The archive is not even being extracted in `%prep` of `test.spec`. `%setup` with `-c` option has to be used in `%prep` or top-level directory has to be added to the archive. And `Application.prepare_sources()` has to be modified to take potential `-c` option into account.

nforro: Ok, I was wrong. `%setup -q -a 5` actually means to extract `Source0` and `Source5`. But the `-c` option or top-level dir in `Source0` archive is still missing. I don't think a package can be built without having top-level dir (created either way).

nforro: So, to conclude this: the test has to be fixed, and source checkers should use paths returned from `Application.prepare_sources()`, which should be modified to take `%prep` into account.

rebasehelper/application.py
```diff
@@ -423,6 +425,8 @@ def prepare_sources(self):
                     if dest_dir:
                         Application.extract_sources(rest, os.path.join(self.execution_dir, sources_dir, dest_dir))
 
+        self.old_dir = old_dir
+        self.new_dir = new_dir
         self.run_package_checkers(self.results_dir, category="SOURCE")
```
nforro: Why don't you just pass `old_dir` and `new_dir` into the method as kwargs?

nforro:
If you want this PR reviewed you should really rebase it to the lastest state of the base branch.

rebasehelper/checkers/licensecheck_tool.py
```diff
@@ -0,0 +1,178 @@
...
+                            changes['changed'][new_key].append(new_file)
+
+        # Get newly appeared files
+        for new_file, new_license in six.iteritems(diffs[1]):
+            if new_file not in diffs[0]:
+                if new_license not in changes['added']:
+                    changes['added'][new_license] = []
+                changes['added'][new_license].append(new_file)
+
+        # Get removed appeared files
```
nforro: This comment is weird ;-)

nforro:
Rebasing `libsamplerate` from 0.1.8 to 0.1.9:

From report.txt:
```
licensecheck
============
License changes occured!
* BSD (2 clause) appeared
* CC0 GENERATED FILE appeared
* GPL appeared
* GPL (with incorrect FSF address) GENERATED FILE appeared
* CC0 GPL (v2 or later) (with incorrect FSF address) GENERATED FILE disappeared
* GPL (v2 or later) disappeared
* GPL (v2 or later) (with incorrect FSF address) GENERATED FILE disappeared
* LGPL (v2.1 or later) disappeared
Detailed output can be found in rebase-helper-results/checkers/licensecheck
```

Why is the output inconsistent with other checkers? I'd expect to see:

```
Details in rebase-helper-results/checkers/licensecheck:
 - report.txt
```

Obviously `GENERATED FILE` has to be ignored. Otherwise it looks fine.
Perhaps license change in LICENSE, COPYING or similar files could be highlighted, because that's the most important change. But that can be done later.

From licensecheck.txt:
```
licensecheck
============
License changes occured!
Removed - license removed to unset or file disappeared
Added - license added(previously unset) or new file appeared

added license(s)
================
* BSD (2 clause)
 - COPYING

changed license(s)
==================
* CC0 GPL (v2 or later) (with incorrect FSF address) GENERATED FILE => CC0 GENERATED FILE
 - Win32/config.h
* GPL (v2 or later) (with incorrect FSF address) GENERATED FILE => GPL (with incorrect FSF address) GENERATED FILE
 - aclocal.m4
* GPL (v2 or later) => GPL
 - M4/libtool.m4
* GPL (v2 or later) => GPL (v2 or later) GENERATED FILE
 - Cfg/ltmain.sh

removed license(s)
==================
* GPL (v2 or later) (with incorrect FSF address)
 - examples/audio_out.c
 - examples/audio_out.h
 - examples/sndfile-resample.c
 - examples/timewarp-file.c
 - examples/varispeed-play.c
 - src/common.h
 - src/fastest_coeffs.h
 - src/high_qual_coeffs.h
 - src/mid_qual_coeffs.h
 - src/samplerate.c
 - src/samplerate.h
 - src/src_linear.c
 - src/src_sinc.c
 - src/src_zoh.c
 - tests/calc_snr.c
 - tests/callback_hang_test.c
 - tests/callback_test.c
 - tests/downsample_test.c
 - tests/float_short_test.c
 - tests/misc_test.c
 - tests/multi_channel_test.c
 - tests/multichan_throughput_test.c
 - tests/reset_test.c
 - tests/simple_test.c
 - tests/snr_bw_test.c
 - tests/src-evaluate.c
 - tests/termination_test.c
 - tests/throughput_test.c
 - tests/util.c
 - tests/util.h
 - tests/varispeed_test.c
* LGPL (v2.1 or later)
 - Win32/unistd.h
 - src/float_cast.h
```

I think this file should look like this:

```
The following licenses appeared:
BSD (2 clause)
 - COPYING

The following licenses disappeared:
GPL (v2 or later) (with incorrect FSF address)
 - examples/audio_out.c
 - examples/audio_out.h
 - examples/sndfile-resample.c
 - examples/timewarp-file.c
 - examples/varispeed-play.c
 - src/common.h
 - src/fastest_coeffs.h
 - src/high_qual_coeffs.h
 - src/mid_qual_coeffs.h
 - src/samplerate.c
 - src/samplerate.h
 - src/src_linear.c
 - src/src_sinc.c
 - src/src_zoh.c
 - tests/calc_snr.c
 - tests/callback_hang_test.c
 - tests/callback_test.c
 - tests/downsample_test.c
 - tests/float_short_test.c
 - tests/misc_test.c
 - tests/multi_channel_test.c
 - tests/multichan_throughput_test.c
 - tests/reset_test.c
 - tests/simple_test.c
 - tests/snr_bw_test.c
 - tests/src-evaluate.c
 - tests/termination_test.c
 - tests/throughput_test.c
 - tests/util.c
 - tests/util.h
 - tests/varispeed_test.c
LGPL (v2.1 or later)
 - Win32/unistd.h
 - src/float_cast.h

The following license transitions occured:
GPL (v2 or later) => GPL
 - M4/libtool.m4
```

nforro:
One more thing, removed files with `UNKNOWN` license shouldn't be listed, as well as added files with `UNKNOWN` license.

nforro:
I mean highlighting it in the report, just like when there are ABI changes.

skisela:
@nforro added by https://github.com/rebase-helper/rebase-helper/pull/439/commits/46e9eb24393986c41a2b00315b532d4ba055edf7

rebasehelper/checkers/licensecheck_tool.py
```diff
@@ -72,7 +72,7 @@ def get_license_changes(cls, old_dir, new_dir):
                 if 'GENERATED FILE' in dlicense:
                     continue
                 file_path = os.path.relpath(file_path, source_dir)
-                if re.search(r'COPYING|LICENSE', file_path):
+                if re.search(r'COPYING|LICENSE', file_path, re.IGNORECASE):
```
nforro: This covers roughly 63% of all variants used in Fedora specs, if you add also `COPYRIGHT`, `LICENCE` and `LICENSING`, it will cover another 10%, which is sufficient for our purposes I think.
There is a number of license files named after the license, so adding `GPL` or `BSD` could be also considered.

rebasehelper/output_tool.py
```diff
@@ -99,6 +96,19 @@ def print_cli_summary(cls, app):
             elif results_store.get_patches()['success']:
                 logger_output.error("\nPatching failed")
 
+    @classmethod
+    def check_for_important_checkers_output(cls):
```
nforro: This doesn't seem right. It should be implemented in respective subclasses (checkers).

rebasehelper/application.py
nforro: The question is still valid. The class variables seem totally unnecessary to me.

rebasehelper/output_tool.py
skisela: Yes you are right. Then I am going to create a separate PR for it and rebase this branch onto it.

rebasehelper/output_tool.py
```diff
@@ -99,6 +96,19 @@ def print_cli_summary(cls, app):
...
+        checkers = results_store.get_checkers()
+        if 'abipkgdiff' in checkers:
+            if AbiCheckerTool.abi_changes:
+                logger_output.warning('\nABI changes occured. Check abipkgdiff output')
+        if 'licensecheck' in checkers:
+            if LicenseCheckTool.license_changes:
+                logger_output.warning('\nLicense changes occured. Check licensecheck output')
+            if LicenseCheckTool.license_files_changes:
```
nforro: I don't think this is necessary, as all the details are in the report.

nforro:
You are referencing `report.txt` in the report but the file is called `licensecheck.txt`.
I think `licensecheck` should be analogic to `rpmdiff` in this regard. So the file should be called `report.txt` and its contents should be just plain text information.

```
licensecheck
============
License changes occured!
Disappeared - license removed to unset or file disappeared
Appeared - license added(previously unset) or new file appeared
```

Especially this section is unnecessary, not to mention that for example `license removed to unset` doesn't make sense at all.

nforro:
Hm, seems something is seriously broken. Rebasing `vim-go` from `1.16` to `1.17` I get:

```
License in LICENSE was modified
```

report.txt:
```
licensecheck
============
No license changes detected.
Details in rebase-helper-results/checkers/licensecheck:
 - report.txt
```

licensecheck.txt:
```
licensecheck
============
License changes occured!
Disappeared - license removed to unset or file disappeared
Appeared - license added(previously unset) or new file appeared
```

skisela:
I've added the mentioned listing of important license files changes.
libsamplerate-0.1.8 -> 0.1.9.
checkers/licensecheck/report.txt
```
licensecheck
============
License changes occured!

The following license files are advised to be inspected
=======================================================
- doc/license.html
- COPYING

The following license(s) appeared
=================================
* BSD (2 clause)
 - COPYING
...
```
I wonder if we are okay with the duplication.

nforro:
No, important changes should be highlighted in the general report, and there should be no duplicities. They should be simply listed first and marked somehow as important.

nforro:
```
licensecheck
============
The following license files are advised to be inspected:
- COPYING
License changes occured!
 - BSD (2 clause) appeared
 - GPL appeared
 - GPL (v2 or later) disappeared
 - LGPL (v2.1 or later) disappeared
Details in rebase-helper-results/checkers/licensecheck:
 - report.txt
```

This looks good (apart from the missing indent before the file), but what I had in mind was to highlight the actual license change, not necessarily showing the filename (but it can be there as well, why not):

```
licensecheck
============
Major license changes:
 - BSD (2 clause) appeared (in COPYING)
Other license changes:
 - GPL appeared
 - GPL (v2 or later) disappeared
 - LGPL (v2.1 or later) disappeared
Details in rebase-helper-results/checkers/licensecheck:
 - report.txt
```

skisela:
Do you think the following is correct?

```
licensecheck
============
Major license changes:
- UNKNOWN => BSD (2 clause) (in COPYING)
Other license changes:
 - BSD (2 clause) appeared
 - GPL appeared
 - GPL (v2 or later) disappeared
 - LGPL (v2.1 or later) disappeared
Details in rebase-helper-results/checkers/licensecheck:
 - report.txt
```
The COPYING file existed in the old sources, so maybe it should be listed as shown?

nforro:
I think that would make it inconsistent with the detailed report.

nforro:
```
Major license changes:
 - BSD (2 clause) (appeared in COPYING)
```

It looks good, only the parentheses are a bit inconsistent with the rest. I would suggest either:
```
BSD (2 clause) appeared (in COPYING)
```
or dropping them entirely.


```
licensecheck
============
No license changes detected.
```

The dot at the end doesn't seem right.

After you fix the styling and linting errors and remove those unnecessary `Application.old_dir` and `Application.new_dir` attributes, I'm going to approve this.

nforro:
It's fine, but I don't get those two commits. The first one won't work by itself because there are undefined variables referenced in `run_package_checkers()`. And I'm not sure if the changes introduced by the second commit are worth separate commit. I'd suggest squashing them to one, or figure out better separation.

PR #444, opened 2018-02-05
Handle ConnectionError when using requests

### LINKED ISSUE #428: Handle ConnectionError in versioneers and DownloadHelper (and in anything else using requests)
```
  File "/home/nforro/devel/rebase-helper/rebasehelper/cli.py", line 174, in run
    app = Application(config, execution_dir, results_dir, debug_log_file)
  File "/home/nforro/devel/rebase-helper/rebasehelper/application.py", line 105, in __init__
    self._prepare_spec_objects()
  File "/home/nforro/devel/rebase-helper/rebasehelper/application.py", line 193, in _prepare_spec_objects
    self.conf.versioneer_blacklist)
  File "/home/nforro/devel/rebase-helper/rebasehelper/versioneer.py", line 92, in run
    result = versioneer.run(package_name)
  File "/home/nforro/devel/rebase-helper/rebasehelper/versioneers/anitya_versioneer.py", line 72, in run
    version = cls._get_version_using_distro_api(package_name)
  File "/home/nforro/devel/rebase-helper/rebasehelper/versioneers/anitya_versioneer.py", line 49, in _get_version    _using_distro_api
    r = requests.get('{}/project/Fedora/{}'.format(cls.API_URL, package_name))
  File "/usr/lib/python2.7/site-packages/requests/api.py", line 70, in get
    return request('get', url, params=params, **kwargs)
  File "/usr/lib/python2.7/site-packages/requests/api.py", line 56, in request
    return session.request(method=method, url=url, **kwargs)
  File "/usr/lib/python2.7/site-packages/requests/sessions.py", line 488, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/lib/python2.7/site-packages/requests/sessions.py", line 609, in send
    r = adapter.send(request, **kwargs)
  File "/usr/lib/python2.7/site-packages/requests/adapters.py", line 487, in send
    raise ConnectionError(e, request=request)
ConnectionError: HTTPSConnectionPool(host='release-monitoring.org', port=443): Max retries exceeded with url: /ap    i/project/Fedora/libtiff (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnec    tion object at 0x7f75261fd8d0>: Failed to establish a new connection: [Errno -2] Name or service not known',))
```

FrNecas:
Resolves #428.

### DIFF
--- a/rebasehelper/spec_hooks/commit_hash_updater.py
+++ b/rebasehelper/spec_hooks/commit_hash_updater.py
@@ -22,10 +22,9 @@
 
 import re
 
-import requests
-
 from rebasehelper.specfile import BaseSpecHook
 from rebasehelper.logger import logger
+from rebasehelper.utils import DownloadHelper
 
 
 class CommitHashUpdaterHook(BaseSpecHook):
@@ -55,7 +54,10 @@ class CommitHashUpdaterHook(BaseSpecHook):
             return None
         baseurl = 'https://api.github.com/repos/{owner}/{project}'.format(**m.groupdict())
         # try to get tag name from a release matching version
-        r = requests.get('{}/releases'.format(baseurl))
+        r = DownloadHelper.request('{}/releases'.format(baseurl))
+        if r is None:
+            return None
+
         if not r.ok:
             if r.status_code == 403 and r.headers.get('X-RateLimit-Remaining') == '0':
                 logger.warning("Rate limit exceeded on Github API! Try again later.")
@@ -67,7 +69,10 @@ class CommitHashUpdaterHook(BaseSpecHook):
             if version in release.get('name'):
                 tag_name = release.get('tag_name')
                 break
-        r = requests.get('{}/tags'.format(baseurl))
+
+        r = DownloadHelper.request('{}/tags'.format(baseurl))
+        if r is None:
+            return None
         if not r.ok:
             if r.status_code == 403 and r.headers.get('X-RateLimit-Remaining') == '0':
                 logger.warning("Rate limit exceeded on Github API! Try again later.")
--- a/rebasehelper/utils.py
+++ b/rebasehelper/utils.py
@@ -288,14 +288,13 @@ class DownloadHelper(object):
         sys.stdout.flush()
 
     @staticmethod
-    def download_file(url, destination_path, blocksize=8192):
+    def request(url, **kwargs):
         """
-        Method for downloading file from HTTP, HTTPS and FTP URL.
+        Method to get a Response object from HTTP, HTTPS or FTP URL
 
-        :param url: URL from which to download the file
-        :param destination_path: path where to store downloaded file
-        :param blocksize: size in Bytes of blocks used for downloading the file and reporting progress
-        :return: None
+        :param url: URL from which the Response is obtained
+        :param kwargs: optional arguments used to get Response object
+        :return: Response object
         """
 
         class FTPAdapter(requests.adapters.BaseAdapter):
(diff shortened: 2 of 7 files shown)


rebasehelper/spec_hooks/commit_hash_updater.py
```diff
@@ -54,8 +54,12 @@ def _get_commit_hash_from_github(cls, spec_file):
         if not m:
             return None
         baseurl = 'https://api.github.com/repos/{owner}/{project}'.format(**m.groupdict())
-        # try to get tag name from a release matching version
-        r = requests.get('{}/releases'.format(baseurl))
+        try:
+            # try to get tag name from a release matching version
+            r = requests.get('{}/releases'.format(baseurl))
+        except requests.exceptions.ConnectionError:
+            logger.debug("Unable to connect to %s", "{}/releases".format(baseurl))
```
nforro: Why don't you just use error message from `ConnectionError`? Is it too generic?

nforro:
Seeing all those code blocks which are all nearly the same, I think it would be best to wrap `requests.get()` into some kind of helper.

rebasehelper/utils.py
```diff
@@ -288,14 +288,13 @@ def format_size(s):
         sys.stdout.flush()
 
     @staticmethod
-    def download_file(url, destination_path, blocksize=8192):
+    def get_response(url, **kwargs):
```
nforro: I think better name for this method would be `request_response`, or simply `request`.

rebasehelper/utils.py
```diff
@@ -323,7 +322,26 @@ def close(self):
         session = requests.Session()
         session.mount('ftp://', FTPAdapter())
 
-        r = session.get(url, stream=True)
+        try:
+            return session.get(url, **kwargs)
+        except requests.exceptions.ConnectionError as e:
+            logger.debug(e)
```
nforro: You should probably prepend the error message with some context, at least the exception name. Also, you are passing exception instance to a method which expects string.

nforro:
Resolves #428.

rebasehelper/utils.py
```diff
@@ -323,7 +322,26 @@ def close(self):
         session = requests.Session()
         session.mount('ftp://', FTPAdapter())
 
-        r = session.get(url, stream=True)
+        try:
+            return session.get(url, **kwargs)
+        except requests.exceptions.ConnectionError as e:
+            logger.debug('ConnectionError: %s', str(e))
```
nforro: It's safer to use `six.text_type()` instead of `str()`, as it returns unicode string on python2.

nforro: At this point you can easily catch other types of exceptions `requests.get()` can raise, so it would be nice to do so.

FrNecas: So I should also catch `Timeout` and `TooManyRedirects`. `HTTPError` is taken care of in other parts of the code by checking the status code.

nforro: Sure, that makes sense.

rebasehelper/utils.py
```diff
@@ -323,7 +322,26 @@ def close(self):
...
+        Method for downloading file from HTTP, HTTPS and FTP URL.
+
+        :param url: URL from which to download the file
+        :param destination_path: path where to store downloaded file
+        :param blocksize: size in Bytes of blocks used for downloading the file and reporting progress
+        :return: None
+        """
+        r = DownloadHelper.request(url, stream=True)
+        if r is None:
+            raise DownloadError
```
nforro: This means rebase-helper will just stop without any message printed on stdout. That's not very good user experience ;-)

FrNecas: Oh, I totally forgot about this part. What would be an ideal message in this case? Maybe something like: `"Unable to connect to %s", %url`?

nforro: I don't think printing URL again is necessary (as it has been printed already few lines above). Something like `Unexpected error occured during download, see ...log`. But perhaps the original error message should be also printed to stdout, not just to debug log.

rebasehelper/utils.py
```diff
@@ -323,7 +322,27 @@ def close(self):
         session = requests.Session()
         session.mount('ftp://', FTPAdapter())
 
-        r = session.get(url, stream=True)
+        try:
+            return session.get(url, **kwargs)
+        except requests.exceptions.RequestException as e:
+            logger.debug('%s: %s', type(e).__name__, six.text_type(e))
+            logger.error('%s: %s', type(e).__name__, six.text_type(e))
```
nforro: Why logging the same message twice?

rebasehelper/utils.py
```diff
@@ -323,7 +322,27 @@ def close(self):
         session = requests.Session()
         session.mount('ftp://', FTPAdapter())
 
-        r = session.get(url, stream=True)
+        try:
+            return session.get(url, **kwargs)
+        except requests.exceptions.RequestException as e:
+            logger.error('%s: %s', type(e).__name__, six.text_type(e))
+            sys.stdout.write('%s: %s' % (type(e).__name__, six.text_type(e)))
```
nforro: Ok, this time it's the same message printed to output twice. You should probably get familiar with how logging works in this project. In short, unless `--verbose` is specified, anything with level [`logging.INFO` or above](https://docs.python.org/3/library/logging.html#logging-levels) will be both printed to console and written to debug log file.

FrNecas: Oh, sorry, I didn't know that. So just logger.error() should be enough to write it to debug log file and inform the user. I will fix it

PR #446, opened 2018-02-08
Refactor build process

nforro:
This PR splits build process into two phases, build of source packages (SRPM) and build of binary packages (RPM).

This is necessary for SRPM checkers to be able to run just after SRPMs are built.
Also now one doesn't have to wait until all old-version RPMs are built only to find out there is some trivial error in rebased SPEC file and new-version SRPM build fails because of that.

### DIFF
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -60,7 +60,6 @@ class Application(object):
     debug_log_file = None
     report_log_file = None
     rebased_patches = {}
-    upstream_monitoring = False
     rebased_repo = None
 
     def __init__(self, cli_conf, execution_dir, results_dir, debug_log_file):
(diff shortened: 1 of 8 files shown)


rebasehelper/application.py
```diff
@@ -482,73 +481,113 @@ def _prepare_rebased_repository(cls, patches, rebased_sources_dir):
...
+                srpm_builder_options=self.conf.srpm_builder_options)
+            try:
+                build_dict.update(builder.build(spec, results_dir, **build_dict))
+                build_dict = self._sanitize_build_dict(build_dict)
+                results_store.set_build_data(version, build_dict)
+            except RebaseHelperError:
+                raise
+            except SourcePackageBuildError as e:
+                build_dict.update(builder.get_logs())
+                build_dict['source_package_build_error'] = six.text_type(e)
```
skisela: If I am not wrong, we are not storing errors from binary builds in this way. Can you explain?

nforro: Actually I haven't touched this, just moved it from the original `build_packages()` method.
It certainly is possible that it's useless. But I plan to rework all the `ResultsStore` related things anyway.

PR #454, opened 2018-02-17
Add haskell versioneer

FrNecas:


rebasehelper/versioneers/haskell_versioneer.py
```diff
@@ -0,0 +1,64 @@
...
+    @classmethod
+    def get_categories(cls):
+        return cls.CATEGORIES
+
+    @classmethod
+    def _get_version(cls, package_name):
+        if package_name.startswith('ghc-'):
+            package_name = package_name.replace('ghc-', '', 1)
+
+        # Accept header is used to get JSON from Hackage API
```
nforro: I think it's pretty self-explanatory, but if you think it's comment-worthy, then keep it.

FrNecas: You are right, I will remove it.

PR #457, opened 2018-03-01
Enhance console logging

FrNecas:
Colorize console output using ColorizingStreamHandler.
Also, use formatter to display logging level in the output.

There are now 2 color maps, they can be either switched manually or rebase-helper can decide which one to use depending on the background color of the terminal.

4 new logging levels have been added 
- SUCCESS for successful operations (colored to green)
- HEADING for separating individual sections of output ( colored to blue/yellow)
- IMPORTANT to highlight important messages (colored to red)
- TRACE to differentiate between traceback and debug messages (colored to red)

### DIFF
--- a/rebasehelper/cli.py
+++ b/rebasehelper/cli.py
@@ -173,18 +173,18 @@ class CliHelper(object):
             app = Application(config, execution_dir, results_dir, debug_log_file)
             app.run()
         except KeyboardInterrupt:
-            logger.info('\nInterrupted by user')
+            logger.info('Interrupted by user')
         except RebaseHelperError as e:
             if e.msg:
-                logger.error('\n%s', e.msg)
+                logger.error('%s', e.msg)
             else:
-                logger.error('\n%s', six.text_type(e))
+                logger.error('%s', six.text_type(e))
             sys.exit(1)
         except SystemExit as e:
             sys.exit(e.code)
         except BaseException:
             if debug_log_file:
-                logger.error('\nrebase-helper failed due to an unexpected error. Please report this problem'
+                logger.error('rebase-helper failed due to an unexpected error. Please report this problem'
                              '\nusing the following link: %s'
                              '\nand include the content of'
                              '\n\'%s\''
@@ -192,12 +192,12 @@ class CliHelper(object):
                              '\nThank you!',
                              NEW_ISSUE_LINK, debug_log_file)
             else:
-                logger.error('\nrebase-helper failed due to an unexpected error. Please report this problem'
+                logger.error('rebase-helper failed due to an unexpected error. Please report this problem'
                              '\nusing the following link: %s'
                              '\nand include the traceback following this message in the report.'
                              '\nThank you!',
                              NEW_ISSUE_LINK)
-            logger.trace('\n', exc_info=1)
+            logger.trace('', exc_info=1)
             sys.exit(1)
 
         sys.exit(0)
--- a/rebasehelper/cli.py
+++ b/rebasehelper/cli.py
@@ -167,7 +167,7 @@ class CliHelper(object):
             config = Config(getattr(cli, 'config-file', None))
             config.merge(cli)
             for handler in [main_handler, output_tool_handler]:
-                handler.set_terminal_background()
+                handler.set_terminal_background(config.background)
 
             ConsoleHelper.use_colors = ConsoleHelper.should_use_colors(config)
             execution_dir, results_dir, debug_log_file = Application.setup(config)
--- a/rebasehelper/cli.py
+++ b/rebasehelper/cli.py
@@ -197,7 +197,7 @@ class CliHelper(object):
                              '\nand include the traceback following this message in the report.'
                              '\nThank you!',
                              NEW_ISSUE_LINK)
-            logger.debug('\n', exc_info=1)
+            logger.trace('\n', exc_info=1)
             sys.exit(1)
 
         sys.exit(0)
--- a/rebasehelper/cli.py
+++ b/rebasehelper/cli.py
@@ -30,7 +30,7 @@ from rebasehelper.options import OPTIONS, traverse_options
 from rebasehelper.constants import PROGRAM_DESCRIPTION, NEW_ISSUE_LINK
 from rebasehelper.version import VERSION
 from rebasehelper.application import Application
-from rebasehelper.logger import logger, handler
+from rebasehelper.logger import logger, main_handler, output_tool_handler
 from rebasehelper.exceptions import RebaseHelperError
 from rebasehelper.utils import ConsoleHelper
 from rebasehelper.config import Config
@@ -166,10 +166,13 @@ class CliHelper(object):
 
             config = Config(getattr(cli, 'config-file', None))
             config.merge(cli)
+            for handler in [main_handler, output_tool_handler]:
+                handler.set_terminal_background()
+
             ConsoleHelper.use_colors = ConsoleHelper.should_use_colors(config)
             execution_dir, results_dir, debug_log_file = Application.setup(config)
             if not config.verbose:
-                handler.setLevel(logging.INFO)
+                main_handler.setLevel(logging.INFO)
             app = Application(config, execution_dir, results_dir, debug_log_file)
             app.run()
         except KeyboardInterrupt:
--- a/rebasehelper/cli.py
+++ b/rebasehelper/cli.py
@@ -30,7 +30,7 @@ from rebasehelper.options import OPTIONS, traverse_options
 from rebasehelper.constants import PROGRAM_DESCRIPTION, NEW_ISSUE_LINK
 from rebasehelper.version import VERSION
 from rebasehelper.application import Application
-from rebasehelper.logger import logger, LoggerHelper
+from rebasehelper.logger import logger, handler
 from rebasehelper.exceptions import RebaseHelperError
 from rebasehelper.utils import ConsoleHelper
 from rebasehelper.config import Config
@@ -159,7 +159,6 @@ class CliHelper(object):
         debug_log_file = None
         try:
             # be verbose until debug_log_file is created
-            handler = LoggerHelper.add_stream_handler(logger, logging.DEBUG)
             cli = CLI()
             if hasattr(cli, 'version'):
                 logger.info(VERSION)
--- a/rebasehelper/logger.py
+++ b/rebasehelper/logger.py
@@ -125,22 +125,41 @@ class LoggerHelper(object):
 
 
 class ColorizingStreamHandler(logging.StreamHandler):
-    level_map = {
-        logging.DEBUG: {'fg': 'brightblack', 'bg': 'default', 'style': None},
-        CustomLogger.TRACE: {'fg': 'red', 'bg': 'default', 'style': None},
-        logging.INFO: {'fg': 'default', 'bg': 'default', 'style': None},
-        CustomLogger.SUCCESS: {'fg': 'green', 'bg': 'default', 'style': None},
-        CustomLogger.HEADING: {'fg': 'yellow', 'bg': 'default', 'style': None},
-        CustomLogger.IMPORTANT: {'fg': 'red', 'bg': 'default', 'style': None},
-        logging.WARNING: {'fg': 'yellow', 'bg': 'default', 'style': None},
-        logging.ERROR: {'fg': 'red', 'bg': 'default', 'style': 'bold'},
-        logging.CRITICAL: {'fg': 'white', 'bg': 'red', 'style': 'bold'},
+    colors = {
+        'dark': {
+            logging.DEBUG: {'fg': 'brightblack', 'bg': 'default', 'style': None},
+            CustomLogger.TRACE: {'fg': 'red', 'bg': 'default', 'style': None},
+            logging.INFO: {'fg': 'default', 'bg': 'default', 'style': None},
+            CustomLogger.SUCCESS: {'fg': 'green', 'bg': 'default', 'style': None},
+            CustomLogger.HEADING: {'fg': 'yellow', 'bg': 'default', 'style': None},
+            CustomLogger.IMPORTANT: {'fg': 'red', 'bg': 'default', 'style': None},
+            logging.WARNING: {'fg': 'yellow', 'bg': 'default', 'style': None},
+            logging.ERROR: {'fg': 'red', 'bg': 'default', 'style': 'bold'},
+            logging.CRITICAL: {'fg': 'white', 'bg': 'red', 'style': 'bold'},
+        },
+        'light': {
+            logging.DEBUG: {'fg': 'brightblack', 'bg': 'default', 'style': None},
+            CustomLogger.TRACE: {'fg': 'red', 'bg': 'default', 'style': None},
+            logging.INFO: {'fg': 'default', 'bg': 'default', 'style': None},
+            CustomLogger.SUCCESS: {'fg': 'green', 'bg': 'default', 'style': None},
+            CustomLogger.HEADING: {'fg': 'blue', 'bg': 'default', 'style': None},
+            CustomLogger.IMPORTANT: {'fg': 'red', 'bg': 'default', 'style': None},
+            logging.WARNING: {'fg': 'blue', 'bg': 'default', 'style': None},
+            logging.ERROR: {'fg': 'red', 'bg': 'default', 'style': 'bold'},
+            logging.CRITICAL: {'fg': 'white', 'bg': 'red', 'style': 'bold'},
+        },
     }
+
+    terminal_background = 'dark'
+
+    def set_terminal_background(self):
+        self.terminal_background = rebasehelper.utils.ConsoleHelper.detect_background()
 
     def emit(self, record):
         try:
             message = self.format(record)
-            rebasehelper.utils.ConsoleHelper.cprint(message, **self.level_map.get(record.levelno, {}))
+            level_settings = self.colors[self.terminal_background].get(record.levelno, {})
+            rebasehelper.utils.ConsoleHelper.cprint(message, **level_settings)
             self.flush()
         except Exception:  # pylint: disable=broad-except
             self.handleError(record)
@@ -153,6 +172,6 @@ logger = LoggerHelper.get_basic_logger('rebase-helper')
 logger_output = LoggerHelper.get_basic_logger('output-tool', logging.INFO)
 logger_report = LoggerHelper.get_basic_logger('rebase-helper-report', logging.INFO)
 logger_upstream = LoggerHelper.get_basic_logger('rebase-helper-upstream')
-LoggerHelper.add_stream_handler(logger_output)
+output_tool_handler = LoggerHelper.add_stream_handler(logger_output)
 formatter = logging.Formatter("%(levelname)s: %(message)s")
-handler = LoggerHelper.add_stream_handler(logger, logging.DEBUG, formatter)
+main_handler = LoggerHelper.add_stream_handler(logger, logging.DEBUG, formatter)
--- a/rebasehelper/logger.py
+++ b/rebasehelper/logger.py
@@ -21,8 +21,40 @@
 #          Tomas Hozza <thozza@redhat.com>
 
 import logging
+import six
 
 import rebasehelper.utils
+
+
+class CustomLogger(logging.Logger):
+
+    SUCCESS = logging.INFO + 5
+    HEADING = logging.INFO + 6
+    IMPORTANT = logging.INFO + 7
+
+    _nameToLevel = {
+        'SUCCESS': SUCCESS,
+        'HEADING': HEADING,
+        'IMPORTANT': IMPORTANT,
+    }
+
+    def __init__(self, name, level=logging.NOTSET):
+        super(CustomLogger, self).__init__(name, level)
+
+        for lev, severity in six.iteritems(self._nameToLevel):
+            logging.addLevelName(severity, lev)
+
+    def __getattr__(self, level):
+        severity = self._nameToLevel.get(level.upper())
+
+        def log(message, *args, **kwargs):
+            if self.isEnabledFor(severity):
+                self._log(severity, message, args, **kwargs)
+
+        if severity:
+            return log
+
+        raise AttributeError
 
 
 class LoggerHelper(object):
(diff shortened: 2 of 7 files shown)


rebasehelper/logger.py
```diff
@@ -73,11 +78,29 @@ def add_file_handler(logger_object, path, formatter_object=None, level=None):
...
+        logging.INFO: {'fg': 'default', 'bg': 'default', 'style': None},
+        logging.WARNING: {'fg': 'yellow', 'bg': 'default', 'style': None},
+        logging.ERROR: {'fg': 'red', 'bg': 'default', 'style': 'bold'},
+        logging.CRITICAL: {'fg': 'white', 'bg': 'red', 'style': 'bold'},
+    }
+
+    def emit(self, record):
+        try:
+            message = self.format(record)
+            rebasehelper.utils.ConsoleHelper.cprint(message, **self.level_map.get(record.levelno, None))
```
nforro: No need for `None` in `get()` method, as that's the default. But you should rather use `{}` instead, because `**None` leads to `TypeError`.

rebasehelper/utils.py
```diff
@@ -101,7 +101,15 @@ class ConsoleHelper(object):
 
     @classmethod
     def should_use_colors(cls, conf):
-        """Determine whether ansi colors should be used for CLI output"""
+        """Determine whether ansi colors should be used for CLI output
```
nforro: This shouldn't be in imperative mode. Also, ANSI is an abbreviation, so it should be capitalized.

nforro: And it should end with a `.`.

rebasehelper/utils.py
```diff
@@ -101,7 +101,15 @@ class ConsoleHelper(object):
...
+
+        Args:
+            conf: configuration object with arguments from the command line
```
nforro: There is no type (should be `rebasehelper.config.Config`), and the description should start with capital `C` and end with a `.`.

rebasehelper/utils.py
```diff
@@ -101,7 +101,15 @@ class ConsoleHelper(object):
...
+
+        Return:
```
nforro: `Returns`, not `Return`.

rebasehelper/utils.py
```diff
@@ -101,7 +101,15 @@ class ConsoleHelper(object):
...
+            True or False, depends on whether colors should be used
```
nforro: Again, missing type (`bool`), and I would write it simply as: `bool: Whether colors should be used.`.

nforro: Take a look on docstrings in [`KojiHelper`](https://github.com/rebase-helper/rebase-helper/blob/master/rebasehelper/utils.py#L912), and official [google doc](http://google.github.io/styleguide/pyguide.html?showone=Comments#Comments).

rebasehelper/logger.py
```diff
@@ -77,11 +77,39 @@ def add_file_handler(logger_object, path, formatter_object=None, level=None):
...
+        if hasattr(logging.getLoggerClass(), method_name):
+            raise AttributeError('{} already defined in logger class'.format(method_name))
+
+        def log_level(self, message, *args, **kwargs):
+            if self.isEnabledFor(level_number):
+                self._log(level_number, message, args, **kwargs)  # pylint: disable=protected-access
+
+        logging.addLevelName(level_number, level_name)
+        setattr(logging, level_name, level_number)
+        setattr(logging.getLoggerClass(), method_name, log_level)
```
nforro: I don't like this at all. Why won't you define a custom class based on `Logger` and register it using `setLoggerClass()`?

rebasehelper/output_tool.py
```diff
@@ -115,9 +114,12 @@ def print_patches_section_cli(cls, color, patch_type):
             return
 
         if patch_type in patches:
-            print('\n%s patches:' % patch_type)
+            logger_output.info('\n%s patches:', patch_type)
             for patch in patches[patch_type]:
-                ConsoleHelper.cprint(patch, color=color)
+                if color == 'green':
```
nforro: No, this is wrong. `BaseOutputTool.print_patches_cli()` should be changed to map to logging levels instead of colors.

rebasehelper/logger.py
```diff
@@ -117,10 +117,30 @@ class ColorizingStreamHandler(logging.StreamHandler):
         logging.CRITICAL: {'fg': 'white', 'bg': 'red', 'style': 'bold'},
     }
 
+    bright_level_map = {
```
nforro: I suspect you wanted to use `light` instead of `bright`.

rebasehelper/logger.py
```diff
@@ -117,10 +117,30 @@ class ColorizingStreamHandler(logging.StreamHandler):
...
+
+    def choose_color_map(self):
+        self.map_used = rebasehelper.utils.ConsoleHelper.dark_or_bright_color_scheme()
+
     def emit(self, record):
         try:
             message = self.format(record)
-            rebasehelper.utils.ConsoleHelper.cprint(message, **self.level_map.get(record.levelno, None))
+            level_settings = self.dark_level_map.get(record.levelno, None) if self.map_used == 'dark' \
+                else self.bright_level_map.get(record.levelno, None)
```
nforro: Oneliner split into multiple lines shouldn't be oneliner.

nforro: Besides, the maps could be joined together and color scheme selected like: `colors[map_used][record_levelno]`.

rebasehelper/utils.py
```diff
@@ -138,6 +142,83 @@ def cprint(cls, message, fg=None, bg=None, style=None):
...
+            dark or bright depending on the background color
+
+        """
+        background_color = ConsoleHelper.exchange_control_sequence('\x1b]11;?\x07')
+        # use black color by default if the terminal doesn't support the control sequence
+        if background_color is None:
+            return 'dark'
+
+        # parse RGB from the return code, for example from 'rgb:0000/0000/0000' for black
+        background_color = background_color.replace('rgb:', '', 1)
```
nforro: This should really be parsed with a regexp. You should keep in mind that the output doesn't always have to have the `RGB Device specification` form.

rebasehelper/utils.py
```diff
@@ -138,6 +142,83 @@ def cprint(cls, message, fg=None, bg=None, style=None):
...
+        rgb = ''.join([x[0:2] for x in background_color.split('/')])
```
nforro: I don't get it. You are for some reason just ignoring half of 16-bit color components (which could also be 4-bit, 8-bit or 12-bit btw, see [`RGB Device specification`](https://tronche.com/gui/x/xlib/color/strings/rgb.html)), joining it together and then splitting it again in `color_is_light()` method. Why?

nforro: You should write a function to parse `RGB Device specification` and return RGB tuple, without limiting precision of components.

rebasehelper/utils.py
```diff
@@ -138,6 +142,83 @@ def cprint(cls, message, fg=None, bg=None, style=None):
...
+        """Get terminal background color and decide whether to use dark or bright color scheme
+
+        Return:
+            dark or bright depending on the background color
+
+        """
+        background_color = ConsoleHelper.exchange_control_sequence('\x1b]11;?\x07')
+        # use black color by default if the terminal doesn't support the control sequence
+        if background_color is None:
+            return 'dark'
```
nforro: This could be written as:

```python
background_color = ConsoleHelper.exchange_control_sequence('\x1b]11;?\x07') or 'dark'
```

rebasehelper/options.py
```diff
@@ -55,6 +55,12 @@
         "name": ["--results-dir"],
         "help": "directory where rebase-helper output will be stored",
     },
+    {
+        "name": ["--bright-color-scheme"],
```
nforro: I would prefer `--background {dark,light,auto}`, default being `auto`, like in vim for example.

rebasehelper/cli.py
```diff
@@ -159,8 +159,11 @@ def run():
         debug_log_file = None
         try:
             # be verbose until debug_log_file is created
-            handler = LoggerHelper.add_stream_handler(logger, logging.DEBUG, formatter)
-            handler.choose_color_map()
+            main_handler = LoggerHelper.add_stream_handler(logger, logging.DEBUG, formatter)
+            output_tool_handler = LoggerHelper.add_stream_handler(logger_output)
```
nforro: All these handlers and loggers are a real mess. There should be just different loggers importable from `rebasehelper.logger` with methods for setting log level and color map.

skisela:
Updates of python-ansicolors in Fedora:
f27: https://bodhi.fedoraproject.org/updates/FEDORA-2018-7955650a80
f28: https://bodhi.fedoraproject.org/updates/FEDORA-2018-cce2d29a5f

nforro:
> After these are pushed to stable Fedora, this can be dropped:
> https://github.com/rebase-helper/rebase-helper/pull/457/files#diff2eeaed663bd0d25b7e608891384b7298R129

Actually, it can't. Not until the change is merged upstream and propagated to PyPI.

rebasehelper/logger.py
```diff
@@ -30,47 +32,79 @@ class LoggerHelper(object):
 
     @staticmethod
     def get_basic_logger(logger_name, level=logging.DEBUG):
-        """
-        Sets-up a basic logger without any handler
+        """Sets up a basic logger without any handler.
+
+        Args:
+            logger_name (str): Logger name.
+            level (int): Severity level set to the logger.
```
nforro: Just a nitpick, but there should be a newline here.

nforro: `Severity level set to the logger` - that doesn't make much sense. How about `Severity level associated with the logger`?

rebasehelper/logger.py
```diff
@@ -30,47 +32,79 @@ class LoggerHelper(object):
...
+        Returns:
+            logging.Logger: Created logger.
```
nforro: `Created logger instance` would be a bit better IMHO.

rebasehelper/logger.py
```diff
@@ -30,47 +32,79 @@ class LoggerHelper(object):
...
         basic_logger = logging.getLogger(logger_name)
         basic_logger.setLevel(level)
         return basic_logger
 
     @staticmethod
-    def add_stream_handler(logger_object, level=None):
-        """
-        Adds console handler with given severity.
+    def add_stream_handler(logger_object, level=None, formatter_object=None):
+        """Adds console handler with given severity.
```
nforro: Adds it where? Also why `console handler` when the method is called `add_stream_handler`?

rebasehelper/logger.py
```diff
@@ -30,47 +32,79 @@ class LoggerHelper(object):
...
+        if formatter_object:
+            console_handler.setFormatter(formatter_object)
         logger_object.addHandler(console_handler)
         return console_handler
 
     @staticmethod
     def add_file_handler(logger_object, path, formatter_object=None, level=None):
-        """
-        Adds FileHandler to a given logger
+        """Adds file handler to a given logger.
```
nforro: I'm pretty sure there should be `the` instead of `a` if you are using `given`.

rebasehelper/logger.py
```diff
@@ -30,47 +32,79 @@ class LoggerHelper(object):
...
+
+        Args:
+            logger_object (logging.Logger): Logger object to add the handler to
```
nforro: Missing dot.

rebasehelper/logger.py
```diff
@@ -30,47 +32,79 @@ class LoggerHelper(object):
...
+            path (str): Path to file where the log will be written.
```
nforro: `Path to a log file` would be sufficient I think.

rebasehelper/utils.py
```diff
@@ -115,27 +123,33 @@ def should_use_colors(cls, conf):
...
-    def cprint(cls, message, color=None):
-        """
-        Print colored output if possible
+    def cprint(cls, message, fg=None, bg=None, style=None):
+        """Prints colored output if possible.
+
+        Args:
+            message (str): String to be printed out.
+            fg (str): Color to be used as a foreground in the printed message.
+            bg (str): Color to be used as a background in the printed message.
```
nforro: This is unnecessarily verbose, just `foreground color`/`background color` is enough.

rebasehelper/utils.py
```diff
@@ -115,27 +123,33 @@ def should_use_colors(cls, conf):
...
+            style (str): Style to be applied to the printed message.
```
nforro: This is not very useful. Style of what? What are the possibilities?

rebasehelper/utils.py
```diff
@@ -115,27 +123,33 @@ def should_use_colors(cls, conf):
...
+        if cls.use_colors:
+            print(colors.color(message, fg=fg, bg=bg, style=style))
         else:
             print(message)
 
     @staticmethod
     def get_message(message, default_yes=True, any_input=False):
-        """
-        Function for command line messages
+        """Prompts a user with yes/no message and gets the response
```
nforro: `an user`.

nforro: And missing dot at the end.

rebasehelper/utils.py
```diff
@@ -115,27 +123,33 @@ def should_use_colors(cls, conf):
...
+
+        Args:
+            message (str): Prompt string.
+            default_yes (bool): If the default value is YES.
+            any_input (bool): If True, return input without checking it first.
+
+        Returns:
+            bool: True or False, based on user's input
```
nforro: Missing dot.

rebasehelper/utils.py
```diff
@@ -115,27 +123,33 @@ def should_use_colors(cls, conf):
...
 
     @staticmethod
     def get_message(message, default_yes=True, any_input=False):
-        """
-        Function for command line messages
+        """Prompts a user with yes/no message and gets the response
+
+        Args:
+            message (str): Prompt string.
+            default_yes (bool): If the default value is YES.
```
nforro: I think it would be better to write `If the default value should be YES.`

rebasehelper/utils.py
```diff
@@ -115,27 +123,33 @@ def should_use_colors(cls, conf):
...
+            any_input (bool): If True, return input without checking it first.
```
nforro: This should be: `Whether to return default value regardless of input.`

rebasehelper/utils.py
FrNecas: As far as I know it depends on the pronunciation in this case, therefore it should be a user (https://ell.stackexchange.com/a/26988)

rebasehelper/logger.py
```diff
@@ -21,10 +21,47 @@
...
+    'HEADING': HEADING,
+    'IMPORTANT': IMPORTANT,
+}
+
+
+class CustomLogger(logging.Logger):
+
+    SUCCESS = SUCCESS
+    HEADING = HEADING
+    IMPORTANT = IMPORTANT
```
nforro: What's this? ;-) Why don't you just move the constants and `_nameToLevel` dict here?

rebasehelper/tests/test_utils.py
```diff
@@ -150,6 +150,30 @@ def write():
         assert capturer.stdout == 'test stdout'
         assert capturer.stderr == 'test stderr'
 
+    @pytest.mark.parametrize('specification, expected_result', [
+        ('rgb:0000/0000/0000', (0, 0, 0, 4)),
+        ('rgb:ffff/ffff/ffff', (65535, 65535, 65535, 4)),
```
nforro: I think using hexadecimal constants would be more appropriate. Same for the other test.

rebasehelper/utils.py
nforro: Never realized that, thanks. I guess I've seen `an user` written so many times that it didn't even occur to me that it could be wrong.

rebasehelper/utils.py
```diff
@@ -39,6 +39,10 @@
 import pyquery
 import hashlib
 import requests
+import termios
+import fcntl
+import tty
+import datetime
```
nforro: The imports are all wrong. Since you are adding a lot of new ones, you could as well fix that. Third party imports are mixed with standard library ones and the order is not alphabetical, which is a pain considering the number of them.

rebasehelper/utils.py
```diff
@@ -138,6 +142,101 @@ def cprint(cls, message, fg=None, bg=None, style=None):
         else:
             print(message)
 
+    @staticmethod
+    def color_is_light(rgb):
+        """Function to determine whether a color is dark or light.
```
nforro: `Function` is redundant here (and it's actually a method).

rebasehelper/utils.py
```diff
@@ -138,6 +142,101 @@ def cprint(cls, message, fg=None, bg=None, style=None):
...
+    def dark_or_light_background():
+        """Gets terminal background color and decides whether it is light or dark.
+
+        Returns:
+            str: Whether to use dark or light color scheme.
+
+        """
+        background_color = ConsoleHelper.exchange_control_sequence('\x1b]11;?\x07')
+
+        if not re.match(r'^rgb:([A-Fa-f0-9]{1,4})/([A-Fa-f0-9]{1,4})/([A-Fa-f0-9]{1,4})$', str(background_color)):
```
nforro: WTF is this? First you use regexp to parse the specification, then you just forget about it and use a different (and much less robust) parsing method.

rebasehelper/utils.py
```diff
@@ -138,6 +142,101 @@ def cprint(cls, message, fg=None, bg=None, style=None):
...
+        Args:
+            specification (str): RGB device specification.
+
+        Returns:
+            tuple: RGB tuple with the last element being number of hex digits it was parsed from.
+
+        """
+        specification = re.sub(r'rgb:', '', specification)
+        rgb = specification.split('/')
+        return tuple(int(x, 16) for x in rgb) + (len(rgb[0]),)
```
nforro: It would be more logical to return a RGB tuple and a bit-width, not one mixed tuple.

rebasehelper/utils.py
```diff
@@ -138,6 +142,101 @@ def cprint(cls, message, fg=None, bg=None, style=None):
...
+        """Function to determine whether a color is dark or light.
+
+        Args:
+            rgb (tuple): RGB tuple with last element specifying number of RGB digits.
+
+        Returns:
+            bool: Whether color is light or dark.
+
+        """
+        brightness = 1 - (0.299 * rgb[0] + 0.587 * rgb[1] + 0.114 * rgb[2]) / (16 ** rgb[3] - 1)
```
nforro: Why are you subtracting one?

FrNecas: When calculating perceptive luminiscence, you are using the highest value which the bit-width can hold. For example for 2-digit RGB it would be ff = 255 so you have to divide by 255 whereas 16 ** 2 = 256. So you have to subtract if I am not mistaken

rebasehelper/utils.py
```diff
@@ -141,6 +145,103 @@ def cprint(cls, message, fg=None, bg=None, style=None):
...
+        Returns:
+            tuple: If the specification follows correct format, the first element is RGB tuple and the second is
+            bit width of the RGB. Otherwise, both elements are None.
+
+        """
+        match = re.match(r'^rgb:([A-Fa-f0-9]{1,4})/([A-Fa-f0-9]{1,4})/([A-Fa-f0-9]{1,4})$', str(specification))
+        if match:
+            rgb = match.groups()
+            bit_width = max([len(x) for x in rgb]) * 4
+            return rgb, bit_width
```
nforro: It would make more sense to return the values as numbers. Otherwise you don't need to return bit-width, it's deducible from the strings.

rebasehelper/utils.py
```diff
@@ -141,6 +145,103 @@ def cprint(cls, message, fg=None, bg=None, style=None):
         else:
             print(message)
 
+    @staticmethod
+    def parse_rgb_device_specification(specification):
+        """Parses RGB device specification.
```
nforro: Missing newline here.

rebasehelper/utils.py
```diff
@@ -21,26 +21,30 @@
...
+
 import copr
+import git
 import pyquery
-import hashlib
 import requests
-
-import git
+import rpm
 import six
```
nforro: There should be one more empty line here.

rebasehelper/logger.py
```diff
@@ -155,6 +155,10 @@ class ColorizingStreamHandler(logging.StreamHandler):
     def choose_color_map(self):
         self.terminal_background = rebasehelper.utils.ConsoleHelper.detect_background()
 
+    def set_color_map(self, color_map):
+        color_map = self.terminal_background if color_map == 'auto' else color_map
```
nforro: This is not such a big deal, but it would be better if no autodetection was performed in case `color_map` is not auto. It could prevent failures in case something unexpected happens during autodetection.

rebasehelper/utils.py
nforro: Right, that makes sense. Thanks.

rebasehelper/utils.py
```diff
@@ -141,6 +146,103 @@ def cprint(cls, message, fg=None, bg=None, style=None):
...
+        Returns:
+            tuple: If the specification follows correct format, the first element is RGB tuple and the second is
+            bit width of the RGB. Otherwise, both elements are None.
+
+        """
+        match = re.match(r'^rgb:([A-Fa-f0-9]{1,4})/([A-Fa-f0-9]{1,4})/([A-Fa-f0-9]{1,4})$', str(specification))
+        if match:
+            rgb = match.groups()
+            bit_width = max([len(str(x)) for x in rgb]) * 4
+            return tuple([int(x, 16) for x in rgb]), bit_width
```
nforro: I don't think it's necessary to convert the list to a tuple, but if you want to do that, you can write just `tuple(int(x, 16) for x in rgb)`.

rebasehelper/tests/test_utils.py
```diff
@@ -150,6 +150,34 @@ def write():
         assert capturer.stdout == 'test stdout'
         assert capturer.stderr == 'test stderr'
 
+    @pytest.mark.parametrize('specification, expected_rgb, expected_bit_width', [
+        ('rgb:0000/0000/0000', (0, 0, 0), 16),
+        ('rgb:ffff/ffff/ffff', (65535, 65535, 65535), 16),
```
nforro: I don't understand why you still write numeric constants in decimal form instead of hexadecimal.

rebasehelper/cli.py
```diff
@@ -167,7 +167,10 @@ def run():
             config = Config(getattr(cli, 'config-file', None))
             config.merge(cli)
             for handler in [main_handler, output_tool_handler]:
-                handler.choose_color_map()
+                if config.background == 'auto':
+                    handler.choose_color_map()
```
nforro: This method seems redundant, why not simply pass `config.background` to `handler.set_color_map()` and handle `auto` there?

Issue #462, opened 2018-03-07
RFE: complete automation


TomasTomecek:
Thank you for making rebase-helper awesome. I absolutely love that tool and it saves me so much time.

I have an idea how to make it even better (sorry if there is an issue for that already).

What if rebase-helper would cover **every** manual task a maintainer needs to do?

Here is how I use rebase-helper now:

1. I run `rebase-helper x.y.z`.
2. If the build succeeds, I copy the rebased sources to the dist-git repo.
3. I run `fedpkg new-sources`.
4. I create a new commit.
5. I push the changes.
6. I build.

What I would love to see rebase-helper doing is:

1. I run `rebase-helper x.y.z`

And it performs all the things mentioned in points 2 up to 6.

skisela:
@TomasTomecek Thanks for feedback and ideas!

You are right at the time all the steps are needed, except `step 2` might be exchanged by applying the generated `changes.patch` in `rebase-helper-results` directory.

Your idea is something we are trying to achieve. To be more exact, we were thinking of using the generated changes to create a pull request, which would trigger CI tests (if any are present) and then be accepted by the maintainer.

The same approach will likely be used for automatic rebases, which are momentarily disabled.

I think that would be ideal, as in case of modifying patches during the rebase, some bugs might be introduced, which should be revealed immediately after.

TomasTomecek:
@skisela can I do the 2nd step with `--results-dir $PWD`?

nforro:
@TomasTomecek: Not really, since `rebase-helper-results` is always appended to the path specified, so `--results-dir $PWD` is actually the default (`$PWD/rebase-helper-results`). Also the rebased files are located in another subdirectory (`$PWD/rebase-helper-results/rebased-sources`). But it shouldn't be a problem to implement an option to change this path.

TomasTomecek:
ah, I see

I'm probably going to create a script which will do exactly what my original post says and as you start implementing some of those, I can update it over time

skisela:
@TomasTomecek do I understand your intention correctly, that you want the files in `$PWD` be in a shape ready for a commit(push) right away?

As I mentioned in https://github.com/rebase-helper/rebase-helper/issues/462#issuecomment-371121197 maybe a little unclearly, the changes in dist-git files are stored in `rebase-helper-results/changes.patch`, therefore what I believe you want to do, is right after `rebase-helper` finishes call `git apply rebase-helper-results/changes.patch`.

[...]

PR #475, opened 2018-04-05
 Add CLI option for applying changes patch

skisela:
After a successful rebase, there is now a possibility to apply the `changes.patch` as a commit to the `dist-git`.

Implements one of the steps towards complete automation mentioned in #462.

### DIFF
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -777,6 +777,20 @@ class Application(object):
         rh_stuff['logs'] = self.get_all_log_files()
         return rh_stuff
 
+    def apply_changes(self):
+        try:
+            repo = git.Repo(self.execution_dir)
+        except git.InvalidGitRepositoryError:
+            repo = git.Repo.init(self.execution_dir)
+        patch = results_store.get_changes_patch()
+        if not patch:
+            logger.warning('Cannot apply changes.patch. No patch file was created')
+        try:
+            repo.git.am(patch['changes_patch'])
+        except git.GitCommandError as e:
+            logger.warning('changes.patch was not applied properly. Please review changes manually.'
+                           '\nThe error message is: %s', six.text_type(e))
+
     def run(self):
         # Certain options can be used only with specific build tools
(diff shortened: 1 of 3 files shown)


rebasehelper/application.py
```diff
@@ -777,6 +777,17 @@ def get_rebasehelper_data(self):
         rh_stuff['logs'] = self.get_all_log_files()
         return rh_stuff
 
+    def apply_changes(self):
+        repo = git.Repo(self.execution_dir)
```
nforro: This could raise an exception if there is no repository. You should initialize one in such case.

rebasehelper/options.py
```diff
@@ -91,6 +91,12 @@
         "dest": "cont",
         "help": "continue previously interrupted rebase",
     },
+    {
+        "name": ["--apply-changes"],
+        "default": False,
+        "switch": True,
+        "help": "apply changes patch after a successful rebase",
```
nforro: This should be either `apply changes.patch` or just `apply changes`.

rebasehelper/application.py
```diff
@@ -777,6 +777,20 @@ def get_rebasehelper_data(self):
...
         return rh_stuff
 
+    def apply_changes(self):
+        try:
+            repo = git.Repo(self.execution_dir)
+        except git.InvalidGitRepositoryError:
+            repo = git.Repo.init(self.execution_dir)
+        patch = results_store.get_changes_patch()
+        if not patch:
+            logger.warning('Cannot apply changes patch. No patch file was created')
```
nforro: Again, this should be `changes.patch`.

rebasehelper/application.py
```diff
@@ -777,6 +777,20 @@ def get_rebasehelper_data(self):
...
+        try:
+            repo.git.am(patch['changes_patch'])
+        except git.GitCommandError as e:
+            logger.warning('Changes patch was not applied properly. Please review changes manually.'
```
nforro: Adn same here.

PR #481, opened 2018-04-13
Merge settings.py into constants.py

FrNecas:


rebasehelper/constants.py
```diff
@@ -20,8 +20,32 @@
...
 NEW_ISSUE_LINK = 'https://github.com/rebase-helper/rebase-helper/issues/new'
 
+REBASE_HELPER_RESULTS_DIR = "rebase-helper-results"
+REBASE_HELPER_WORKSPACE_DIR = "rebase-helper-workspace"
+REBASE_HELPER_REBASED_SOURCES_DIR = os.path.join(REBASE_HELPER_RESULTS_DIR, "rebased-sources")
+
+REBASE_HELPER_LOGS = "logs"
+REBASE_HELPER_DEBUG_LOG = 'debug.log'
+REBASE_HELPER_REPORT_LOG = 'report.log'
+REBASE_HELPER_LOGS_DIR = os.path.join(REBASE_HELPER_RESULTS_DIR, REBASE_HELPER_LOGS)
```
nforro: This is not used anywhere, but `REBASE_HELPER_LOGS` is. That doesn't make much sense to me, and it's also inconsistent with `REBASE_HELPER_REBASED_SOURCES_DIR` for example.

rebasehelper/constants.py
```diff
@@ -20,8 +20,32 @@
...
+
+OLD_SOURCES = 'old_sources'
+NEW_SOURCES = 'new_sources'
+OLD_SOURCES_DIR = os.path.join(REBASE_HELPER_WORKSPACE_DIR, OLD_SOURCES)
+NEW_SOURCES_DIR = os.path.join(REBASE_HELPER_WORKSPACE_DIR, NEW_SOURCES)
+
+GIT_CONFIG = '.gitconfig'
+
+BEGIN_COMMENT = '#BEGIN THIS MODIFIED BY REBASE-HELPER'
+END_COMMENT = '#END THIS MODIFIED BY REBASE-HELPER'
```
nforro: This is not exactly in sope of this PR, but I'll mention it here. I'm not sure if the code that inserts these strings is actually working, but in any case, the text is kind of weird. I think something like `# BEGIN rebase-helper MODIFIED BLOCK` would be more appropriate.

rebasehelper/constants.py
```diff
@@ -20,8 +20,32 @@
...
+
+ABIDIFF_ERROR = 1
+ABIDIFF_USAGE_ERROR = 2
```
nforro: These two should be part of `AbiCheckerTool` class.

rebasehelper/constants.py
FrNecas: Good point, the comment messages don't make sense. I will have a look at the code using these constants and see if it works. If it doesn't, I will leave it for another PR and fix it

rebasehelper/constants.py
nforro: Actually my point was to use `REBASE_HELPER_LOGS_DIR` and not `REBASE_HELPER_LOGS`. The inconsistency remains. Take look at this for example: https://github.com/rebase-helper/rebase-helper/blob/master/rebasehelper/application.py#L347-L351

rebasehelper/tests/test_specfile.py
```diff
@@ -331,9 +331,9 @@ def test_spec_missing_file(self, spec_object):
         files = {'missing': ['/usr/bin/test2']}
         spec_object.modify_spec_files_section(files)
         section = spec_object.get_spec_section('%files')
-        expected = ['#BEGIN THIS MODIFIED BY REBASE-HELPER\n',
+        expected = ['#BEGIN rebase-helper MODIFIED BLOCK\n',
```
nforro: Wouldn't it be better to simply use the constant in the tests as well?

rebasehelper/constants.py
nforro: Or the other way around, drop `REBASE_HELPER_REBASED_SOURCES_DIR`, I'm not sure which is better.

FrNecas: It probably makes more sense to use `REBASE_HELPER_LOGS_DIR`. `REBASE_HELPER_LOGS` is always used to create a path but the path is created in the code. It is not used in any other way so it is better to have the path as a constant.

rebasehelper/application.py
```diff
@@ -344,11 +341,11 @@ def _check_results_dir(results_dir):
...
-        os.makedirs(os.path.join(results_dir, settings.REBASE_HELPER_LOGS))
-        os.makedirs(os.path.join(results_dir, 'old-build'))
-        os.makedirs(os.path.join(results_dir, 'new-build'))
-        os.makedirs(os.path.join(results_dir, 'checkers'))
-        os.makedirs(os.path.join(results_dir, 'rebased-sources'))
+        os.makedirs(os.path.join(results_dir, constants.REBASE_HELPER_LOGS_DIR))
+        os.makedirs(os.path.join(results_dir, constants.REBASE_HELPER_OLD_BUILD_DIR))
+        os.makedirs(os.path.join(results_dir, constants.REBASE_HELPER_NEW_BUILD_DIR))
+        os.makedirs(os.path.join(results_dir, constants.REBASE_HELPER_CHECKERS_DIR))
+        os.makedirs(os.path.join(results_dir, constants.REBASE_HELPER_REBASED_SOURCES_DIR))
```
nforro: Maybe it would be worth it in this case to import all the constants using wildcard (`from rebasehelper.constants import *`). Also `REBASE_HELPER_` prefix is kind of unnecessary I would say.

rebasehelper/application.py
```diff
@@ -131,7 +131,7 @@ def setup(cli_conf):
         # parameter even when directory does not exist
         if not os.path.exists(results_dir):
             os.makedirs(results_dir)
-            os.makedirs(os.path.join(results_dir, settings.REBASE_HELPER_LOGS))
+            os.makedirs(constants.REBASE_HELPER_LOGS_DIR)
```
nforro: You are ignoring `results_dir` here.

rebasehelper/constants.py
```diff
@@ -20,8 +20,28 @@
...
+WORKSPACE_DIR = 'rebase-helper-workspace'
+
+REBASED_SOURCES_DIR = 'rebased-sources'
+OLD_BUILD_DIR = 'old-build'
+NEW_BUILD_DIR = 'new-build'
+CHECKERS_DIR = 'checkers'
+
+LOGS_DIR = 'logs'
+DEBUG_LOG = 'debug.log'
+REPORT = 'report.'
```
nforro: I think the dot shouldn't be part of the constant.

PR #486, opened 2018-04-19
Add licensecheck dependency for licensecheck

skisela:


nforro:
You should also add it to https://github.com/rebase-helper/rebase-helper/blob/master/docs/user_guide/installation.rst.

Issue #448, opened 2018-02-09
Make logging consistent and allow for multiple verbosity choices


nforro:
Log messages should be consistent in level and form. There should be multiple log files of different verbosity, and verbosity of the console output should be configurable in several steps.

PR #488, opened 2018-04-21
Add another verbosity level and create logs of different verbosity

FrNecas:
Resolves: #448

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,15 +35,15 @@ See [docker reference](https://rebase-helper.readthedocs.io/en/latest/user_guide
 
 **rebase-helper** always creates *rebase-helper-results* directory containing the following items:
 
-| Path                      | Description                                                       |
-|:------------------------- |:----------------------------------------------------------------- |
-| *report.txt*              | summary report with all important information                     |
-| *changes.patch*           | diff against original files, directly applicable to dist-git repo |
-| *rebased-sources/*        | git repository with all modified files                            |
-| *checkers/*               | reports from individual checkers that were run                    |
-| *old-build/*              | logs and results of old (original) version build                  |
-| *new-build/*              | logs and results of new (rebased) version build                   |
-| *debug.log*               | log containing the most verbose output of a rebase                |
+| Path                  | Description                                                       |
+|:--------------------- |:----------------------------------------------------------------- |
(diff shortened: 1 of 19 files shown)


README.md
```diff
@@ -39,11 +39,11 @@ See [docker reference](https://rebase-helper.readthedocs.io/en/latest/user_guide
 |:------------------------- |:----------------------------------------------------------------- |
 | *report.txt*              | summary report with all important information                     |
 | *changes.patch*           | diff against original files, directly applicable to dist-git repo |
+| *logs*                    | logs with various verbosity levels                                |
```
nforro: This should end with `/` to make it obvious that it's a directory.

nforro: Also, the width of the first column could be adjusted to match the longest string, it is now unnecessary wide.

nforro:
I think the commits (the first one especially) should be split, one commit for adding the new level, another for `CustomAction` modification, and yet another for `logger.debug()` to `logger.verbose()` changes.

README.md
```diff
@@ -35,15 +35,15 @@ See [docker reference](https://rebase-helper.readthedocs.io/en/latest/user_guide
...
-| *rebased-sources/*        | git repository with all modified files                            |
-| *checkers/*               | reports from individual checkers that were run                    |
-| *old-build/*              | logs and results of old (original) version build                  |
-| *new-build/*              | logs and results of new (rebased) version build                   |
-| *debug.log*               | log containing the most verbose output of a rebase                |
+| Path                  | Description                                                       |
+|:--------------------- |:----------------------------------------------------------------- |
+| *report.txt*          | summary report with all important information                     |
+| *changes.patch*       | diff against original files, directly applicable to dist-git repo |
+| *logs/*               | logs with various verbosity levels                                |
```
nforro: I would use `of` instead of `with`. And also maybe `log files` instead of `logs`.

rebasehelper/cli.py
```diff
@@ -49,6 +49,7 @@ def _expand_help(self, action):
 class CustomAction(argparse.Action):
     def __init__(self, option_strings,
                  switch=False,
+                 counter=False,
```
nforro: The attribute name should be updated also in the commit message.

PR #499, opened 2018-05-17
Fix rebuilding

### LINKED ISSUE #468: Making changes to a spec file before build retry doesn't work anymore
SRPM needs to be rebuilt as well for the changes to take effect.

skisela:
Fixes #468

### DIFF
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -570,113 +570,88 @@ class Application(object):
                 six.text_type(e), Builder.get_supported_tools()))
 
         for version in ['old', 'new']:
-            successful_builds = 0
-            try_build_again = False
-            while successful_builds < 1:
-                results_dir = '{}-build'.format(os.path.join(self.results_dir, version))
-                spec = None
-                task_id = None
-                koji_build_id = None
-                build_dict = {}
+            results_dir = '{}-build'.format(os.path.join(self.results_dir, version))
+            spec = None
+            task_id = None
+            koji_build_id = None
+            build_dict = {}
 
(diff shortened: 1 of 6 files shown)


rebasehelper/application.py
```diff
@@ -804,6 +779,26 @@ def apply_changes(self):
             logger.warning('changes.patch was not applied properly. Please review changes manually.'
                            '\nThe error message is: %s', six.text_type(e))
 
+    def fix_specfile(self, results_dir):
```
nforro: This shouldn't be called `fix_specfile`, that's misleading. Something like `prepare_next_run` would be more fitting.

rebasehelper/application.py
```diff
@@ -846,18 +841,22 @@ def run(self):
...
+                while True:
+                    try:
+                        if self.conf.build_tasks is None:
+                            self.build_source_packages()
+                        self.run_package_checkers(self.results_dir, category='SRPM')
+                        self.build_binary_packages()
+                        if self.conf.builds_nowait and not self.conf.build_tasks:
+                            return
+                        self.run_package_checkers(self.results_dir, category='RPM')
+                        break
```
nforro: Just a nitpick, but I think this `break` fits more into a `else` block of the `try ... except` statement.

skisela:
Fixed, thanks!

I also added commit https://github.com/rebase-helper/rebase-helper/pull/499/commits/5c57d19a95f3c6bfb2f8a5ace41ed19952ced135 so that `setup.py` knows about the freshly added `helpers` package created by @FrNecas .

PR #502, opened 2018-05-23
Transform sources URLs from pypi.org to files.pythonhosted.org

FrNecas:


rebasehelper/spec_hooks/pypi_url_fix.py
```diff
@@ -62,6 +63,6 @@ def _transform_url(cls, line):
         Perform predefined URL transformations
         """
         for trans in cls.SOURCES_URL_TRANSFORMATIONS:
-            if re.search(r".*https?://pypi.python.org/.*", line):
+            if re.search(r".*" + trans[0] + r".*", line):
```
nforro: First, why prepend and append `r'.*'`? `re.search()` would find it anyway.
Second, why two regexps, when one is a subset of the other?
Third, this could be done with just `re.sub()` or `re.search()`, using both is unnecessary.

rebasehelper/spec_hooks/pypi_url_fix.py
```diff
@@ -34,7 +34,7 @@ class PyPIURLFixHook(BaseSpecHook):
     NAME = 'PyPI URL Fix'
     CATEGORIES = ['python']
     SOURCES_URL_TRANSFORMATIONS = [
-        ('https?://pypi.python.org/', 'https://files.pythonhosted.org/'),
+        (r'https?://pypi(.python)?.org/', 'https://files.pythonhosted.org/'),
```
nforro: This is almost fine, only `.` matches any character, not just literal `.`, so it would be safer to escape it.

rebasehelper/spec_hooks/pypi_url_fix.py
```diff
@@ -34,7 +34,7 @@ class PyPIURLFixHook(BaseSpecHook):
     NAME = 'PyPI URL Fix'
     CATEGORIES = ['python']
     SOURCES_URL_TRANSFORMATIONS = [
-        ('https?://pypi.python.org/', 'https://files.pythonhosted.org/'),
+        (r'https?://pypi(.python)?\.org/', 'https://files.pythonhosted.org/'),
```
nforro: What about the first dot?

PR #503, opened 2018-05-23
Transform URL from pypi.python.org to pypi.org

FrNecas:


rebasehelper/spec_hooks/pypi_url_fix.py
```diff
@@ -36,6 +36,9 @@ class PyPIURLFixHook(BaseSpecHook):
     SOURCES_URL_TRANSFORMATIONS = [
         (r'https?://pypi(\.python)?\.org/', 'https://files.pythonhosted.org/'),
     ]
+    URL_TRANSFORMATIONS = [
+        (r'https?://pypi(\.python)?\.org/', 'https://pypi.org/'),
```
nforro: I think the actual change is from `https://pypi.python.org/pypi/<NAME>` to `https://pypi.org/project/<NAME>`.

rebasehelper/spec_hooks/pypi_url_fix.py
```diff
@@ -53,14 +56,25 @@ def run(cls, spec_file, rebase_spec_file, **kwargs):
...
+            elif line.startswith("URL"):
                 rebase_spec_file.spec_content[index] = cls._transform_url(line)
         rebase_spec_file.save()
 
     @classmethod
-    def _transform_url(cls, line):
+    def _transform_sources_url(cls, line):
         """
-        Perform predefined URL transformations
+        Perform predefined URL transformations to sources line.
```
nforro: What is `sources line`? It should be `on Source line` probably.

nforro: But the docstring is probably redundant in this case, the method is short and it's obvious what it does.

rebasehelper/spec_hooks/pypi_url_fix.py
```diff
@@ -26,16 +26,19 @@
 
 
 class PyPIURLFixHook(BaseSpecHook):
-    """
-    SpecHook for transforming old nonfunctional python package url to
-    a safe url(eg. https://pypi.* to https://files.pythonhosted.*)
+    """SpecHook for transforming an old python package url to a new url(eg. https://pypi.* to
+    https://files.pythonhosted.* or https://pypi.python.org/pypi/* to https://pypi.org/project/*)
```
nforro: I'd rather write that as `https://files.pythonhosted.org/packages/*` instead of `https://files.pythonhosted.*` - that's just too generic.

rebasehelper/spec_hooks/pypi_url_fix.py
```diff
@@ -26,16 +26,19 @@
...
+
     """
 
     NAME = 'PyPI URL Fix'
     CATEGORIES = ['python']
     SOURCES_URL_TRANSFORMATIONS = [
         (r'https?://pypi(\.python)?\.org/', 'https://files.pythonhosted.org/'),
     ]
+    URL_TRANSFORMATIONS = [
+        (r'https?://pypi(\.python)?\.org/pypi/', 'https://pypi.org/project/'),
```
nforro: I don't think anybody is using `https://pypi.org/pypi`, as that never has been a valid URL AFAIK.

rebasehelper/spec_hooks/pypi_url_fix.py
```diff
@@ -47,20 +50,21 @@ def get_categories(cls):
...
     @classmethod
     def run(cls, spec_file, rebase_spec_file, **kwargs):
-        """
-        Run _transform_url() for all sources to replace all pypi.* urls by
-        files.pythonhosted.* urls.
-        """
         for index, line in enumerate(rebase_spec_file.spec_content):
             if line.startswith("Source"):
+                rebase_spec_file.spec_content[index] = cls._transform_sources_url(line)
+            elif line.startswith("URL"):
```
nforro: Also the order here ;-)

rebasehelper/spec_hooks/pypi_url_fix.py
```diff
@@ -26,13 +26,16 @@
 
 
 class PyPIURLFixHook(BaseSpecHook):
-    """
-    SpecHook for transforming old nonfunctional python package url to
-    a safe url(eg. https://pypi.* to https://files.pythonhosted.*)
+    """SpecHook for transforming an old python package url to a new url(eg. https://pypi.python.org/packages/* to
+    https://files.pythonhosted.org/packages/* or https://pypi.python.org/pypi/* to https://pypi.org/project/*)
+
```
nforro: This is fine, but the lines are too long. The 120 characters per-line limit doesn't apply to docstrings, because they can be used in different environments that may be expecting maximum of 80 characters per line, such as interactive python shell or sphinx documentation.

FrNecas: Oh, thanks for pointing that out, didn't know that.

rebasehelper/spec_hooks/pypi_url_fix.py
```diff
@@ -26,13 +26,18 @@
 
 
 class PyPIURLFixHook(BaseSpecHook):
-    """
-    SpecHook for transforming old nonfunctional python package url to
-    a safe url(eg. https://pypi.* to https://files.pythonhosted.*)
+    """SpecHook for transforming an old python package url to a new url(eg.
```
nforro: While at it, how about moving the opening bracket to the next line (there is missing space anyway), and capitalizing URL (I don't really understand why it's lowercase now)?

PR #509, opened 2018-06-25
Add PathsToRPMMacros SPEC Hook

FrNecas:


nforro:
You should also create a test for the SPEC hook.

rebasehelper/tests/testing_files/test.spec
```diff
@@ -75,6 +75,10 @@ make check
 
 %files
 %{_bindir}/file.txt
+/usr/share/man/man1/*
+/usr/share/test1.txt
+/usr/bin/test2.txt
+/dev/test3
```
nforro: I think we should try to keep the testing spec file sane. It doesn't make sense to install a `txt` file to `/usr/bin`, nor installing anything to `/dev`. Since it's probably impossible to find a real path that wouldn't be substituted with any macro, I'd use something like `/no/macros/here`.

You should also add something like `%config(noreplace) /etc/test/test.conf`.

rebasehelper/spec_hooks/paths_to_rpm_macros.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @classmethod
+    def _expand_macros(cls, macros):
+        for macro in macros:
+            macro['value'] = MacroHelper.expand(macro['value'])
+        return macros
+
+    @classmethod
+    def run(cls, spec_file, rebase_spec_file, **kwargs):
+        files = rebase_spec_file.get_spec_section('%files')
```
nforro: This doesn't work with `%files` section with arguments, e.g. `%files -f %{name}.lang`.
Every package containing translations is (or at least should be) using this, see: https://fedoraproject.org/wiki/Packaging:Guidelines#Handling_Locale_Files

rebasehelper/tests/test_specfile.py
```diff
@@ -341,6 +350,27 @@ def test_typo_fix_spec_hook(self, spec_object):
...
+            assert path in spec_object.spec_content.sections['%files']
+        for path in files_devel:
+            assert path in spec_object.spec_content.sections['%files devel']
+        PathsToRPMMacrosHook.run(spec_object, spec_object)
+
+        for expected_output in six.itervalues(files):
+            assert expected_output in spec_object.spec_content.sections['%files']
+        for expected_output in six.itervalues(files_devel):
+            assert expected_output in spec_object.spec_content.sections['%files devel']
+
```
nforro: I think it would be better to compare the whole section, like in `test_spec_missing_file()` and `test_spec_missing_and_remove_file()`.

FrNecas: Why do you think it is better? If you add a line to the section, you have to modify every test working with the section. This way it only checks the needed lines

nforro: Because you are not really testing that the specified paths changed as expected (line index is discarded), if someone adds a different path that is equal to some of the current paths after processing, the test will no longer be reliable.
But I suppose it's not really such a big problem here.

FrNecas: I didn't think about that, thanks for pointing it out. I will change the test to compare whole sections

PR #510, opened 2018-06-28
Run only one SPEC hook when testing its functionality

FrNecas:


rebasehelper/tests/test_specfile.py
```diff
@@ -459,16 +459,14 @@ def test_find_archive_target_in_prep(self, spec_object):
     def test_typo_fix_spec_hook(self, spec_object):
         assert TypoFixHook.get_name() in spec_hooks_runner.spec_hooks
```
nforro: This assert does not make sense anymore.

rebasehelper/tests/test_specfile.py
```diff
@@ -459,16 +459,14 @@ def test_find_archive_target_in_prep(self, spec_object):
...
         assert '- This is chnagelog entry with some indentional typos\n' in spec_object.spec_content
-        spec_object.category = 'sample'
-        spec_hooks_runner.run_spec_hooks(spec_object, spec_object)
+        TypoFixHook.run(spec_object, spec_object)
         assert '- This is changelog entry with some intentional typos\n' in spec_object.spec_content
 
     def test_pypi_to_python_hosted_url_trans(self, spec_object):
         # pylint: disable=protected-access
         assert PyPIURLFixHook.get_name() in spec_hooks_runner.spec_hooks
```
nforro: Same as above.

PR #516, opened 2018-07-30
Enhance SpecFile class

FrNecas:


rebasehelper/specfile.py
```diff
@@ -102,6 +103,85 @@ def get_strip(self):
         return self.strip
 
 
+class SpecContent(object):
+
+    """Class representing content of a SPEC file."""
```
nforro: There shouldn't be empty lines around one-line docstring.

rebasehelper/specfile.py
```diff
@@ -102,6 +103,85 @@ def get_strip(self):
...
+
+    defined_section_headers = [
```
nforro: I'd use `known_section_headers` or maybe just `section_headers`. Also, if I'm not mistaken, in most other classes *class constants* are uppercase.

rebasehelper/specfile.py
```diff
@@ -102,6 +103,85 @@ def get_strip(self):
...
+
+    def _split_sections(self, content):
+        """Splits content of a SPEC file into sections.
+
+        Args:
+            content (str): Content of the SPEC file
+
+        """
+        lines = content.splitlines()
+        section_headers_re = [re.compile(r'^{0}.*'.format(x), re.I) for x in self.defined_section_headers]
```
nforro: You should always use `re.escape()` when constructing regexp from variables. Though it's not strictly necessary here, it's a good practice to follow.

rebasehelper/specfile.py
```diff
@@ -1281,9 +1252,8 @@ def _correct_removed_files(self, sources):
                 for f in f_exists:
                     for index, row in enumerate(sec_content):
                         if f in row:
-                            sec_content[index: index+1] = SpecFile.construct_string_with_comment('#' + row)
-                            break
-                self.rpm_sections[key] = (sec_name, sec_content)
+                            del sec_content[index]
```
nforro: You can't delete items from a list while iterating through it. You have to either iterate backwards, or use list comprehension for example.

rebasehelper/specfile.py
```diff
@@ -102,6 +103,82 @@ def get_strip(self):
...
+        '%filetriggerpostun',
+        '%transfiletriggerin',
+        '%transfiletrigger',
+        '%transfiletriggerun',
+        '%transfiletriggerpostun',
+    ]
+
+    def __init__(self, content):
+        self.sections = CaseInsensitiveDict()
+        self._split_sections(content)
```
nforro: This could be implemented as a class method returning the dict, that way it would be possible to use it even outside this class (not that it's necessary, but it makes a little more sense to me).

rebasehelper/specfile.py
```diff
@@ -102,6 +103,82 @@ def get_strip(self):
...
+
+    def _split_sections(self, content):
+        """Splits content of a SPEC file into sections.
+
+        Args:
+            content (str): Content of the SPEC file
+
+        """
+        lines = content.splitlines()
+        section_headers_re = [re.compile(r'^{0}.*'.format(re.escape(x)), re.I) for x in self.SECTION_HEADERS]
```
nforro: I would rather see `re.IGNORECASE` here, for clarity.

rebasehelper/specfile.py
```diff
@@ -102,6 +103,82 @@ def get_strip(self):
...
+                    if header.match(line):
+                        section_beginnings.append(i)
+        section_beginnings.append(None)
+
+        self.sections['%package'] = lines[:section_beginnings[0]]
+
+        for i in range(len(section_beginnings) - 1):
+            start = sb[i] + 1
+            end = sb[i + 1]
+            self.sections[lines[sb[i]]] = lines[start:end]
```
nforro: I don't think shortening `section_beginnings` is necessary now, and you could also use lines[start - 1] here.

PR #517, opened 2018-08-03
Enable --strategy-option for rebasing

seki98:
"git rebase" --strategy-option makes use of prefering one or the other
sides that are being merged.
"upstream" in context of rebase-helper rebasing refers to the upstream code,
while "downstream" refers to the downstream code(with the downstream patches).

https://git-scm.com/docs/git-rebase#git-rebase--Xltstrategy-optiongt

### DIFF
--- a/rebasehelper/options.py
+++ b/rebasehelper/options.py
@@ -147,6 +147,13 @@ OPTIONS = [
         "help": "do not interact with user",
     },
     {
+        "name": ["--favor-on-conflict"],
+        "choices": ["downstream", "upstream", "off"],
+        "default": "off",
+        "dest": "favor_on_conflict",
+        "help": "favor downstream or upstream changes when conflicts appear",
+    },
+    {
         "name": ["--not-download-sources"],
         "default": False,
         "switch": True,
--- a/rebasehelper/patch_helper.py
+++ b/rebasehelper/patch_helper.py
@@ -145,8 +145,15 @@ class GitPatchTool(PatchBase):
             cls.old_repo.create_remote(upstream, url=cls.new_sources).fetch()
             root_commit = cls.old_repo.git.rev_list('HEAD', max_parents=0)
             last_commit = cls.old_repo.commit('HEAD')
+            if cls.favor_on_conflict == 'upstream':
+                strategy_option = 'ours'
+            elif cls.favor_on_conflict == 'downstream':
+                strategy_option = 'theirs'
+            else:
+                strategy_option = False
             try:
                 cls.output_data = cls.old_repo.git.rebase(root_commit, last_commit,
+                                                          strategy_option=strategy_option,
                                                           onto='{}/master'.format(upstream),
                                                           stdout_as_string=six.PY3)
             except git.GitCommandError as e:
@@ -279,6 +286,7 @@ class GitPatchTool(PatchBase):
         cls.rest_sources = rest_sources
         cls.patches = patches
         cls.non_interactive = kwargs.get('non_interactive')
+        cls.favor_on_conflict = kwargs.get('favor_on_conflict')
         if not os.path.isdir(os.path.join(cls.old_sources, '.git')):
             cls.old_repo = cls.init_git(old_dir)
             cls.new_repo = cls.init_git(new_dir)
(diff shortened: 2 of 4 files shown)


skisela:
With this PR, there will be 4 options for patching:
1, Resolving conflicts manually.
2, `--non-interactive mode` - not applying conflicting patches.
3, `--strategy-option ours` - prefering all upstream changes without interaction
4,` --strategy-option theirs` - prefering all downstream changes without interaction

rebasehelper/options.py
```diff
@@ -146,6 +146,15 @@
         "dest": "non_interactive",
         "help": "do not interact with user",
     },
+    {
+        "name": ["--strategy-option"],
+        "choices": ["ours", "theirs"],
```
nforro: I don't like this too much. Although this is close to actual git arguments, I think it would be more user friendly to call the option something like `--favor-on-conflict` with choices `upstream` and `downstream`.

rebasehelper/options.py
```diff
@@ -146,6 +146,13 @@
         "dest": "non_interactive",
         "help": "do not interact with user",
     },
+    {
+        "name": ["--favor-on-conflict"],
+        "choices": ["downstream", "upstream"],
+        "default": False,
+        "dest": "strategy_option",
```
nforro: I think this is incorrect.

rebasehelper/patch_helper.py
```diff
@@ -279,6 +280,10 @@ def run_patch(cls, old_dir, new_dir, rest_sources, patches, **kwargs):
         cls.rest_sources = rest_sources
         cls.patches = patches
         cls.non_interactive = kwargs.get('non_interactive')
+        cls.favor_on_conflict = kwargs.get('favor_on_conflict')
+        if cls.favor_on_conflict:
+            # transform upstream to 'ours' and downstream to 'theirs' according to git strategy option
+            cls.favor_on_conflict = 'ours' if (cls.favor_on_conflict == 'upstream') else 'theirs'
```
nforro: Not that it matters too much, but I think it would be less confusing to do this in `_git_rebase` using a local variable called `strategy_option` for instance.

rebasehelper/tests/functional/test_rebase.py
```diff
@@ -111,3 +112,57 @@ def test_rebase(self, buildtool):
...
         assert '- New upstream release {}'.format(self.NEW_VERSION) in repo.commit().summary
+
+    @pytest.mark.parametrize('buildtool', [
+        pytest.mark.skipif(
+            os.geteuid() != 0,
+            reason='requires superuser privileges')
+        ('rpmbuild'),
+        pytest.mark.long_running('mock'),
+    ])
+    @pytest.mark.parametrize('favor_on_conflict', ['upstream', 'downstream'])
```
nforro: Actually what I had in mind was to parametrize `test_rebase` method itself, with `favor_on_conflict` being `None`, `'upstream'` and `'downstream'`.

nforro:
Could you change base of this PR to `nforro-extend-functional-test`?

rebasehelper/patch_helper.py
```diff
@@ -145,8 +145,13 @@ def _git_rebase(cls):
             cls.old_repo.create_remote(upstream, url=cls.new_sources).fetch()
             root_commit = cls.old_repo.git.rev_list('HEAD', max_parents=0)
             last_commit = cls.old_repo.commit('HEAD')
+            strategy_option = False
+            if cls.favor_on_conflict:
+                # transform upstream to 'ours' and downstream to 'theirs' according to git strategy option
+                strategy_option = 'ours' if (cls.favor_on_conflict == 'upstream') else 'theirs'
```
nforro: I don't think the parentheses are necessary.

rebasehelper/options.py
```diff
@@ -146,6 +146,13 @@
         "dest": "non_interactive",
         "help": "do not interact with user",
     },
+    {
+        "name": ["--favor-on-conflict"],
+        "choices": ["downstream", "upstream"],
+        "default": False,
```
nforro: I think `None` would fit here better.

Issue #498, opened 2018-05-17
--get-old-build-from-koji builds old sources and saves RPM's to wrong dir


skisela:
`--get-old-build-from-koji` should download (S)RPM's and place them in the correct directory. Instead, it builds old SRPM, downloads RPM's and places them in `$(RESULTS_DIR)/old-build`.

PR #518, opened 2018-08-03
Place RPMs from old build to correct directory

skisela:
Place RPMs from old build gathered from previous koji build
into the correct old-build/RPMS directory.

Fixes: #498

Signed-off-by: Sebastian Kisela <skisela@redhat.com>

### DIFF
--- a/rebasehelper/helpers/koji_helper.py
+++ b/rebasehelper/helpers/koji_helper.py
@@ -288,13 +288,14 @@ class KojiHelper(object):
         return None, None
 
     @classmethod
-    def download_build(cls, session, build_id, destination):
+    def download_build(cls, session, build_id, destination, arches):
         """Downloads RPMs and logs of a Koji build.
 
         Args:
             session (koji.ClientSession): Active Koji session instance.
             build_id (str): Koji build ID.
             destination (str): Path where to download files to.
+            arches (list): List of architectures to be downloaded.
 
         Returns:
             tuple: List of downloaded RPMs and list of downloaded logs.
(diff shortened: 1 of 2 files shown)


nforro:
I think this doesn't fix #498 entirely, SRPMs shouldn't be built, but also downloaded from Koji.

rebasehelper/helpers/koji_helper.py
```diff
@@ -288,13 +288,14 @@ def get_latest_build(cls, session, package):
         return None, None
 
     @classmethod
-    def download_build(cls, session, build_id, destination):
+    def download_build(cls, session, build_id, destination, pkg_type='RPM'):
```
nforro: Why not make this method take list of architectures as an argument? And special case `src` as it has no logs?

rebasehelper/helpers/koji_helper.py
```diff
@@ -307,36 +308,80 @@ def download_build(cls, session, build_id, destination):
...
+                if koji_version != package_version:
+                    logger.warning('Version of the latest Koji build (%s) with id (%s) '
+                                   'differs from version in SPEC file (%s)!',
+                                   koji_version, koji_build_id, package_version)
+                package_version = package_full_version = koji_version
+                return koji_build_id, package_version, package_full_version
+            else:
+                logger.warning('Unable to find the latest Koji build!')
+        else:
+            logger.warning('Unable to get the latest Koji build!')
```
nforro: What happens if any of these two error cases occurs? The method is expected to return a 3-tuple.

skisela: I guess some kind of new `KojiError` exception should be raised, as no info is provided at all.

rebasehelper/helpers/koji_helper.py
skisela: so `new-build/` directory content of `rebase-helper-results` would be changed from `RPM, SRPM` to `x86_64, src, aarch64..`

nforro: No, nothing would change, only this method would accept list of architectures - `['noarch', 'x86_64']` in case of RPM build and `['src']` in case of SRPM build.

rebasehelper/helpers/koji_helper.py
nforro: You'll get `ValueError` when trying to unpack `None` (the implicit return value) into `koji_build_id, package_version, package_full_version`.
The old code expects `koji_build_id` to be either build id or `None`, so you could just return `None, None, None` and it should continue working, unless I'm missing something.

rebasehelper/helpers/koji_helper.py
```diff
@@ -307,12 +308,30 @@ def download_build(cls, session, build_id, destination):
...
         rpms = []
         logs = []
+        os.makedirs(os.path.join(destination, 'RPM'), exist_ok=True)
+        os.makedirs(os.path.join(destination, 'SRPM'), exist_ok=True)
         for pkg in packages:
-            # FIXME: multiple arches
-            if pkg['arch'] not in ['noarch', 'x86_64']:
+            if pkg['arch'] not in arches:
                 continue
+            if pkg['arch'] == 'src':
```
nforro: Why the duplicate code? You can simply skip downloading logs for `'src'` arch, can't you? RPM/SRPM should already be part of `destination`.

PR #530, opened 2018-10-31
Add build log hooks

### LINKED ISSUE #492: Get rid of SpecFile.get_paths_with_rpm_macros()
Move its functionality into a spec hook and avoid using hardcoded macro values.

FrNecas:
Resolves #492


rebasehelper/helpers/macro_helper.py
```diff
@@ -41,6 +41,40 @@ def expand(s, default=None):
...
+        Returns:
+            list: List of macros with expanded values.
+
+        """
+        for macro in macros:
+            macro['value'] = MacroHelper.expand(macro['value'])
+        return macros
+
+    @staticmethod
+    def substitute_path_for_macro(path, macros):
```
nforro: Why is the method called like this? Shouldn't it rather be `substitute_path_with_macros`?

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -0,0 +1,169 @@
...
+
+        Args:
+            log_path (str): Path to the RPM build log.
+
+        Returns:
+            tuple: The first element is the type of error that was found
+                in the log (missing or deleted). The second element is
+                a set of problematic files.
+
+        """
```
nforro: I'd rather see this entire method implemented with regular expressions, maybe including the 'BUILDSTDERR' prefix.

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -0,0 +1,169 @@
...
+        # ensure maximal greediness
+        macros.sort(key=lambda k: len(k['value']), reverse=True)
+
+        for file in files:
+            section = cls._get_best_matching_files_section(rebase_spec_file, file)
+            if section:
+                substituted_path = MacroHelper.substitute_path_for_macro(file, macros)
+                rebase_spec_file.spec_content.sections[section].append(substituted_path)
+            else:
+                logger.info('Could not correctly add file %s to the SPEC file.', file)
```
nforro: What about falling back to `%files` (the general section)?

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -0,0 +1,169 @@
...
+    def _correct_deleted_files(cls, rebase_spec_file, files):
+        """Removes files missing in sources from the SPEC file."""
+        for sec_name, sec_content in six.iteritems(rebase_spec_file.spec_content.sections):
+            if sec_name.startswith('%files'):
+                modified_section = []
+                for line in sec_content:
+                    file_is_ok = True
+                    if line != '':
+                        for file in files:
+                            if MacroHelper.expand(line).split(' ')[-1] in file:
```
nforro: The expansion and splitting should happen before the loop, it's pointless to do it in every iteration.

rebasehelper/build_log_hooks/files_build_log_hook.py
nforro: With sample build.log lines in comments.

rebasehelper/build_log_hooks/files_build_log_hook.py
nforro: Also, why `in`?

rebasehelper/options.py
```diff
@@ -213,6 +213,12 @@
         "switch": True,
         "help": "update \"sources\" file and upload new sources to lookaside cache",
     },
+    {
+        "name": ["--run-build-log-hooks"],
```
nforro: What about `--force-build-log-hooks`?

rebasehelper/build_log_hooks/files_build_log_hook.py
FrNecas: It expands and splits lines of the SPEC file so in each iteration it expands and splits one unique line from the spec. The files in build log are already expanded.

Also, the paths from build log contain build directory path, therefore the specfile path is a subset of the log path. I think it will be optimal to strip build directory from the path and then compare using fnmatch.

nforro: > It expands and splits lines of the SPEC file so in each iteration it expands and splits one unique line from the spec. The files in build log are already expanded.

Are you saying that in each iteration of `for file in files:` loop `line` has different value?

> Also, the paths from build log contain build directory path, therefore the specfile path is a subset of the log path. I think it will be optimal to strip build directory from the path and then compare using fnmatch.

Yes, that sounds much better.

FrNecas: Good point, I missed that one, thanks

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -0,0 +1,193 @@
...
+            return None, None
+
+        error_type = None
+        files = set()
+        next_line_is_missing = False
+
+        for line in lines:
+            # The file on the line is in the SPEC file but not in sources. Example line:
+            # File not found: /root/rpmbuild/BUILDROOT/pello-0.1.1-1.fc27.x86_64/should/not/be/here
+            match_deleted = re.match(r'(BUILDSTDERR:)?\s*File not found: (/.*)', line)
```
nforro: So, what about this?
```py
re.compile(r'^(BUILDSTDERR:)?\s*((?P<missing>File not found:\s*)|(?P<unpackaged>Installed \(but unpackaged\) file\(s\) found:))?(?P<path>/.*)?$')
```

Single regexp for all cases, of course compiled before entering the loop.

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -0,0 +1,193 @@
...
+        try:
+            with open(log_path, 'r') as build_log:
+                lines = build_log.read().splitlines()
+        except IOError:
+            logger.info('There was an error opening %s', log_path)
+            return None, None
+
+        error_type = None
+        files = set()
+        next_line_is_missing = False
```
nforro: This variable is redundant, you can use `error_type` instead, and break out of the loop or return after the last path match.

rebasehelper/build_log_hooks/files_build_log_hook.py
nforro: Two more things:

- there can be more paths on the same line, see [libtiff.spec](https://src.fedoraproject.org/rpms/libtiff/blob/b9f5e1a576c4a7e95050e3380c86a446fd4b891f/f/libtiff.spec#_169) for example
- the macro expansion will expand for example `%license COPYING` into e.g. `GPLv2+ COPYING`, which could be fine in theory, but it would be better to expand only the individual paths

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -0,0 +1,198 @@
...
+            tuple: The first element is the type of error that was found
+                in the log (missing or deleted). The second element is
+                a list of problematic files.
+
+        """
+        try:
+            with open(log_path, 'r') as build_log:
+                lines = build_log.read().splitlines()
+        except IOError:
+            logger.info('There was an error opening %s', log_path)
```
nforro: Shouldn't this rather be `logger.error()`?

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -0,0 +1,198 @@
...
+            return None, None
+
+        error_type = None
+        # Use set to avoid duplicate files.
+        files = set()
+
+        # Two types of error can occur. The file is either in the SPEC file but missing in sources
+        # or the other way around. Example lines of the first case:
+        # File not found: /root/rpmbuild/BUILDROOT/pello-0.1.1-1.fc27.x86_64/should/not/be/here
+        # File not found: /root/rpmbuild/BUILDROOT/pello-0.1.1-1.fc27_x86_64/should/not/be/here2
```
nforro: > pello-0.1.1-1.fc27_x86_64

The `_` before arch is a typo, right?

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -0,0 +1,198 @@
...
+        # Example lines of the second case:
+        # Installed (but unpackaged) file(s) found:
+        #   /usr/bin/pello
+        error_re = re.compile(r'^(BUILDSTDERR:)?\s*((?P<missing>File not found:\s*)|'
+                              r'(?P<unpackaged>Installed \(but unpackaged\) file\(s\) found:))?'
+                              r'(/.*/{}[\.\w]+)?(?P<path>/.*)?$'.format(re.escape(nvr)))
```
nforro: Why `[\.\w]+` and not `\.\w+`?

nforro: The regexp would probably be more readable if it was verbose (`re.VERBOSE`).

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -0,0 +1,198 @@
...
+        for line in lines:
+            match = re.match(error_re, line)
```
nforro: Why not `match = error_re.match(line)` directly?

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -0,0 +1,198 @@
...
+                    files.add(match.group('path'))
+                elif error_type and not match.group('path'):
+                    break
+
+        return error_type, list(files)
+
+    @classmethod
+    def _get_best_matching_files_section(cls, rebase_spec_file, file):
+        """Finds a section with a file that has the closest match with
+        the given file.
```
nforro: This formulation is rather confusing. How about `Finds a %files section containing a file that has the closest match with the specified file`?

rebasehelper/constants.py
```diff
@@ -51,3 +51,25 @@
 
 DEFENC = locale.getpreferredencoding()
 DEFENC = 'utf-8' if DEFENC == 'ascii' else DEFENC
+
+# Taken from rpm/build/files.c
```
nforro: This doesn't make sense. How about `# taken from build/files.c in RPM source`?

rebasehelper/constants.py
```diff
@@ -51,3 +51,25 @@
...
+FILES_DIRECTIVES = [
```
nforro: Does it make sense to have `SECTION_HEADERS` in `SpecContent` and `FILES_DIRECTIVES` here?

rebasehelper/specfile.py
```diff
@@ -1110,6 +1113,24 @@ def get_prep_section(self):
...
+        """Finds the exact name of the main files section.
+
+        Returns:
+            str: Name of the main files section.
+
+        """
+        parser = SilentArgumentParser()
+        parser.add_argument('-n', default=None)
+        parser.add_argument('-f')
+        parser.add_argument('section-name', nargs='?', default=None)
```
nforro: This should be rather `subpackage-name`.

nforro: And if you name it `subpackage_name` or simply `subpackage`, you don't have to use `getattr` to access it.

rebasehelper/build_log_hook.py
```diff
@@ -64,7 +64,10 @@ def run(self, spec_file, rebase_spec_file, **kwargs):
 
         """
         if not kwargs.get('non_interactive'):
+            blacklist = kwargs.get('build_log_hook_blacklist', [])
             for name, build_log_hook in six.iteritems(self.build_log_hooks):
+                if  name in blacklist:
```
nforro: Extra space here.

nforro: Ok, it is fixed in one of the following commits, but it should be fixed in this one.

rebasehelper/build_log_hooks/files_build_log_hook.py
FrNecas: I used `[\.\w]+` because the string after NVR (in the example line) is `.fc27.x86_64` so the dot can appear twice

FrNecas: I used `[\.\w]+` because the string after NVR (in the example line) is `.fc27.x86_64` so the dot can appear twice

nforro: No, it is not, `.fc27` is part of Release, see https://fedoraproject.org/wiki/Packaging:DistTag.

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -0,0 +1,241 @@
...
+        # Installed (but unpackaged) file(s) found:
+        #   /usr/bin/pello
+        error_re = re.compile(
+            r'''
+            ^
+            (BUILDSTDERR:)?
+            \s*
+            (
+            (?P<missing>File not found:\s*)|
+            (?P<unpackaged>Installed \(but unpackaged\) file\(s\) found:)
```
nforro: Spaces are ignored in verbose regexps, you should use `[ ]`, or `\ ` or more general `\s`.

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -0,0 +1,241 @@
...
+
+    @classmethod
+    def _correct_deleted_files(cls, rebase_spec_file, files):
+        """Removes files missing in sources from the SPEC file."""
+        for sec_name, sec_content in six.iteritems(rebase_spec_file.spec_content.sections):
+            if sec_name.startswith('%files'):
+                i = 0
+                while i < len(sec_content):
+                    line = sec_content[i]
+                    if line != '':
```
nforro: If you reverse this condition and make the loop continue, you can save one indent unit.

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -0,0 +1,241 @@
...
+                        split_line = line.split(' ')
```
nforro: Not specifying the separator character makes `split()` method split on consecutive whitespaces and trim atuomatically, which I think is better suited for this case.

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -0,0 +1,241 @@
...
+                        directives = []
+                        for element in reversed(split_line):
+                            if element in cls.FILES_DIRECTIVES:
+                                if element == '%doc' or element == '%license':
```
nforro: Since the directives are defined as constants, it doesn't make sense to use some of them directly here. How about adding some property to the constants, specifying their default value?

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -0,0 +1,241 @@
...
+                                    insert_path = MacroHelper.expand(
+                                        '%{_defaultdocdir}' if element == '%doc' else '%{_defaultlicensedir}')
+                                    # expand the files on the line
```
nforro: I don't like this at all. I think this should be done in two passes, first identifying all the directives, gathering the "insert path", and then in a second pass expanding the actual paths.

nforro: And I don't like the "insert path" approach either. Why not simply prepending all the paths found with corresponding macro (`%{_defaultdocdir}` or `%{_defaultlicensedir}`) if necessary, before the expansion?

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -0,0 +1,241 @@
...
+                                directives.insert(0, element)
+                                split_line.remove(element)
+
+                        expanded_line = [MacroHelper.expand(f) for f in split_line]
+
+                        j = 0
+                        while j < len(expanded_line):
+                            file = expanded_line[j]
+                            for deleted_file in reversed(files):
+                                if fnmatch.fnmatch(deleted_file, file):
```
nforro: Same as above, you can test the negative case first, and increment the counter and continue with the loop, and then do the actual work, avoiding `else` altogether. It saves you one indentation level and it is more compact and readable in my opinion.

rebasehelper/specfile.py
```diff
@@ -536,6 +536,9 @@ def write_updated_patches(self, patches, disable_inapplicable):
     # PACKAGE VERSION RELATED METHODS #
     ###################################
 
+    def get_NVR(self):
+        return '{}-{}-{}'.format(self.get_package_name(), self.get_full_version(), self.get_release_number())
```
nforro: You should use `get_release()` instead of `get_release_number()`. Dist tag is part of Release, NVR always contains it (unless it's undefined, but that's special case).

rebasehelper/specfile.py
```diff
@@ -1110,6 +1113,24 @@ def get_prep_section(self):
...
+
+        """
+        parser = SilentArgumentParser()
+        parser.add_argument('-n', default=None)
+        parser.add_argument('-f')
+        parser.add_argument('subpackage', nargs='?', default=None)
+
+        for sec_name in self.spec_content.sections:
+            if sec_name.startswith('%files'):
+                ns, _ = parser.parse_known_args(sec_name.split(' ')[1:])
```
nforro: Same note about `split()` separator argument as above, but I think it's safer to use `shlex.split()` here. Although e.g. arguments containing whitespaces are probably not allowed in a spec file, it's better to be safe than sorry.

rebasehelper/specfile.py
```diff
@@ -1110,6 +1113,24 @@ def get_prep_section(self):
...
+                if not ns.subpackage and not getattr(ns, 'n'):
```
nforro: Is there a problem with using `ns.n`?

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -0,0 +1,257 @@
...
+        # Installed (but unpackaged) file(s) found:
+        #   /usr/bin/pello
+        error_re = re.compile(
+            r'''
+            ^
+            (BUILDSTDERR:)?
+            \s*
+            (
+            (?P<missing>File \s not \s found:\s*)|
+            (?P<unpackaged>Installed \s \(but \s unpackaged\) \s file\(s\) \s found:)
```
nforro: While this is not wrong, it's a little confusing (to me at least). I think you should left out those spaces around `\s`. And since you are matching general whitespace, you could as well use `\s+`.

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -0,0 +1,257 @@
...
+                    for element in reversed(split_line):
+                        if element in cls.FILES_DIRECTIVES:
+                            if cls.FILES_DIRECTIVES[element]:
+                                prepend_macro = cls.FILES_DIRECTIVES[element]
+                            directives.insert(0, element)
+                            split_line.remove(element)
+
+                    if prepend_macro:
+                        split_line = [os.path.join(prepend_macro, file) for file in split_line]
+                    split_line = [MacroHelper.expand(f) for f in split_line]
```
nforro: Why `f` here and `file` above? Single-char variables in list comprehensions are preferred. Also, why not `p` instead of `f`? It's not always a file, but it's always a path.

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -0,0 +1,257 @@
...
+                                j += 1
+                                continue
+
+                            original_file = original_line[len(directives) + j]
+
+                            del split_line[j]
+                            del original_line[len(directives) + j]
+                            files.remove(deleted_file)
+                            result['removed'][sec_name].append(original_file)
+                            logger.info('Removed %s from %s section', original_file, sec_name)
```
nforro: I may be missing something, but shouldn't you increment `j` here?

FrNecas: If I incremented `j`, wouldn't I skip a file? `j` is the index of the file currently inspected, if I delete element on index `j` then the next file which hasn't been inspected yet is still on `j`, isn't it?

nforro: Of course, that's what I was missing :-)

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -0,0 +1,257 @@
...
+        # Installed (but unpackaged) file(s) found:
+        #   /usr/bin/pello
+        error_re = re.compile(
+            r'''
+            ^
+            (BUILDSTDERR:)?
+            \s*
+            (
+            (?P<missing>File\s+not\s+found:\s*)|
+            (?P<unpackaged>Installed\s+\(but\s+unpackaged\)\s+file\(s\)\s+found:)
```
nforro: Just a nitpick, but these two lines could be indented so it's obvious they are in the same group.

rebasehelper/specfile.py
```diff
@@ -1110,6 +1113,24 @@ def get_prep_section(self):
                 result.append(prep.pop(0))
         return result
 
+    def get_main_files_section(self):
+        """Finds the exact name of the main files section.
```
nforro: `of the main %files section` would make more sense.

rebasehelper/output_tools/text_output_tool.py
```diff
@@ -172,6 +175,14 @@ def print_checkers_text_output(cls, checkers_results):
                 if check == check_tool.get_checker_name():
                     logger_report.info('\n'.join(check_tool.format(data)))
 
+    @classmethod
+    def print_build_log_hooks_result(cls, build_log_hooks_result):
+        for build_log_hook in six.itervalues(build_log_hook_runner.build_log_hooks):
+            for hook, data in six.iteritems(build_log_hooks_result):
```
nforro: I don't understand why this loop is here. Can't you simply use `build_log_hooks_result[build_log_hook.get_name()]`?

rebasehelper/build_log_hook.py
```diff
@@ -71,8 +71,12 @@ def run(self, spec_file, rebase_spec_file, **kwargs):
             rebase_spec_file (rebasehelper.specfile.SpecFile): Rebased SpecFile object.
             kwargs (dict): Keyword arguments from instance of Application.
 
+        Returns:
+            bool: Whether build log hooks made some changes to the SPEC file.
+
         """
-        if not kwargs.get('non_interactive'):
+        changes_made = False
+        if not kwargs.get('non_interactive') or kwargs.get('force_build_log_hooks'):
```
nforro: Wouldn't it be better to have these as arguments?

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -0,0 +1,300 @@
...
+        logger.info('Could not remove the following files:')
+        for file in files:
+            logger.info('\t%s', file)
+
+        result['unable_to_remove'] = files
+        return result
+
+    @classmethod
+    def run(cls, spec_file, rebase_spec_file, **kwargs):
+        results_dir = kwargs.get('results_dir')
```
nforro: Same as above, why not have this as an argument? This line is unnecessary.

PR #545, opened 2019-01-16
Update SpecFile._guess_category

FrNecas:


rebasehelper/constants.py
```diff
@@ -49,5 +50,14 @@
 CONFIG_PATH = '$XDG_CONFIG_HOME'
 CONFIG_FILENAME = 'rebase-helper.cfg'
 
+CATEGORIES = {
```
nforro: I think this is a bit too generic, `PACKAGE_CATEGORIES` would be better.

FrNecas: You are right, it will be more self-explanatory. The dict is also referred to on a few places in the code, I will modify the comments as well

PR #560, opened 2019-01-24
Fix build log hooks

FrNecas:


rebasehelper/specfile.py
```diff
@@ -1114,6 +1114,21 @@ def get_prep_section(self):
                 result.append(prep.pop(0))
         return result
 
+    @staticmethod
+    def get_subpackage_name(files_section):
+        """Gets subpackage name based on the %files section."""
+        parser = SilentArgumentParser()
+        parser.add_argument('-n', default=None)
+        parser.add_argument('-f')
+        parser.add_argument('subpackage', nargs='?', default=None)
```
nforro: The same argument parser is used in `SpecFile.get_main_files_section()`. I think it would be better to merge the two methods, or call this one in `SpecFile.get_main_files_section()`.

Also, the `-n` option should be `store_true` action, it just affects how `subpackage` argument is treated.

rebasehelper/specfile.py
```diff
@@ -1114,6 +1114,12 @@ def get_prep_section(self):
                 result.append(prep.pop(0))
         return result
 
+    def get_last_section_line_index(self, section_name):
```
nforro: Hm, how about using this oneliner instead?

```python
[i for i, l in enumerate(self.spec_content.sections[section_name]) if l][-1]
```

rebasehelper/helpers/macro_helper.py
```diff
@@ -38,6 +38,8 @@ class MacroHelper(object):
         '_sysconfdir',
         '_bindir',
         '_libdir',
+        'python2_sitelib',
+        'python3_sitelib',
```
nforro: Is there a reason to add these macros to this position? I think it would be nicer if the whole list was sorted alphabetically.

FrNecas: I don't think there is the reason for the current list ordering. I put python{2,3}_sitelib after _libdir because they are related. But you are right that it may be better to sort alphabetically. Do you suggest taking the starting underscores into consideration or should I ignore them?

nforro: No, please, don't ignore the underscores, that would be confusing (to me at least).

rebasehelper/specfile.py
FrNecas: Should it really be `store_true`? The value of the argument is returned as the full subpackage name so isn't it necessary to use it as a regular argument with `nargs=1`?

nforro: But that's not a value of the `-n` option, rather it's `subpackage` argument.

FrNecas: I just tested it and I don't think that RPM implements it this way. Tested on package `pidgin`:

With `%files -f pidgin.lang -n libpurple` the build passed but after swapping the arguments to `%files -n -f pidgin.lang libpurple`, building SRPM failed with error message `Too many names`.  Based on this fact, I assume that the `-n` option doesn't work as a switch.

nforro: Ok, I stand corrected.

rebasehelper/build_log_hooks/files_build_log_hook.py
```diff
@@ -227,7 +228,7 @@ def _correct_deleted_files(cls, rebase_spec_file, files):
                             split_line.remove(element)
 
                     if prepend_macro:
-                        split_line = [os.path.join(prepend_macro, p) for p in split_line]
+                        split_line = [os.path.join(prepend_macro, package_name, p.split('/')[-1]) for p in split_line]
```
nforro: I think you should use `os.path.basename()` here.

PR #587, opened 2019-03-04
Add spec hook for replacing old version string

### LINKED ISSUE #382: RFE: spec hook for replacing occurences of old version string in a spec file
Example:

```specfile
Summary: Timezone data
Name: tzdata
Version: 2018c
%define tzdata_version 2018c
%define tzcode_version 2018c
Release: 2%{?dist}
License: Public Domain
Group: System Environment/Base
URL: https://www.iana.org/time-zones
Source0: ftp://ftp.iana.org/tz/releases/tzdata%{tzdata_version}.tar.gz
Source1: ftp://ftp.iana.org/tz/releases/tzcode%{tzcode_version}.tar.gz

...
```

FrNecas:
Resolves #382

### DIFF
--- a/rebasehelper/spec_hooks/replace_old_version.py
+++ b/rebasehelper/spec_hooks/replace_old_version.py
@@ -1,0 +1,60 @@
+# -*- coding: utf-8 -*-
+#
+# This tool helps you to rebase package to the latest version
+# Copyright (C) 2013-2014 Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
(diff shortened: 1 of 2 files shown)


rebasehelper/spec_hooks/replace_old_version.py
```diff
@@ -0,0 +1,64 @@
...
+                be used as a substitution.
+
+        Returns:
+            str: Modified line with replaced old version strings.
+
+        """
+        if re.search(r"Version:", line):
+            return line
+
+        comment = re.search(r"#.*", line)
```
nforro: Is there a reason to use regular expressions here at all? What about this?

```python
try:
    comment = line.index('#')
except ValueError:
    comment = None
return line.replace(old, '%{version}' if replace_with_macro else new, line[:comment]) + line[comment:]
```

FrNecas: What about setting comment to len(line) - 1, if no `#` is found? This way, the condition regarding comment wouldn't be necessary

nforro: Good idea, but why `- 1`?

FrNecas: Oh, `-1` shouldn't be there, it is a slice.

rebasehelper/spec_hooks/replace_old_version.py
```diff
@@ -0,0 +1,64 @@
...
+            old (str): Old version string.
+            new (str): New version string.
+            replace_with_macro (bool): Whether %{version} macro should
+                be used as a substitution.
+
+        Returns:
+            str: Modified line with replaced old version strings.
+
+        """
+        if re.search(r"Version:", line):
```
nforro: Is this check really necessary? Isn't the version already bumped in `rebase_spec_file`?

Issue #465, opened 2018-03-09
Allow plugins to define their own arguments


nforro:
For example `--get-old-build-from-koji` could (should) be part of `koji` buildtool.

PR #590, opened 2019-03-05
Allow plugins to define their own arguments

FrNecas:
Resolves: #465 

Rewrote docstrings for the run methods. We should also consider moving base plugin classes somewhere else. Some files (such as `build_log_hook.py` or `versioneer.py`) now contain only the base class and nothing else. I don't think that moving all of them to `plugins.py` would be a good idea, the module would become too long.


rebasehelper/plugins.py
```diff
@@ -46,5 +63,175 @@ def load(cls, entrypoint):
...
+class PluginRunner(object):
+    PLUGIN_ENTRYPOINTS = {
+        'build_tools': 'rebasehelper.build_tools',
+        'srpm_build_tools': 'rebasehelper.srpm_build_tools',
+        'checkers': 'rebasehelper.checkers',
+        'spec_hooks': 'rebasehelper.spec_hooks',
+        'build_log_hooks': 'rebasehelper.build_log_hooks',
+        'versioneers': 'rebasehelper.versioneers',
+        'output_tools': 'rebasehelper.output_tools',
+    }
```
nforro: This mapping is completely unnecessary, you can use `pkg_resources.get_entry_map('rebasehelper')` to get a dictionary of all groups and their entrypoints. Then you could just strip `rebasehelper.` prefix and use that as an attribute name.

rebasehelper/plugins.py
```diff
@@ -20,16 +20,33 @@
...
 class Plugin(object):
     name = None
+    runner = None
 
 
 class PluginLoader(object):
     @classmethod
-    def load(cls, entrypoint):
+    def load(cls, entrypoint, runner):
+        """Loads an entrypoint from setup.py.
```
nforro: This is not exactly true, entrypoints could be defined in `setup.cfg` for example (and perhaps that would be a smart thing to do).

nforro:
> Some files (such as `build_log_hook.py` or `versioneer.py`) now contain only the base class and nothing else.

I don't see that as a problem.

Actually, I think `PluginRunner` should have one generic `run_X` method and specific implementation should stay in specific runner classes.

rebasehelper/cli.py
```diff
@@ -48,7 +49,7 @@ def build_parser(available_choices_only=False):
 
         group = parser.add_mutually_exclusive_group()
         current_group = 0
-        for option in traverse_options(OPTIONS):
+        for option in traverse_options(OPTIONS + plugin_runner.get_options()):
```
nforro: I don't like the name, I think `plugin_manager` would be more fitting, as it does more than just run plugins.

rebasehelper/plugins/checkers/csmock.py
```diff
@@ -31,10 +31,10 @@
 from rebasehelper.helpers.path_helper import PathHelper
 from rebasehelper.exceptions import CheckerNotFoundError
 from rebasehelper.results_store import results_store
-from rebasehelper.checker import BaseChecker
+from rebasehelper.plugins.checkers import BaseChecker
 
 
-class CsmockTool(BaseChecker):
+class Csmock(BaseChecker):
```
nforro: I think the name should be `CsMock`, to be consistent with the rest of classes.

PR #598, opened 2019-03-15
Add method for updating changelog

FrNecas:


rebasehelper/specfile.py
```diff
@@ -1269,6 +1269,12 @@ def get_requires(self):
         """
         return [r.decode(constants.DEFENC) if six.PY3 else r for r in self.hdr[rpm.RPMTAG_REQUIRES]]
 
+    def update_changelog(self):
```
TomasTomecek: I would actually expect to pass my changelog entries here.

FrNecas: Would you prefer to pass the whole changelog entry to the method (so a list containing all information, the date, author, message, etc.) or just the message of the entry?

The former would look like so:
```python
new_entry = specfile.get_new_log('Your Message') # you could also create this entry in a different way
specfile.update_changelog(new_entry)
```

The latter like so (`specfile.update_changelog` would obviously internally call `get_new_log`.):
```python
specfile.update_changelog("Your Message")
```

The arguments with changelog entry could default to SpecFile.changelog_entry if it isn't explicitly set.

TomasTomecek:
how about a new method:
```python
def add_changelog_item(self, message) :
```
this would suit our use case the most

PR #600, opened 2019-03-19
Refactor how adding changelog entries works

FrNecas:
Changelog_entry isn't now passed into the constructor of a SpecFile which should make it more logical to use.

Any objections @TomasTomecek , does this suit your use-case better?

### DIFF
--- a/rebasehelper/specfile.py
+++ b/rebasehelper/specfile.py
@@ -199,11 +199,10 @@ class SpecFile(object):
     prep_section = []
     removed_patches = []
 
-    def __init__(self, path, changelog_entry, sources_location='', download=True):
+    def __init__(self, path, sources_location='', download=True):
         self.path = path
         self.download = download
         self.sources_location = sources_location
-        self.changelog_entry = changelog_entry
         #  Read the content of the whole SPEC file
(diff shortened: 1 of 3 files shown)


rebasehelper/specfile.py
```diff
@@ -1271,13 +1270,27 @@ def get_requires(self):
         """
         return [r.decode(constants.DEFENC) if six.PY3 else r for r in self.hdr[rpm.RPMTAG_REQUIRES]]
 
-    def update_changelog(self):
-        """Inserts new entry into the changelog and saves the SpecFile."""
-        new_entry = self.get_new_log()
+    def update_changelog(self, changelog_entry):
```
TomasTomecek: Yup, this is exactly what we were looking for.

FrNecas: I am going to merge this. I may refactor the class even further (passing `download` argument to the constructor doesn't make sense for packit nor for rebase-helper). If I make any changes to the methods that packit uses, I will create a PR in packit to take the changes into account.

TomasTomecek: Awesome! thank you.

PR #603, opened 2019-03-27
Do not use dict to store spec sections

nforro:
Sections in a spec file can be enclosed in conditions, which means they can be duplicate.
Use list of tuples instead of dict to store them.

### DIFF
--- a/rebasehelper/specfile.py
+++ b/rebasehelper/specfile.py
@@ -37,7 +37,6 @@ from datetime import date
 from difflib import SequenceMatcher
 from operator import itemgetter
 
-from requests.structures import CaseInsensitiveDict
 from six.moves import urllib
 
 from rebasehelper.logger import logger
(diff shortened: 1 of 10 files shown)


rebasehelper/specfile.py
```diff
@@ -147,13 +146,49 @@ def __init__(self, content):
...
         content = []
-        for header, section in six.iteritems(self.sections):
+        for header, section in self.sections:
             if header != '%package':
                 content.append(header + '\n')
             for line in section:
                 content.append(line + '\n')
         return ''.join(content)
 
+    def section(self, name):
```
FrNecas: Shouldn't this rather be named `get_section`?

nforro: It certainly could, but it is supposed to be a drop-in replacement for the current `spec_content.section[NAME]`.

rebasehelper/specfile.py
```diff
@@ -147,13 +146,49 @@ def __init__(self, content):
...
+        """Replaces content of a section.
+
+        In case there are multiple sections with the same name, the first one is replaced.
+
+        Args:
+            name (str): Section name.
+            content (list): Section content as a list of lines.
+
+        Returns:
+            bool: False if section was not found else True
```
FrNecas: Just a nitpick - there should be a dot at the end of the line

rebasehelper/specfile.py
FrNecas: You are right that, moreover `get_section` would be a bit too long with our use-case in mind

PR #607, opened 2019-04-03
Fix replacement of old version string

FrNecas:


rebasehelper/spec_hooks/replace_old_version.py
```diff
@@ -20,12 +20,23 @@
...
 
 class ReplaceOldVersionSpecHook(BaseSpecHook):
     """SpecHook for replacing occurrences of old version string."""
 
+    @classmethod
+    def _is_local_source(cls, line):
+        if not (line.startswith('Patch') or line.startswith('Source')):
+            return False
+        source = line.split()[1]
+        if bool(urllib.parse.urlparse(source).scheme):
```
nforro: `bool()` is unnecessary here.

rebasehelper/spec_hooks/replace_old_version.py
```diff
@@ -20,12 +20,23 @@
 # Authors: Petr Hracek <phracek@redhat.com>
 #          Tomas Hozza <thozza@redhat.com>
 
+import urllib.parse
```
nforro: Does this work with both python2 and python3?

FrNecas: It didn't work for python2, should be fixed now

PR #620, opened 2019-05-22
Enable replacement of old version with macro

FrNecas:


rebasehelper/tests/test_specfile.py
```diff
@@ -289,14 +289,25 @@ def test_escape_macros_spec_hook(self, spec_object):
         EscapeMacros.run(spec_object, spec_object)
         assert spec_object.spec_content.section('%build')[0] == "autoreconf -vi # Unescaped macros %%name %%{name}"
 
-    def test_replace_old_version_spec_hook(self, spec_object):
+    @pytest.mark.parametrize('replace_with_macro', [
+        (True,),
+        (False,),
```
nforro: Are tuples really needed here?

rebasehelper/application.py
```diff
@@ -98,6 +98,8 @@ def __init__(self, cli_conf, execution_dir, results_dir, debug_log_file):
 
         self.kwargs['changelog_entry'] = self.conf.changelog_entry
 
+        self.kwargs['replace_old_version_with_macro'] = self.conf.replace_old_version_with_macro
```
nforro: Having to do this defeats the purpose of defining options in plugins. `self.kwargs` should be populated with `self.conf` attibutes automatically, or `self.conf` should be passed to plugins.

rebasehelper/application.py
```diff
@@ -88,21 +88,11 @@ def __init__(self, cli_conf, execution_dir, results_dir, debug_log_file):
...
-
-        self.kwargs['changelog_entry'] = self.conf.changelog_entry
-
-        self.kwargs['spec_hook_blacklist'] = self.conf.spec_hook_blacklist
-
-        self.kwargs['build_log_hook_blacklist'] = self.conf.build_log_hook_blacklist
-
-        self.kwargs['force_build_log_hooks'] = self.conf.force_build_log_hooks
+        for option, value in self.conf.config.items():
+            self.kwargs[option] = value
```
nforro: How about `self.kwargs.update(self.conf.config)`?

PR #623, opened 2019-05-29
Fix regex for finding comments in Spec file

FrNecas:


rebasehelper/specfile.py
```diff
@@ -147,6 +147,16 @@ class SpecContent:
         '%transfiletriggerpostun',
     ]
 
+    # Comments in these sections can only be on a separate line.
+    BLACKLIST_COMMENT_IN: List[str] = [
```
nforro: I don't like the name. The same goes for `comment_allowed` variable below. I don't think it's exact and the meaning isn't obvious. Let's define the term **inline comment** for a comment that doesn't start at the beginning of a line. Then you can name this attribute e.g. `DISALLOW_INLINE_COMMENTS` and the variable `inline_comment_allowed`.

rebasehelper/specfile.py
```diff
@@ -160,6 +170,22 @@ def __str__(self):
...
+    def get_comment_span(cls, line: str, section: str) -> Tuple[int, int]:
+        """Gets span of a comment depending on the section.
+
+        Args:
+            line: Line to find the comment in.
+            section: Section the line is in.
+
+        Returns:
+            Span of the comment
+
```
nforro: I think the docstring should explain what (and why) is returned when no comment is found.

nforro: Just a nitpick, I would replace
`to length of line to make it convenient to use the result in a slice`
with
`to the length of the line for convenient use in a slice`.

PR #624, opened 2019-05-30
Restructure tests

FrNecas:
Restructure tests to correspond with the changes to plugins. Also split `test_helpers` into multiple modules to be consistent


rebasehelper/tests/test_specfile.py
```diff
@@ -30,18 +30,14 @@
...
-from rebasehelper.plugins.spec_hooks.escape_macros import EscapeMacros
-from rebasehelper.plugins.spec_hooks.replace_old_version import ReplaceOldVersion
-from rebasehelper.plugins.spec_hooks.paths_to_rpm_macros import PathsToRPMMacros
+from rebasehelper.tests.conftest import SPEC_FILE
 
 
 class TestSpecFile:
     NAME: str = 'test'
     VERSION: str = '1.0.2'
+    SPEC_FILE = SPEC_FILE
```
nforro: This is missing a type hint.

rebasehelper/tests/conftest.py
```diff
@@ -27,9 +27,18 @@
...
 
 TESTS_DIR: str = os.path.dirname(__file__)
 TEST_FILES_DIR: str = os.path.join(TESTS_DIR, 'testing_files')
+SPEC_FILE: str = 'test.spec'
+
+
+@pytest.fixture
+def spec_object(workdir):  # pylint: disable=redefined-outer-name
+    sf = SpecFile(SPEC_FILE, workdir)
+    return sf
```
nforro: I think it would be more convenient if the fixture would copy the specfile to `workdir`, to avoid the need to put it in `TEST_FILES`.

rebasehelper/tests/conftest.py
```diff
@@ -27,9 +27,20 @@
...
 
 TESTS_DIR: str = os.path.dirname(__file__)
 TEST_FILES_DIR: str = os.path.join(TESTS_DIR, 'testing_files')
+SPEC_FILE: str = 'test.spec'
+
+
+@pytest.fixture
+def spec_object(workdir):  # pylint: disable=redefined-outer-name
+    # copy Spec file into workdir
+    shutil.copy(os.path.join(TEST_FILES_DIR, SPEC_FILE), os.getcwd())
```
nforro: Why `os.getcwd()` instead of `workdir`?

rebasehelper/tests/test_specfile.py
```diff
@@ -30,18 +30,14 @@
...
-from rebasehelper.plugins.spec_hooks.escape_macros import EscapeMacros
-from rebasehelper.plugins.spec_hooks.replace_old_version import ReplaceOldVersion
-from rebasehelper.plugins.spec_hooks.paths_to_rpm_macros import PathsToRPMMacros
+from rebasehelper.tests.conftest import SPEC_FILE
 
 
 class TestSpecFile:
     NAME: str = 'test'
     VERSION: str = '1.0.2'
+    SPEC_FILE: str = SPEC_FILE
```
nforro: This feels really unnecessary, imported `SPEC_FILE` can be used directly, can it not?

FrNecas: You are right. Now that I am looking at it, `TestSpecFile.SPEC_FILE` is only used in one test, it will definitely be better to use `SPEC_FILE` directly

PR #625, opened 2019-06-06
Refactor save/read SpecFile workflow

FrNecas:


rebasehelper/specfile.py
```diff
@@ -62,6 +62,15 @@ def get_rebase_name(dir_name, name):
     return os.path.join(dir_name, file_name)
 
 
+def saves(func):
```
nforro: Wouldn't it make more sense if the decorator was defined directly above the `SpecFile` class?

PR #626, opened 2019-06-06
Refactor class attributes in specfile module

FrNecas:


rebasehelper/specfile.py
```diff
@@ -1329,21 +1313,13 @@ def is_test_suite_enabled(self):
...
-        return RpmHelper.decode(self.hdr[rpm.RPMTAG_NAME])
-
-    def get_requires(self):
-        """
-        Function returns a package requirements
-
-        :return:
-        """
-        return [RpmHelper.decode(r) for r in self.hdr[rpm.RPMTAG_REQUIRES]]
+    def get_requires(self) -> List:
```
nforro: Shouldn't this be `List[str]`?

Issue #629, opened 2019-06-19
files build log hook should ignore macros in %files


nforro:
Macros like `%global` or `%if` are valid in context of `%files` section, `files` hook should be able to deal with them.

See [tlog](https://src.fedoraproject.org/rpms/tlog/blob/master/f/tlog.spec#_80) for example.

PR #630, opened 2019-06-24
Files build log hook fixes

FrNecas:
- handle %global and %if keywords correctly
- follow Fedora Packaging Guidelines when adding manpages

Resolves: #629

### DIFF
--- a/rebasehelper/plugins/build_log_hooks/files.py
+++ b/rebasehelper/plugins/build_log_hooks/files.py
@@ -28,7 +28,7 @@ import fnmatch
 import os
 import re
 
-from typing import Dict, Optional
+from typing import Dict, List, Optional
 
 from rebasehelper.plugins.build_log_hooks import BaseBuildLogHook
 from rebasehelper.types import PackageCategories
@@ -60,6 +60,12 @@ class Files(BaseBuildLogHook):
         '%readme': None,
         '%verify': None,
     }
+
+    PROHIBITED_KEYWORDS: List[str] = [
+        '%if',
+        '%else',
+        '%endif',
+    ]
 
     @classmethod
     def format(cls, data):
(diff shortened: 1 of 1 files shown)


rebasehelper/plugins/build_log_hooks/files.py
```diff
@@ -191,6 +191,17 @@ def _get_best_matching_files_section(cls, rebase_spec_file, file):
...
         return best_match_section or rebase_spec_file.get_main_files_section()
 
+    @classmethod
+    def _sanitize_path(cls, path):
+        """Changes the path to follow Fedora Packaging Guidelines."""
+        if path.startswith('%{_mandir}'):
+            manpage_start = path.rfind('/')
+            # find 2nd occurrence of '.', replace everything after it
+            # with * wildcard.
+            extension_start = path[manpage_start:].index('.', path[manpage_start:].index('.') + 1)
```
nforro: This won't work with uncompressed man pages.

rebasehelper/plugins/build_log_hooks/files.py
```diff
@@ -191,6 +191,17 @@ def _get_best_matching_files_section(cls, rebase_spec_file, file):
 
         return best_match_section or rebase_spec_file.get_main_files_section()
 
+    @classmethod
+    def _sanitize_path(cls, path):
+        """Changes the path to follow Fedora Packaging Guidelines."""
+        if path.startswith('%{_mandir}'):
+            manpage_start = path.rfind('/')
```
nforro: What's wrong with `os.path.basename()`? Or `os.path.split()`, as you need the ret of the path.

rebasehelper/plugins/build_log_hooks/files.py
```diff
@@ -191,6 +191,17 @@ def _get_best_matching_files_section(cls, rebase_spec_file, file):
...
+            # find 2nd occurrence of '.', replace everything after it
+            # with * wildcard.
+            extension_start = path[manpage_start:].index('.', path[manpage_start:].index('.') + 1)
+            path = path[:manpage_start+extension_start] + '*'
```
nforro: Something like this should be sufficient:

```python
dir, name = os.path.split(path)
return os.path.join(dir, name.split('.')[:2] + '*')
```

FrNecas: Thanks. Would you keep the separate function for doing this? There could be more path modifications that we are missing.

nforro: Sure, why not.

PR #631, opened 2019-06-24
Replace old version for new only in whole words

FrNecas:


rebasehelper/plugins/spec_hooks/replace_old_version.py
```diff
@@ -73,7 +74,9 @@ def _replace(cls, line, old, new, replace_with_macro=False):
         """
         if cls._is_local_source(line):
             return line
-        return line.replace(old, '%{version}' if replace_with_macro else new)
+        pattern = re.compile(r'([/\-\s]){}([/.\-\s])'.format(re.escape(old)))
```
nforro: This could be compiled outside the loop (old version is already known) to improve performance.

rebasehelper/plugins/spec_hooks/replace_old_version.py
```diff
@@ -57,35 +60,37 @@ def _is_local_source(cls, line):
...
         new_version = rebase_spec_file.get_version()
         replace_with_macro = kwargs.get('replace_old_version_with_macro')
+        pattern = re.compile(r'([/\-\s]){}([/.\-\s])'.format(re.escape(old_version)))
         for sec_name, section in rebase_spec_file.spec_content.sections:
             if sec_name.startswith('%changelog'):
                 continue
             for index, line in enumerate(section):
                 start, end = spec_file.spec_content.get_comment_span(line, sec_name)
-                updated_line = cls._replace(line[:start], old_version, new_version, replace_with_macro)
+                updated_line = cls._replace(line[:start], pattern, new_version, replace_with_macro)
```
nforro: Does it still make sense to have extra method for this?

PR #632, opened 2019-06-27
Use rpm.reloadConfig() before parsing SPEC

nforro:
`rpm.reloadConfig()`, among other things, resets all macros to default values and removes previously defined macros. That should prevent issues with polluted global macro namespace when parsing multiple different SPEC files.

### DIFF
--- a/rebasehelper/tests/test_specfile.py
+++ b/rebasehelper/tests/test_specfile.py
@@ -231,6 +231,8 @@ class TestSpecFile:
         assert spec_object.get_extra_version() == 'rc1'
 
     def test_update_setup_dirname(self, spec_object):
+        spec_object.set_extra_version('rc1')
+
         prep = spec_object.spec_content.section('%prep')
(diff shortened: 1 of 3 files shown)


rebasehelper/tests/test_specfile.py
```diff
@@ -231,6 +231,8 @@ def test_get_extra_version_set(self, spec_object):
         assert spec_object.get_extra_version() == 'rc1'
 
     def test_update_setup_dirname(self, spec_object):
+        spec_object.set_extra_version('rc1')
```
FrNecas: Wouldn't it make more sense to set extra version below line `assert spec_object.spec_content.section('%prep') == prep` since extra version doesn't influence the first call of `update_setup_dirname`?

nforro: I don't think it would, this is basically part of setup phase of this test, and as such, it should be at the beginning.

PR #635, opened 2019-07-04
Make optionality of checkers work again

FrNecas:


rebasehelper/plugins/checkers/__init__.py
```diff
@@ -123,3 +123,10 @@ def run(self, results_dir, checker_name, **kwargs):
 
         logger.info("Running checks on packages using '%s'", checker_name)
         return checker.run_check(results_dir, **kwargs)
+
+    def get_supported_plugins(self) -> List[BaseChecker]:
```
nforro: I don't think the return type is list of BaseChecker instances. Shouldn't it rather be `List[Type[BaseChecker]]`?

nforro: Also, since you are overriding methods of base class, I think you should annotate those too.

PR #637, opened 2019-07-04
Format output lists in a more readable way

FrNecas:
Printing lists or dict keys is not very user friendly (`Supported archives are dict_keys(['.tar.xz', ...])` vs `Supported archives are .tar.xz, .bz2, ...`). Also fix a bug in Application, where an incorrect method of PluginCollection was called to get supported tools.

### DIFF
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -352,7 +352,7 @@ class Application:
             archive = Archive(archive_path)
         except NotImplementedError as e:
             raise RebaseHelperError('{}. Supported archives are {}'.format(str(e),
-                                    Archive.get_supported_archives()))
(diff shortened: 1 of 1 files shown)


rebasehelper/application.py
```diff
@@ -352,7 +352,7 @@ def extract_archive(archive_path, destination):
             archive = Archive(archive_path)
         except NotImplementedError as e:
             raise RebaseHelperError('{}. Supported archives are {}'.format(str(e),
-                                    Archive.get_supported_archives()))
+                                    ", ".join(Archive.get_supported_archives())))
```
nforro: Any reason to use `"`, especially since `'` are used just above?

Issue #597, opened 2019-03-15
.get_full_version() shows commit hash twice


TomasTomecek:
Spec:
```
Version:        0.0.3.11.gc9956c9
Release:        1%{?dist}
```

Code:
```
print(rebasehelper.SpecFile(path).get_full_version())
```

Output:
```
'0.0.3.11.gc9956c9gc9956c9'
```

rebase-helper outputs the commit hash twice, even though it's there only once

Tarballs and SRPM generated afterwards:
```
-rw-rw-r--. 1 tt tt  19K Mar 15 12:47 ogr-0.0.3.11.gc9956c9.tar.gz
-rw-rw-r--. 1 tt tt  26K Mar 15 12:47 python-ogr-0.0.3.11.gc9956c9-1.fc29.src.rpm

```

nforro:
Looking at the code, it seems to me `SpecFile.get_full_version()` method is useless and nonsensical. I'm going to remove it.
@TomasTomecek you probably want to use `SpecFile.get_version()` instead.

TomasTomecek:
lol :D oki, will do, thank you

PR #639, opened 2019-07-08
Minor SpecFile changes

FrNecas:
Resolves: #597


rebasehelper/application.py
```diff
@@ -538,7 +538,7 @@ def build_source_packages(self):
             spec = self.spec_file if version == 'old' else self.rebase_spec_file
             package_name = spec.get_package_name()
             package_version = spec.get_version()
-            package_full_version = spec.get_full_version()
+            package_full_version = spec.get_version()
```
nforro: I don't think there is a reason to have both `package_version` and `package_full_version` if they are the same.

PR #646, opened 2019-07-12
Use local builder when the build can't be downloaded from koji

FrNecas:
In `golang-github-google-cmp` rebase from 0.2.0 to 0.3.0, the old build wasn't in koji, whereas the new one was. rebase-helper couldn't handle such thing. What we should also consider is changing `KojiHelper.get_old_build_info()`. Currently if the old version build can't be obtained, it fallbacks to getting the latest build. Since this method is used only when `--get-old-build-from-koji` is active, it imho doesn't make much sense (we want the old build or nothing, getting the info about new build is pointless). What's your opinion @nforro ? Is there any case, where the fallback could be useful?

### DIFF
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -541,7 +541,9 @@ class Application:
             logger.info('Building source package for %s version %s', package_name, package_version)
 
             if version == 'old' and self.conf.get_old_build_from_koji:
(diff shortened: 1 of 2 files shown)


nforro:
The reasoning behind the fallback is that due to Koji garbage collection, there is a high chance a specific version build doesn't exist anymore. But I agree that it make more sense to fallback to local build instead.

PR #650, opened 2019-07-17
Fix patch strip parsing

FrNecas:


rebasehelper/specfile.py
```diff
@@ -440,8 +441,10 @@ def _get_patch_strip_options(self, patches):
             rest = [os.path.basename(a) for a in rest]
             indexes = [p[1] for p in patches if p[0] in rest]
             for idx in indexes:
-                result[idx] = ns.p
-        return result
+                result[idx].append(ns.p)
```
nforro: I think the following simple change would be sufficient here:

```python
if idx not in result or result[idx] < ns.p:
    result[idx] = ns.p
```

FrNecas: Good point, my solution was pointlessly complicated :+1:

PR #651, opened 2019-07-17
Ignore debug information when manipulating %files

FrNecas:


rebasehelper/plugins/build_log_hooks/files.py
```diff
@@ -157,7 +157,9 @@ def _parse_build_log(cls, log_path, nvr):
                 elif match.group('unpackaged'):
                     error_type = 'missing'
                 elif error_type == 'missing' and match.group('path'):
-                    files.add(match.group('path'))
+                    # Ignore debug information
+                    if '/usr/lib/debug' not in match.group('path'):
```
nforro: I think it would be safer to use `startswith()`.

PR #652, opened 2019-07-17
Fix listing default checkers

nforro:


FrNecas:
I recently changed this one because optionality of checkers didn't work and rebase-helper would fail if a checker wasn't installed. Where did the problem arise? Does it make sense to have a default checker that is not available?

nforro:
Among other things, [here](https://rebase-helper.readthedocs.io/en/latest/user_guide/usage.html#id14).

FrNecas:
I just tested the bugfix and it indeed breaks checkers optionality. Uninstalling licensecheck made rebase-helper traceback due to a missing /usr/bin/licensecheck. I assume it may be necessary to do a similar workaround as we had to do with choices (all x available).

Issue #644, opened 2019-07-10
Spec hook for replacing only parts of hardcoded URLs


FrNecas:
See https://src.fedoraproject.org/rpms/gnome-calendar/blob/master/f/gnome-calendar.spec#_14

In the previous version, the Source contained 32 instead of 33. ReplaceOldVersionSpecHook can't handle this because it's only a part of the version

PR #657, opened 2019-07-19
Try to replace hardcoded version parts in Sources and Patches

FrNecas:
Fixes: #644

### DIFF
--- a/rebasehelper/plugins/spec_hooks/replace_old_version.py
+++ b/rebasehelper/plugins/spec_hooks/replace_old_version.py
@@ -25,8 +25,11 @@
 import re
 import urllib.parse
 
+from typing import Any, List, Pattern, Tuple
+
 from rebasehelper.types import Options
 from rebasehelper.plugins.spec_hooks import BaseSpecHook
+from rebasehelper.specfile import SpecFile
 
 
 class ReplaceOldVersion(BaseSpecHook):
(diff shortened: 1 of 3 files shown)


rebasehelper/plugins/spec_hooks/replace_old_version.py
```diff
@@ -58,22 +61,54 @@ def _is_local_source(cls, line: str) -> bool:
...
+        replacement = replacement_pattern.format('%{version}' if replace_with_macro else new_version)
+
         for sec_name, section in rebase_spec_file.spec_content.sections:
             if sec_name.startswith('%changelog'):
                 continue
             for index, line in enumerate(section):
                 # special case Version tag to avoid mistakes when version format changes
                 # e.g. in a rebase from 2.5 to 2.5.1, version would be changed to 2.5.1.1
-                if cls._is_local_source(line) or line.startswith('Version:'):
+                if cls._is_local_source(line) or line.startswith('Version'):
```
nforro: What's the reason for this?

rebasehelper/plugins/spec_hooks/replace_old_version.py
```diff
@@ -58,22 +61,54 @@ def _is_local_source(cls, line: str) -> bool:
...
+    def run(cls, spec_file: SpecFile, rebase_spec_file: SpecFile, **kwargs: Any):
         old_version = spec_file.get_version()
         new_version = rebase_spec_file.get_version()
         replace_with_macro = kwargs.get('replace_old_version_with_macro')
-        pattern = re.compile(r'([/\-\s]){}([/.\-\s])'.format(re.escape(old_version)))
+
+        replacement_pattern = r'\g<1>{}\g<2>'
+        subversion_patterns = cls._create_possible_replacements(old_version, new_version, replacement_pattern)
+
+        main_pattern = re.compile(r'([/\-\s]){}([/.\-\s])'.format(re.escape(old_version)))
```
nforro: This shouldn't be special-cased, it should be part of the dictionary of patterns.

rebasehelper/plugins/spec_hooks/replace_old_version.py
```diff
@@ -58,22 +61,54 @@ def _is_local_source(cls, line: str) -> bool:
...
+        return res
+
+    @classmethod
+    def run(cls, spec_file: SpecFile, rebase_spec_file: SpecFile, **kwargs: Any):
         old_version = spec_file.get_version()
         new_version = rebase_spec_file.get_version()
         replace_with_macro = kwargs.get('replace_old_version_with_macro')
-        pattern = re.compile(r'([/\-\s]){}([/.\-\s])'.format(re.escape(old_version)))
+
+        replacement_pattern = r'\g<1>{}\g<2>'
```
nforro: Is it really necessary to define this outside of `_create_possible_replacements()`?

rebasehelper/plugins/spec_hooks/replace_old_version.py
FrNecas: For consistency. I've checked the rest of the code and from what I've seen that colon is never used when looking for particular preamble tags.

rebasehelper/plugins/spec_hooks/replace_old_version.py
FrNecas: I've considered this, the question is - do we want to try to replace subversions everywhere, not just in remote Sources/Patches? Couldn't there be any drawbacks? With this implementation, the full version is replaced everywhere (except for `%changelog`) and the version substrings only in Sources and Patches.

Regarding the next comment - if we opted for having the main replacement pattern in the dictionary, then it would be possible to define the replacement in `_create_pssible_replacements()`.

rebasehelper/plugins/spec_hooks/replace_old_version.py
nforro: Ok.

rebasehelper/plugins/spec_hooks/replace_old_version.py
nforro: That's a good point, but for example you can return a list of (pattern, replacement) tuples, with the first one always being the entire version.

FrNecas: Good idea, that should be more understandable, because all replacements will be created in one place.

PR #659, opened 2019-07-24
Use branch of GitPython where kwargs transformation works

FrNecas:


setup.py
```diff
@@ -51,7 +51,7 @@ def get_requirements():
         'pyquery',
         'python-pam',
         'requests',
-        'GitPython',
+        'GitPython@git+https://github.com/FrNecas/GitPython.git@frnecas-kwargs',
```
nforro: How about adding a link to the pull request, just like with ansicolors?

PR #660, opened 2019-07-24
Try to fallback to x86_64 if log for noarch is not found

FrNecas:


nforro:
I don't like this very much. I think we should reimplement this entirely and use `session.listTaskOutput()` to download logs associated with a task that belongs to a specific build.

FrNecas:
You are right that it could use some improvement. By using `session.listTaskOutput()` we wouldn't have to hardcode the log names either. The problem is somewhat related to #493 as well

PR #661, opened 2019-07-25
Download remote sources into correct directory

FrNecas:


rebasehelper/helpers/lookaside_cache_helper.py
```diff
@@ -112,14 +112,14 @@ def _download_source(cls, tool, url, package, filename, hashtype, hsh, target=No
             raise LookasideCacheError(str(e))
 
     @classmethod
-    def download(cls, tool, basepath, package):
+    def download(cls, tool, basepath, package, target=None):
```
nforro: If you are adding support for downloading sources elsewhere, you should also add support for uploading them from elsewhere.

FrNecas: That is already supported, isn't it? `LookasideCacheHelper._upload_source()` has an argument `filename` which can be both relative and absolute.

nforro: No, there are never absolute paths in `sources` file, and there shouldn't be.

nforro: Maybe I should rather say, there are no paths in `sources`, only filenames.

rebasehelper/helpers/lookaside_cache_helper.py
```diff
@@ -176,7 +177,7 @@ def post(check_only=False):
             sys.stdout.flush()
 
     @classmethod
-    def update_sources(cls, tool, basepath, package, old_sources, new_sources, upload=True):
+    def update_sources(cls, tool, basepath, package, old_sources, new_sources, upload=True, target=''):
```
nforro: `target`, really? That doesn't sound right to me at all.

FrNecas: You are right. What about `source_directory`? Or do you have something better in your mind?

nforro: That's much better.

rebasehelper/helpers/lookaside_cache_helper.py
```diff
@@ -163,6 +164,7 @@ def post(check_only=False):
                 raise LookasideCacheError(r.reason)
             return r.content
 
+        filename = os.path.join(source_dir, filename)
```
nforro: I'd prefer to create a new variable instead of overwriting the existing one.

rebasehelper/helpers/lookaside_cache_helper.py
```diff
@@ -122,7 +122,8 @@ def download(cls, tool, basepath, package, target=None):
             cls._download_source(tool, url, package, source['filename'], source['hashtype'], source['hash'], target)
 
     @classmethod
-    def _upload_source(cls, url, package, filename, hashtype, hsh, auth=requests_gssapi.HTTPSPNEGOAuth()):
+    def _upload_source(cls, url, package, filename, hashtype, hsh, auth=requests_gssapi.HTTPSPNEGOAuth(),
+                       source_dir=''):
```
nforro: I think it would make more sense if `source_dir` was positioned before `filename` argument.

FrNecas: It definitely would but we would have to remove the default value as non-default arguments can't follow a default argument.

nforro: Sure, is that a problem?

PR #665, opened 2019-07-30
Fallback to copying when an archive can't be extracted

FrNecas:


rebasehelper/application.py
```diff
@@ -364,7 +360,14 @@ def extract_archive(archive_path, destination):
...
     def extract_sources(archive_path, destination):
         """Function extracts a given Archive and returns a full dirname to sources"""
-        Application.extract_archive(archive_path, destination)
+        try:
+            Application.extract_archive(archive_path, destination)
+        except NotImplementedError:
+            # not a standard archive type, can't extract it, fallback to copying
+            base = os.path.basename(archive_path)
+            archive_dir = os.path.join(destination, os.path.splitext(base)[0])
+            os.makedirs(archive_dir)
```
nforro: Is making a directory really necessary?

FrNecas: It's necessary to create just the `destination` directory which hasn't been created inside `rebase-helper-workspace` yet. Fixed

PR #671, opened 2019-08-01
Release 0.17.1

nforro:


rebase-helper.spec
```diff
@@ -98,8 +98,8 @@ PYTHONPATH=$(pwd) py.test-3 -v %{pkgname}
 
 
 %changelog
-* Wed Jul 31 2019 Nikola Forró <nforro@redhat.com> - 0.17.0-1
```
FrNecas: Shouldn't this stay in the changelog as well?

nforro: Why? It was never published.

FrNecas: Ah, thanks

PR #672, opened 2019-08-01
Test Application more thoroughly

FrNecas:


FrNecas:
Ready for review - I think this solution is a bit cleaner as it avoids a lot of code duplication

rebasehelper/tests/test_application.py
```diff
@@ -67,7 +67,23 @@ class TestApplication:
 
     cmd_line_args: List[str] = ['--not-download-sources', '1.0.3']
 
-    def test_application_sources(self, workdir):
+    @pytest.fixture
+    def config(self):
```
nforro: The name is confusing, as the method doesn't return instance of `Config`. It should be called `make_config()` or so.

rebasehelper/tests/test_application.py
```diff
@@ -112,16 +128,48 @@ def test_application_sources(self, workdir):
...
 
+    def test_setup(self, config):
+        config = config(self.cmd_line_args)
+        files = Application.setup(config)
+        for file in files:
+            assert os.path.exists(file)
+
+    def test_setup_continue(self, config):
+        config = config(self.cmd_line_args + ['--continue'])
+        files = Application.setup(config)
```
nforro: I don't like this. You are already testing this in the previous test, so the check below is redunant. I also don't like referring to `results_dir` as `files[1]`.

rebasehelper/tests/test_application.py
```diff
@@ -112,16 +128,48 @@ def test_application_sources(self, workdir):
...
+            assert os.path.exists(file)
+
+        # check that if rebase-helper-results exists, it does not get removed
+        test_file = os.path.join(files[1], 'test')
+        open(test_file, 'w').close()
+        Application.setup(config)
+        assert os.path.exists(test_file)
+
+    def test_rpm_packages(self, app):
+        from rebasehelper.results_store import results_store
```
nforro: Why not global import?

rebasehelper/tests/test_application.py
```diff
@@ -112,16 +128,48 @@ def test_application_sources(self, workdir):
...
+        res_dir = app.results_dir
+        for version in ('old', 'new'):
+            os.makedirs(os.path.join(res_dir, version, 'RPM'))
+        res = app.get_rpm_packages(res_dir)
+        assert not res
```
nforro: If you are not using the variable anywhere else, why declare it in the first place?

rebasehelper/tests/test_application.py
```diff
@@ -112,16 +128,48 @@ def test_application_sources(self, workdir):
...
+
+        # check that if rebase-helper-results exists, it does not get removed
+        test_file = os.path.join(files[1], 'test')
+        open(test_file, 'w').close()
+        Application.setup(config)
+        assert os.path.exists(test_file)
+
+    def test_rpm_packages(self, app):
+        from rebasehelper.results_store import results_store
+        res_dir = app.results_dir
```
nforro: Just a nitpick, but I don't think `res_dir` is that much shorther and more self-explanatory than `app.results_dir` to justify its existence.

rebasehelper/tests/test_application.py
FrNecas: Well, the assertions are the same as in the previous test but the directories are created by different branches of code so it imho makes sense to test it here as well. I agree with your second point, will change that.

rebasehelper/tests/test_application.py
FrNecas: I am not completely sure how pytest namespaces work. What would happen if there were 2 tests that affect `results_store`? Wouldn't the changes from the first test be present even in the second test? I expect the tests to be independent of each other.

nforro: And how exactly does this prevent that?

First of all, `results_store` is instantiated already when application.py imports it, and second, it's cleared in `Application` constructor.

PR #677, opened 2019-08-06
Refactor logging

FrNecas:


rebasehelper/logger.py
```diff
@@ -64,38 +66,85 @@ def log(message, *args, **kwargs):
...
-            level (int): Severity threshold.
+            logger_name: Logger name.
+            level: Severity threshold.
 
         Returns:
-            logging.Logger: Created logger instance.
+            Created logger instance.
 
         """
+        logging.setLoggerClass(CustomLogger)
```
nforro: Is this method really necessary? `logging.setLoggerClass(CustomLogger)` can be called just once when the module is loaded, it shouldn't be here.

rebasehelper/logger.py
```diff
@@ -107,17 +156,18 @@ def add_stream_handler(logger_object, level=None, formatter_object=None):
         return console_handler
 
     @staticmethod
-    def add_file_handler(logger_object, path, formatter_object=None, level=None):
+    def add_file_handler(logger_object: logging.Logger, path: str, formatter_object: logging.Formatter = None,
+                         level: int = None):
```
nforro: There is no return type here.

rebasehelper/logger.py
FrNecas: I initially didn't want to use the method and just use `logging.getLogger(__name__)` everywhere instead. Where would you put the `setLoggerClass` call to ensure that it gets called? For the use case of rebase-helper it could be called at the module level of logger.py but what about if one wanted to use only a part of the API (e.g. SpecFile class). We need to make sure that logger class gets properly set even in such case.

nforro: That's a good point. Let me think about it.

nforro: Well, what about `rebasehelper/__init__.py`?

rebasehelper/logger.py
```diff
@@ -132,65 +182,35 @@ def add_file_handler(logger_object, path, formatter_object=None, level=None):
...
+    log_formatter = logging.Formatter("%(message)s")
+    debug_log_formatter = logging.Formatter("%(asctime)s %(filename)s:%(lineno)s %(funcName)s: %(message)s")
+
+    debug_log = os.path.join(logs_dir, constants.DEBUG_LOG)
+    LoggerHelper.add_file_handler(logger, debug_log, debug_log_formatter, logging.DEBUG)
+    verbose_log = os.path.join(logs_dir, constants.VERBOSE_LOG)
+    LoggerHelper.add_file_handler(logger, verbose_log, log_formatter, CustomLogger.VERBOSE)
+    info_log = os.path.join(logs_dir, constants.INFO_LOG)
+    LoggerHelper.add_file_handler(logger, info_log, log_formatter, logging.INFO)
+    return debug_log
```
nforro: I think returning `debug_log` is unnecessary. `Appplication` shouldn't need to access it at all.

rebasehelper/logger.py
```diff
@@ -132,65 +182,35 @@ def add_file_handler(logger_object, path, formatter_object=None, level=None):
...
+
+
+def create_stream_handlers() -> Tuple[ColorizingStreamHandler, ColorizingStreamHandler]:
+    logger = logging.getLogger('rebasehelper')
+    formatter = logging.Formatter("%(levelname)s: %(message)s")
+    main = LoggerHelper.add_stream_handler(logger, logging.INFO, formatter)
+    logger_output = LoggerHelper.get_logger('special.output_tool', logging.INFO)
```
nforro: The name should definitely start with `rebasehelper` to prevent potential conflicts with loggers from other (3rd party) modules. Actually, I don't think `logger_output` should be defined here. It should be private attribute/implementation detail of `text` output tool.

FrNecas: `report` and `output_tool` logger seem to be different things. `output_tool` is used to display information to the console at then end of then run, `report` is used to create `report.txt`. I think `rebasehelper._output_tool_` looks a bit weird so I would personally just use `rebasehelper.report` and `rebasehelper.output_tool`. Do you agree?

nforro: True, but `output_tool` is really unfortunate name. What about `rebasehelper.summary` instead?

rebasehelper/__init__.py
```diff
@@ -22,10 +22,21 @@
...
 
+import logging
 import os
 
 import pkg_resources
 
+try:
+    from rebasehelper.logger import CustomLogger
+except ModuleNotFoundError:
+    # setup.py is being run, dependencies have not been installed
```
nforro: None of those conditions is necessarily true. Thinking about it now, perhaps simple `# bootstrap` would be sufficient.

rebasehelper/logger.py
```diff
@@ -178,19 +112,90 @@ def emit(self, record):
...
-logger_output: CustomLogger = LoggerHelper.get_basic_logger('output-tool', logging.INFO)
-logger_report: CustomLogger = LoggerHelper.get_basic_logger('rebase-helper-report', logging.INFO)
-logger_traceback: CustomLogger = LoggerHelper.get_basic_logger('traceback', CustomLogger.TRACE)
-logger_upstream: CustomLogger = LoggerHelper.get_basic_logger('rebase-helper-upstream')
+class LoggerHelper:
+    """Helper class for setting up a logger."""
+
+    @staticmethod
+    def add_stream_handler(logger_object: logging.Logger, level: int = None,
+                           formatter_object: logging.Formatter = None) -> ColorizingStreamHandler:
```
nforro: I think these `_object` suffixes are unnecessary.

Shouldn't the type here be `Optional[logging.Formatter]`?

rebasehelper/logger.py
```diff
@@ -178,19 +112,90 @@ def emit(self, record):
...
+
+    Returns:
+        Path to debug log file.
+
+    """
+    logs_dir = os.path.join(results_dir, constants.LOGS_DIR)
+    logger = logging.getLogger("rebasehelper")
+
+    log_formatter = logging.Formatter("%(message)s")
+    debug_log_formatter = logging.Formatter("%(asctime)s %(filename)s:%(lineno)s %(funcName)s: %(message)s")
```
nforro: Why `"` and not `'` here?

rebasehelper/logger.py
```diff
@@ -178,19 +112,90 @@ def emit(self, record):
...
+    verbose_log = os.path.join(logs_dir, constants.VERBOSE_LOG)
+    LoggerHelper.add_file_handler(logger, verbose_log, log_formatter, CustomLogger.VERBOSE)
+    info_log = os.path.join(logs_dir, constants.INFO_LOG)
+    LoggerHelper.add_file_handler(logger, info_log, log_formatter, logging.INFO)
+
+
+def create_stream_handlers() -> Tuple[ColorizingStreamHandler, ColorizingStreamHandler]:
+    logger = logging.getLogger('rebasehelper')
+    formatter = logging.Formatter("%(levelname)s: %(message)s")
+    main = LoggerHelper.add_stream_handler(logger, logging.INFO, formatter)
```
nforro: Just a nitpick, but the stream handlers don't have to be named. They are being used as a list anyway.

nforro: Ok, the first one is special, but you could still return a list instead of a tuple.

rebasehelper/logger.py
```diff
@@ -178,19 +112,90 @@ def emit(self, record):
...
+
+
+def create_file_handlers(results_dir: str) -> None:
+    """Creates rebase-helper file handlers.
+
+    Args:
+        results_dir: Path to rebase-helper-results directory.
+
+    Returns:
+        Path to debug log file.
```
nforro: This is no longer valid.

rebasehelper/logger.py
FrNecas: As you already said, the first one is used to set verbostiy. I think that with only 2 handlers, using tuple and naming them is a bit more clear than using a list.

rebasehelper/logger.py
```diff
@@ -178,19 +112,84 @@ def emit(self, record):
...
+    LoggerHelper.add_file_handler(logger, debug_log, debug_log_formatter, logging.DEBUG)
+    verbose_log = os.path.join(logs_dir, constants.VERBOSE_LOG)
+    LoggerHelper.add_file_handler(logger, verbose_log, log_formatter, CustomLogger.VERBOSE)
+    info_log = os.path.join(logs_dir, constants.INFO_LOG)
+    LoggerHelper.add_file_handler(logger, info_log, log_formatter, logging.INFO)
+
+
+def create_stream_handlers() -> Tuple[ColorizingStreamHandler, ColorizingStreamHandler]:
+    logger = logging.getLogger('rebasehelper')
+    formatter = logging.Formatter("%(levelname)s: %(message)s")
```
nforro: There are still `"` here.

rebasehelper/cli.py
```diff
@@ -134,20 +139,28 @@ def run():
...
                              '\n\'%s\''
                              '\nin the report.'
                              '\nThank you!',
-                             NEW_ISSUE_LINK, debug_log_file, traceback_log)
+                             NEW_ISSUE_LINK, debug_log, traceback_log)
                 LoggerHelper.add_file_handler(logger_traceback, traceback_log)
                 logger_traceback.trace('', exc_info=1)
             else:
+                LoggerHelper.add_stream_handler(logger_traceback, CustomLogger.TRACE)
+                logger_traceback.trace('', exc_info=1)
```
nforro: I think it looks better if the traceback follows the error message, not vice versa.

FrNecas: I've thought about that, both look fine to me. But from the UX perspective - I usually look at the bottom of the terminal first and I expect the most important information to be there. Imho users won't try to examine the traceback. Also if the traceback is too long, it could bury the instructions for reporting the problem. There are definitely pros and cons to both approaches.

nforro: Ok, but you should do something about the formatting. There is an empty line before the traceback, and no empty line between it and the actual error message.

FrNecas:
Changed the order of imports in this PR, added documentation for logging. Could you please have a look at the docs @nforro

rebasehelper/__init__.py
```diff
@@ -22,10 +22,21 @@
 #          Nikola Forró <nforro@redhat.com>
 #          František Nečas <fifinecas@seznam.cz>
 
+import logging
 import os
 
 import pkg_resources
```
nforro: Just noticed the lack of two empty lines between imports and code. Not sure if they should be here (do we consider the import below to be an import, if it's inside a `try-except` block?).

rebasehelper/application.py
```diff
@@ -23,20 +23,18 @@
 #          František Nečas <fifinecas@seznam.cz>
 
 import fnmatch
+import logging
 import os
 import shutil
-import logging
+from typing import Any, Dict, List, Optional, cast
```
nforro: I see a lot of inconsistencies. There is no empty line here, but there is one in many other files.

rebasehelper/__init__.py
FrNecas: That's a good question, we have something similat in `koji_helper.py` and there is only one blank line as well. I feel like it makes more sense to consider it as an import.

nforro: I have the opposite opinion. After all, imports can be part of the code (python standard library does that in a lot of places), and since the block here contains also code other than imports, I consider it a code block and as such it should be separated.

docs/api/logger.rst
```diff
@@ -1,6 +1,23 @@
...
+In addition to the module loggers, there are several loggers with a special purpose
+
+=============================== ======================================================
+Logger name                     Purpose
+=============================== ======================================================
+:samp:`rebasehelper.traceback`  for outputting traceback
+:samp:`rebasehelper.summary`    for storing messages from text output tool
+:samp:`rebasehelper.report`     for creating the final rebase-helper report
+=============================== ======================================================
+
```
nforro: I think this should be a separate chapter in user guide.

docs/api/logger.rst
```diff
@@ -1,6 +1,23 @@
 Logger module
 =============
 
+Each module of :program:`rebase-helper` uses a dedicated logger. Its name corresponds to
+the position in the module hierarchy (e.g. module :mod:`rebasehelper.application` uses the logger `rebasehelper.application`).
```
nforro: This entire paragraph could be replaced with:
```rst
Since version 0.18, :program:`rebase-helper` uses standard module-level loggers.
```

docs/api/logger.rst
```diff
@@ -1,6 +1,23 @@
...
+The :class:`rebasehelper.logger.CustomLogger` class is used instead of `logging.Logger` to add more levels of logging.
```
nforro: What about this?
```rst
:program:`rebase-helper` uses :class:`rebasehelper.logger.CustomLogger` logger class, which provides extra logging levels.
```

docs/api/logger.rst
```diff
@@ -1,6 +1,23 @@
...
+
+By default, all loggers are set to logging level `DEBUG` to ensure that all messages are caught.
```
nforro: This is unimportant. End users don't care, and they get log files of 3 verbosity levels anyway.

docs/api/logger.rst
```diff
@@ -1,6 +1,23 @@
...
+No handlers are created for the loggers, :program:`rebase-helper` adds them in :meth:`rebasehelper.cli.CliHelper.run`.
```
nforro: Also unimportant. This is implementation detail, instead you should describe how to set up default handles with `rebasehelper.logger.create_stream_handlers()`.

docs/api/logger.rst
```diff
@@ -1,6 +1,23 @@
...
+
+In addition to the module loggers, there are several loggers with a special purpose
```
nforro: What abou this?
`There are also 3 special purpose loggers:`

docs/api/logger.rst
```diff
@@ -1,6 +1,23 @@
...
+
+=============================== ======================================================
+Logger name                     Purpose
+=============================== ======================================================
+:samp:`rebasehelper.traceback`  for outputting traceback
+:samp:`rebasehelper.summary`    for storing messages from text output tool
+:samp:`rebasehelper.report`     for creating the final rebase-helper report
+=============================== ======================================================
```
nforro: I think `for` is reduntant.

docs/api/logger.rst
```diff
@@ -1,6 +1,23 @@
...
+
+By default, all loggers are set to logging level `DEBUG` to ensure that all messages are caught.
+No handlers are created for the loggers, :program:`rebase-helper` adds them in :meth:`rebasehelper.cli.CliHelper.run`.
+
+In addition to the module loggers, there are several loggers with a special purpose
+
+=============================== ======================================================
+Logger name                     Purpose
+=============================== ======================================================
+:samp:`rebasehelper.traceback`  for outputting traceback
```
nforro: This is not useful at all. What about: `captures tracebacks generated by unhandled exceptions`?

docs/api/logger.rst
```diff
@@ -1,6 +1,23 @@
...
+:samp:`rebasehelper.summary`    for storing messages from text output tool
```
nforro: What about something like `captures messages forming the rebase summary`?

docs/api/logger.rst
```diff
@@ -1,6 +1,23 @@
...
+:samp:`rebasehelper.report`     for creating the final rebase-helper report
```
nforro: What about `captures messages forming the textual rebase report (used with text output tool)`?

docs/api/logger.rst
nforro: Also, an example, or more examples, would be nice.

docs/api/logger.rst
nforro: Thinking about it, `captures` in my suggestions above is also unnecessary :wink:

docs/user_guide/logging.rst
```diff
@@ -0,0 +1,57 @@
...
+
+There are also 3 loggers with a special purpose:
+
+=============================== =====================================================================
+Logger name                     Purpose
+=============================== =====================================================================
+:samp:`rebasehelper.traceback`  traceback generated by unhandled exceptions
+:samp:`rebasehelper.summary`    messages forming the rebase summary
+:samp:`rebasehelper.report`     messages forming the textual rebase report (used by text output tool)
+=============================== =====================================================================
```
nforro: It would make more sense if the table was just after the first sentence, because it also describes loggers.

docs/user_guide/logging.rst
```diff
@@ -0,0 +1,57 @@
+Logging
+=======
+
+Since version 0.18.0 :program:`rebase-helper` uses standard module-level loggers.
+:program:`rebase-helper` uses :class:`rebasehelper.logger.CustomLogger` logger class which provides extra
+logging levels.
```
nforro: This should be a separate paragraph.

docs/user_guide/logging.rst
```diff
@@ -0,0 +1,57 @@
...
+
+Default console handlers can be setup by calling :meth:`rebasehelper.logger.create_stream_handlers` and default file
```
nforro: setup -> set up

docs/user_guide/logging.rst
```diff
@@ -0,0 +1,57 @@
...
+handlers by calling :meth:`rebasehelper.logger.create_file_handlers`.
```
nforro: I think it would be better to start this paragraph with something like `rebasehelper.logger module provides two utility functions to setup default handlers`.
Also, some description of what those default handlers actually are would be nice.

docs/user_guide/logging.rst
```diff
@@ -0,0 +1,57 @@
...
+
+.. code-block:: python
+    :caption: myapp.py
+
+    import logging
+
+    from rebasehelper.specfile import SpecFile
+
+    # Add a handler, alternatively, rebasehelper.logger.create_stream_handlers
+    # could be used to setup the default handlers.
```
nforro: Why mention the alternative here if it's described in the following example?

docs/user_guide/logging.rst
```diff
@@ -0,0 +1,57 @@
...
+    logger = logging.getLogger('rebasehelper.specfile')
+    handler = logging.StreamHandler()
+    handler.setLevel(logging.DEBUG)
```
nforro: I'd rather use [logging.basicConfig()](https://docs.python.org/3/library/logging.html#logging.basicConfig), to keep the example as simple as possible.

docs/user_guide/logging.rst
nforro: Instead, the example should be started with a comment explaining this use-case.
Something like: `# use a single module and get logs from it`

docs/user_guide/logging.rst
```diff
@@ -0,0 +1,57 @@
...
+:samp:`rebasehelper.summary`    messages forming the rebase summary
+:samp:`rebasehelper.report`     messages forming the textual rebase report (used by text output tool)
+=============================== =====================================================================
+
+
+Examples
+--------
+
+.. code-block:: python
+    :caption: myapp.py
```
nforro: I don't think this caption is useful.

docs/user_guide/logging.rst
```diff
@@ -0,0 +1,57 @@
...
+
+
+.. code-block:: python
+    :caption: rebase.py
+
+    from rebasehelper.logger import create_stream_handlers
+    from rebasehelper.config import Config
+    from rebasehelper.cli import CLI
+    from rebasehelper.application import Application
+
```
nforro: Again, something like `# run a complete rebase with default log handlers` should be here.

docs/user_guide/logging.rst
```diff
@@ -0,0 +1,57 @@
...
+    create_stream_handlers()
+    config = Config()
+    cli = CLI()
+    config.merge(cli)
+    exec_dir, res_dir = Application.setup(config)
```
nforro: Perhaps a comment mentioning `Application.setup()` calls `create_file_handlers()` could be here.

docs/user_guide/logging.rst
```diff
@@ -0,0 +1,61 @@
...
+--------
+
+.. code-block:: python
+
+    # use a single module and get logs from it
+    import logging
+
+    from rebasehelper.specfile import SpecFile
+
+    logging.basicConfig()
```
nforro: Sorry, I didn't realize using `basicConfig()` defeats the purpose of the example (as it always uses root logger).
I suggest something like this to keep the code straightforward:
```python
logger = logging.getLogger('rebasehelper.specfile')
logger.addHandler(logging.StreamHandler())
```

docs/user_guide/logging.rst
```diff
@@ -0,0 +1,61 @@
...
+logger. 3 handlers are created, they differ in the logging level. The used logging levels are DEBUG, VERBOSE and INFO.
+
+
+
+Examples
+--------
+
+.. code-block:: python
+
+    # use a single module and get logs from it
```
nforro: I would add a blank line after the comment.

docs/user_guide/logging.rst
```diff
@@ -0,0 +1,61 @@
...
+    from rebasehelper.specfile import SpecFile
+
+    logging.basicConfig()
+    spec = SpecFile('test.spec')
+    spec.set_release_number(2)
+
+
+.. code-block:: python
+
+
```
nforro: There is extra blank line here.

docs/user_guide/logging.rst
```diff
@@ -0,0 +1,63 @@
...
+:program:`rebase-helper` uses :class:`rebasehelper.logger.CustomLogger` logger class which provides extra
+logging levels.
+
+:mod:`rebasehelper.logger` module provides two utility functions to setup default handlers.
+:meth:`rebasehelper.logger.create_stream_handlers` sets up the default console handlers - with level INFO
+and a formatter displaying logging level for :samp:`rebasehelper` logger and with level DEBUG for
+:samp:`rebasehelper.summary` logger. Both of the handlers are instances of
+:class:`rebasehelper.logger.ColorizingStreameHandler` to provide colored output.
+:meth:`rebasehelper.logger.create_file_handlers` sets up the default file handlers for the main :samp:`rebasehelper`
+logger. 3 handlers are created, they differ in the logging level. The used logging levels are DEBUG, VERBOSE and INFO.
```
nforro: This is too much poorly formatted information :wink:
Something like this should suffice:
```rst
:meth:`rebasehelper.logger.create_stream_handlers` sets up the default console handlers for :samp:`rebasehelper` and :samp:`rebasehelper.summary` loggers.
:meth:`rebasehelper.logger.create_file_handlers` sets up the default file handlers for :samp:`rebasehelper` logger, with INFO, VERBOSE and DEBUG levels. The respective log files are located in :file:`rebase-helper-results/logs/`.
```
It would probably be better to format it as a list or a table.

docs/user_guide/logging.rst
```diff
@@ -0,0 +1,63 @@
...
+Since version 0.18.0 :program:`rebase-helper` uses standard module-level loggers.
+
+There are also 3 loggers with a special purpose:
+
+=============================== =====================================================================
+Logger name                     Purpose
+=============================== =====================================================================
+:samp:`rebasehelper.traceback`  traceback generated by unhandled exceptions
+:samp:`rebasehelper.summary`    messages forming the rebase summary
+:samp:`rebasehelper.report`     messages forming the textual rebase report (used by text output tool)
```
nforro: I think `text` should be formatted in some way, to indicate it's a name of a plugin.

PR #682, opened 2019-08-13
 Move setup dependencies to setup_requires

nforro:
`setuptools_scm` and `setuptools_scm_git_archive` are required for `setup.py` to run, so they need to be listed in `setup_requires`, not `install_requires`.


FrNecas:
Hmm, apparently packit build started to work but the check is failing - weird

PR #683, opened 2019-08-14
Disable stdout/stderr buffering in release webhook

nforro:
I hope this will prevent the incomplete output issue.


TomasTomecek:
Neat, so it seems the last error from the packit service was a hickup and I can see it works nicely here.

nforro:
@TomasTomecek yes, it's looking good, thanks.

PR #685, opened 2019-08-14
Remove automatically created logging file handlers

### LINKED ISSUE #684: Implement Application.cleanup()
`Application.setup()` calls `logger.create_file_handlers()`, but there is no `Application.cleaup()` that would destroy the handlers.

So if one calls `Application.setup()` multiple times, with a different working directory (which is a perfectly valid use case), the handlers from previous runs are still present and used.

I think it would make sense if complementary functions `logger.destroy_stream_handlers()` and `logger.destroy_file_handlers()` were introduced. And perhaps also `logger.destroy_all_handlers()`. `Application.cleanup()` would then call `logger.destroy_file_handlers()`.

nforro:
Fixes #684.

I decided not to implement `cleanup()` method, because it is necessary to remember the created handler instances, and since their lifetime matches the lifetime of `Application` instance, creating them in its constructor and removing them in its destructor makes more sense.

### DIFF
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -34,8 +34,9 @@ from pkg_resources import parse_version
 from rebasehelper import VERSION
 from rebasehelper.archive import Archive
 from rebasehelper.specfile import SpecFile, get_rebase_name
-from rebasehelper.logger import create_file_handlers, CustomLogger
+from rebasehelper.logger import CustomLogger, LoggerHelper
 from rebasehelper import constants
+from rebasehelper.config import Config
 from rebasehelper.patcher import Patcher
 from rebasehelper.plugins.plugin_manager import plugin_manager
 from rebasehelper.plugins.checkers import CheckerCategory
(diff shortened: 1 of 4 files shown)


nforro:
> There may be cases where you want to rebase without creating log files.

Really? I can't think of such case. I think it's fine to create log files if `rebase-helper-results` directory is being created and used for other stuff anyway. Which might not be the case if someone is using just some `rebasehelper` modules though.

Sure, I can move `create_file_handlers` before `Application` instantiation and `remove_file_handlers` after `Application.run()`, but it makes the whole "doing a rebase" thing rather cumbersome.

FrNecas:
I can think of 2 use-cases for not creating log files automatically. You either want to just use stream handlers and log everything important into terminal. Or you want to setup logs but with yout own levels and formatters. But both of these seem to be rather niche.

You are right that it's probably not worth it. It would definitely add more control but also a lot of unnecessary steps. That idea was just something for consideration. I assume that this PR mainly aims to make rebasing work in `batch.py` where the handlers have to be removed after each run. So it definitely makes sense to use destructor

nforro:
> I can think of 2 use-cases for not creating log files automatically. You either want to just use stream handlers and log everything important into terminal. Or you want to setup logs but with yout own levels and formatters.

You can do both without caring about the default file handlers, you can just add your own.

But I agree with you, there should be a way to disable it. How about adding some boolean argument to `Application` constructor?

FrNecas:
Sounds good to me. I think it should create the handlers by default.

rebasehelper/logger.py
```diff
@@ -157,38 +160,58 @@ def add_file_handler(logger: logging.Logger, path: str, formatter: Optional[logg
...
 
-    logger_report = logging.getLogger('rebasehelper.report')
-    logger_report.propagate = False
-    return main, summary
+        logger_report = logging.getLogger('rebasehelper.report')
+        logger_report.propagate = False
+        return main, summary
+
+    @classmethod
+    def remove_file_handlers(cls, handlers: List[logging.FileHandler]) -> None:
```
FrNecas: Shouldn't this rather be called `remove_handlers`? Since it removes every handler that is passed and even stream handlers could be removed.

rebasehelper/application.py
```diff
@@ -70,15 +71,20 @@ class Application:
...
+            execution_dir: Working directory.
+            results_dir: Location of rebase results.
+            create_logs: Whether to create default logging file handlers.
 
-        :param cli_conf: CLI object with configuration gathered from commandline
-        :return:
         """
         results_store.clear()
 
+        self.handlers = LoggerHelper.create_file_handlers(results_dir) if create_logs else None
```
FrNecas: What about `self.handlers = LoggerHelper.create_file_handlers(results_dir) if create_logs else []`? This way you won't have to use the if condition in the destructor

rebasehelper/logger.py
nforro: True, but it does that only for `rebasehelper` logger. You can't remove default summary stream handler this way.

PR #687, opened 2019-08-16
Refactor Application attributes

FrNecas:


rebasehelper/application.py
```diff
@@ -244,12 +242,13 @@ def get_rpm_packages(self, dirname):
             return False
         return True
 
-    def _get_spec_file(self):
+    def _get_spec_file(self) -> str:
         """Function gets the spec file from the execution_dir directory"""
```
nforro: How about updating the docstring as well?

rebasehelper/application.py
```diff
@@ -244,12 +242,13 @@ def get_rpm_packages(self, dirname):
             return False
         return True
 
-    def _get_spec_file(self):
+    def _get_spec_file(self) -> str:
```
nforro: I think `_find_spec_file` would be much clearer name.

PR #692, opened 2019-08-22
Avoid false-positives when replacing with macros

FrNecas:


rebasehelper/helpers/macro_helper.py
```diff
@@ -94,7 +94,8 @@ def substitute_path_with_macros(path, macros):
 
         """
         for m in macros:
-            if m['value'] and m['value'] in path:
+            # avoid false-positives caused by replacing one character by a macro
+            if len(m['value']) > 1 and m['value'] in path:
```
nforro: Actually, I don't think the filter here is necessary, as the whitelisted macros will always pass anyway.
However, I wonder why [this](https://github.com/rebase-helper/rebase-helper/blob/master/rebasehelper/plugins/spec_hooks/paths_to_rpm_macros.py#L35-L37) is not part of this method?

rebasehelper/specfile.py
```diff
@@ -1402,6 +1402,8 @@ def update_setup_dirname(self, dirname):
                     macros = [m for m in MacroHelper.filter(self.macros, level=-3) if m['name'] in ('name', 'version')]
                     # add all macros from spec file scope
                     macros.extend(MacroHelper.filter(self.macros, level=0))
+                    # avoid false-positives caused by replacing one character by a macro
```
nforro: I really don't feel like false-positive is a suitable term here. How about simple `# omit extremely short macros`?

rebasehelper/helpers/macro_helper.py
FrNecas: I put the filter there as the macros could possibly be redefined. But looking at it now, it would be completely illogical so it's pointless to have it there

FrNecas: Regarding your second point - expanding and sorting the macros on each substitution could affect the performance. It's enough to expand and sort only once.

nforro: True, but the method won't function correctly with unexpanded and unsorted macros. So I feel like those steps should be part of it. But let's keep it this way for now.

FrNecas: Hmm, you are right. What about moving the whole `macros` definition to this method? So it would look like this:

```python
macros = [m for m in rebase_spec_file.macros if m['name'] in cls.MACROS_WHITELIST]
macros = cls.expand_macros(macros)
# ensure maximal greediness
macros.sort(key=lambda k: len(k['value']), reverse=True)
for m in macros:
      ....
```

We would lose the ability to control which macros are used as replacement but with the current use it would be enough. We are probably not going to want to use any other macros outside of `MACROS_WHITELIST` as substitution.

nforro: But that will only help if you make the list of macros a class attribute.

FrNecas: That doesn't sound right either. I will leave it like that for now

PR #693, opened 2019-08-23
Fix --comparepkgs-only option

FrNecas:


rebasehelper/application.py
```diff
@@ -217,12 +217,16 @@ def _initialize_data(self):
...
-        Function returns RPM packages stored in dirname/old and dirname/new directories
+    def get_rpm_packages(self, dirname: str) -> bool:
+        """Function returns RPM packages stored in dirname/old-build
+        and dirname/new-build directories.
+
+        Args:
+            dirname: Path to a directory with RPMs.
+
+        Returns:
+            Whether any packages were found.
```
nforro: This is in contradiction with the description, which is plain wrong anyway and needs to be replaced.

rebasehelper/application.py
```diff
@@ -233,9 +237,10 @@ def get_rpm_packages(self, dirname):
             else:
                 spec_version = self.rebase_spec_file.get_version()
             data['version'] = spec_version
-            data['rpm'] = PathHelper.find_all_files(os.path.join(os.path.realpath(dirname), version, 'RPM'), '*.rpm')
+            dir_path = os.path.join(os.path.realpath(dirname), version + '-build', 'RPM')
```
nforro: `dir_path`? What does that even mean? :-)

rebasehelper/logger.py
```diff
@@ -176,7 +176,10 @@ def create_file_handlers(cls, results_dir: str) -> List[logging.FileHandler]:
 
         """
         logs_dir = os.path.join(results_dir, constants.LOGS_DIR)
-        os.makedirs(logs_dir)
+        # the logs directory can already exist when using some rebase-helper
+        # options (e.g --comparepkgs-only)
+        if not os.path.exists(logs_dir):
+            os.makedirs(logs_dir)
```
nforro: Why not `exist_ok=True`?

rebasehelper/application.py
```diff
@@ -217,12 +217,16 @@ def _initialize_data(self):
         self.old_rest_sources = [os.path.abspath(x) for x in self.spec_file.get_sources()[1:]]
         self.new_rest_sources = [os.path.abspath(x) for x in self.rebase_spec_file.get_sources()[1:]]
 
-    def get_rpm_packages(self, dirname):
-        """
-        Function returns RPM packages stored in dirname/old and dirname/new directories
+    def get_rpm_packages(self, dirname: str) -> bool:
+        """Checks if there are RPM packages in dirname/old-build and
+        dirname/new-build. Saves the found packages to result store
```
nforro: Hm, how about `Updates results store with build data found in the specified directory, if any.`?

rebasehelper/application.py
```diff
@@ -217,12 +217,16 @@ def _initialize_data(self):
         self.old_rest_sources = [os.path.abspath(x) for x in self.spec_file.get_sources()[1:]]
         self.new_rest_sources = [os.path.abspath(x) for x in self.rebase_spec_file.get_sources()[1:]]
 
-    def get_rpm_packages(self, dirname):
-        """
-        Function returns RPM packages stored in dirname/old and dirname/new directories
+    def get_rpm_packages(self, dirname: str) -> bool:
```
nforro: The method name also doesn't make much sense.

FrNecas: So what name do you suggest? Perhaps `check_rpm_packages` or `find_rpm_packages`?

nforro: Well, it's hard to say, because the method does more than just checking if there are some RPMs, but something like `update_build_data` doesn't sound right either.

FrNecas: It is probably impossible to find a short and descriptive name for this method. I fee like `check` is the closest we may get as it describes the use-case more than `update_build_data`. Both are better than the current name

nforro: Ok.

rebasehelper/application.py
```diff
@@ -217,12 +217,17 @@ def _initialize_data(self):
         self.old_rest_sources = [os.path.abspath(x) for x in self.spec_file.get_sources()[1:]]
         self.new_rest_sources = [os.path.abspath(x) for x in self.rebase_spec_file.get_sources()[1:]]
 
-    def get_rpm_packages(self, dirname):
-        """
-        Function returns RPM packages stored in dirname/old and dirname/new directories
+    def check_rpm_packages(self, dirname: str) -> bool:
+        """Checks if there are RPM packages in dirname/old-build and
+        dirname/new-build. Updates results store with the build data
+        found in the specified directory, if any.
```
nforro: This is rather redundant. If you want to keep the first sentence, you might as well just add ` and updates results store accordingly` and drop the second one.

PR #695, opened 2019-08-27
Remove action control options

FrNecas:
The `check_rpm_packages` could also be removed, it is no longer used after removing `--comparepkgs-only`. Should I remove it @nforro or do we want to keep it (perhaps it could be useful in the future)?


nforro:
Sure, you can remove it as well.
What about `--continue`, does that option still work?

Issue #697, opened 2019-08-30
abipkgdiff checker has to inform about SONAME bumps


nforro:
abipkgdiff report after rebasing **exempi** from 2.4.5 to 2.5.1:

```
Removed binaries:
  /usr/lib64/libexempi.so.3.4.5, SONAME: libexempi.so.3
Added binaries:
  /usr/lib64/libexempi.so.8.0.0, SONAME: libexempi.so.8
```

Despite the fact no ABI changes were detected, SONAME changed and that means all dependent packages have to be rebuilt, so the maintainer <ins>has to</ins> be informed.

PR #699, opened 2019-09-03
Add sonamecheck

FrNecas:
Resolves: #697

### DIFF
--- a/rebasehelper/plugins/checkers/sonamecheck.py
+++ b/rebasehelper/plugins/checkers/sonamecheck.py
@@ -1,0 +1,102 @@
+# -*- coding: utf-8 -*-
+#
+# This tool helps you rebase your package to the latest version
+# Copyright (C) 2013-2019 Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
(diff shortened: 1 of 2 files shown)


rebasehelper/plugins/checkers/sonamecheck.py
```diff
@@ -0,0 +1,106 @@
...
+    @classmethod
+    def run_check(cls, results_dir, **kwargs):
+        cls.results_dir = os.path.join(results_dir, cls.name)
+        os.makedirs(cls.results_dir)
+        old_pkgs = results_store.get_old_build().get('rpm', [])
+        new_pkgs = results_store.get_new_build().get('rpm', [])
+
+        soname_changes = {}
+        for old_pkg in old_pkgs:
+            name = RpmHelper.split_nevra(os.path.basename(old_pkg))['name']
```
nforro: I think it would be better to gather headers from all RPMs first, and continue using that information later on.

FrNecas: Would you mind explaining why is this approach better? I don't really see any advantages, the comparisons will be the same and you will have to iterate over the gathered headers anyway

nforro: Well, exactly, you only need the headers. And getting package name from RPM header is far more realiable than getting it from filename.

nforro:
Looking at how cumbersome the `rpm.hdr` class is to work with, perhaps it would be worth it to make a helper method in `RpmHelper` to convert it to `dict` (with all values already decoded of course).

rebasehelper/plugins/checkers/sonamecheck.py
```diff
@@ -0,0 +1,102 @@
...
+    DEFAULT: bool = True
+    CATEGORY: Optional[CheckerCategory] = CheckerCategory.RPM
+    results_dir: Optional[str] = ''
+
+    @classmethod
+    def is_available(cls):
+        return True
+
+    @classmethod
+    def _get_soname(cls, headers: rpm.hdr) -> Optional[str]:
```
nforro: `headers` should be singular, shouldn't it?

PR #704, opened 2019-09-05
Preserve whitespace when modifying Patch tags

nforro:


FrNecas:
Looks good to me but could you please change the test in TestBuildLogHooks as well? I know it isn't failing now but it could prevent potential failure in the future and it will also make the tests consistent.

nforro:
Sure, thanks, I completely missed that.

PR #706, opened 2019-09-06
Fix copr build tool and make it more user friendly.

FrNecas:


rebasehelper/helpers/download_helper.py
```diff
@@ -152,6 +152,9 @@ def download_file(url, destination_path, blocksize=8192):
             raise DownloadError(r.reason)
 
         file_size = int(r.headers.get('content-length', -1))
+        encoding = r.headers.get('content-encoding', None)
+        if encoding and encoding != 'identity':
```
nforro: You could squash this to a single line: `if r.headers.get('content-encoding', 'identity') != 'identity':`.

And I believe this deserves a comment, something like `use infinite progress bar in case of compressed content, since there is no way to determine the total uncompressed size in advance`.

rebasehelper/helpers/copr_helper.py
```diff
@@ -146,12 +145,6 @@ def download_build(cls, client, build_id, destination):
...
                 elif fn.endswith('.log.gz'):
                     extracted = dest.replace('.log.gz', '.log')
-                    try:
-                        with gzip.open(dest, 'rb') as archive:
-                            with open(extracted, 'wb') as f:
-                                f.write(archive.read())
-                    except (IOError, EOFError):
-                        raise RebaseHelperError(
-                            'Failed to extract {}'.format(dest))
+                    os.rename(dest, extracted)
```
nforro: `extracted` doesn't really make sense anymore.

rebasehelper/helpers/download_helper.py
```diff
@@ -152,6 +152,10 @@ def download_file(url, destination_path, blocksize=8192):
             raise DownloadError(r.reason)
 
         file_size = int(r.headers.get('content-length', -1))
+        # use infinite progress bar in case of compressed content, there is no
+        # way to determine the uncompressed size in advance
```
nforro: I think the comment should be after the condition, at least it makes more sense to me that way.

Issue #708, opened 2019-09-12
Add option to create temporary projects in Copr


nforro:
Copr allows to create temporary projects that are auto-deleted after 60 days. `copr` build tool should have an option to enable that.

Issue #461, opened 2018-03-06
Add options for remote build tools


nforro:
Currently there is no way how to choose for example specific build tag or architecture. Tag `rawhide` and architecture `x86_64` are hardcoded.

PR #714, opened 2019-09-20
Add basic copr options

FrNecas:
Resolves: #708 
Related issue: #461 (adds a mechanism for passing options to builders)

### DIFF
--- a/rebasehelper/helpers/copr_helper.py
+++ b/rebasehelper/helpers/copr_helper.py
@@ -41,6 +41,8 @@ logger: CustomLogger = cast(CustomLogger, logging.getLogger(__name__))
 
 class CoprHelper:
 
+    DELETE_PROJECT_AFTER = 60
+
     @classmethod
     def get_client(cls):
         try:
@@ -52,11 +54,13 @@ class CoprHelper:
             return client
 
     @classmethod
-    def create_project(cls, client, project, chroot, description, instructions):
+    def create_project(cls, client, project, chroot, description, instructions, permanent=False, hide=True):
         try:
             client.project_proxy.add(ownername=client.config.get('username'),
                                      projectname=project,
                                      chroots=[chroot],
+                                     delete_after_days=None if permanent else cls.DELETE_PROJECT_AFTER,
+                                     unlisted_on_hp=hide,
                                      description=description,
                                      instructions=instructions)
         except CoprRequestException:
--- a/rebasehelper/plugins/build_tools/rpm/copr_.py
+++ b/rebasehelper/plugins/build_tools/rpm/copr_.py
@@ -41,6 +41,21 @@ class Copr(BuildToolBase):
     Class representing Copr build tool.
     """
 
+    OPTIONS = [
+        {
+            'name': ['--copr-project-permanent'],
+            'switch': True,
+            'default': False,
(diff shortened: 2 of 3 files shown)


rebasehelper/plugins/build_tools/rpm/copr_.py
```diff
@@ -41,6 +41,21 @@ class Copr(BuildToolBase):
     Class representing Copr build tool.
     """
 
+    OPTIONS = [
+        {
+            'name': ['--remove-copr-project-after'],
+            'default': 60,
+            'help': 'after how many days the created copr build will be '
+                    'deleted, -1 for never, defaults to %(default)s',
```
nforro: I think a simple switch would be sufficient, something like `--copr-project-permanent`, defaulting to `False`.

rebasehelper/plugins/build_tools/rpm/copr_.py
```diff
@@ -41,6 +41,21 @@ class Copr(BuildToolBase):
...
+        },
+        {
+            'name': ['--visible-copr-project'],
```
nforro: `--copr-project-frontpage` sounds a bit better to me.

rebasehelper/plugins/build_tools/rpm/copr_.py
FrNecas: You are right that this is too complex. The number of days can be later changed in the web interface. So it should either be permanent or last 60 days?

nforro: Exactly.

rebasehelper/helpers/copr_helper.py
```diff
@@ -52,11 +52,13 @@ def get_client(cls):
             return client
 
     @classmethod
-    def create_project(cls, client, project, chroot, description, instructions):
+    def create_project(cls, client, project, chroot, description, instructions, permanent=False, hide=True):
         try:
             client.project_proxy.add(ownername=client.config.get('username'),
                                      projectname=project,
                                      chroots=[chroot],
+                                     delete_after_days=None if permanent else 60,
```
nforro: I would rather see a constant.

rebasehelper/plugins/build_tools/rpm/copr_.py
```diff
@@ -41,6 +41,21 @@ class Copr(BuildToolBase):
     Class representing Copr build tool.
     """
 
+    OPTIONS = [
+        {
+            'name': ['--copr-permanent-project'],
```
nforro: Why not `--copr-project-permanent`?

FrNecas: I meant permanent as an adjective tied to project. Though you are right, that `--copr-project-permanent` is better, it implies that bot of the added options influence the project.

rebasehelper/plugins/build_tools/rpm/copr_.py
```diff
@@ -41,6 +41,21 @@ class Copr(BuildToolBase):
     Class representing Copr build tool.
     """
 
+    OPTIONS = [
```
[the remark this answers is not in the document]
FrNecas: I will fix it in the PR where I will enable specifying chroots

PR #718, opened 2019-09-24
Add option to change copr chroot

FrNecas:


rebasehelper/helpers/copr_helper.py
```diff
@@ -54,18 +55,30 @@ def get_client(cls):
...
+                                     chroots=chroots,
                                      delete_after_days=None if permanent else cls.DELETE_PROJECT_AFTER,
                                      unlisted_on_hp=hide,
                                      description=description,
                                      instructions=instructions)
-        except CoprRequestException:
-            # reuse existing project
-            pass
+        except CoprRequestException as e:
+            error = e.result.error
```
nforro: Can you list all possible variations of this in a comment?

rebasehelper/helpers/copr_helper.py
```diff
@@ -54,18 +54,27 @@ def get_client(cls):
...
+            try:
+                error = next(iter(error.values()))[0]
```
nforro: I'm still suspicious about this. How does the error *object* look like?

rebasehelper/helpers/copr_helper.py
```diff
@@ -54,18 +54,27 @@ def get_client(cls):
...
+            except AttributeError:
+                pass
+
+            if error.startswith('You already have project named'):
+                # reuse existing project
+                pass
+            else:
+                raise RebaseHelperError(error)
```
nforro: I think there should be some prefix to avoid confusion.

rebasehelper/helpers/copr_helper.py
nforro: Ok, so it really is a dict containing single-item list :roll_eyes: 
In that case I think this is much more readable:
```python
[[error]] = error.values()
```

rebasehelper/helpers/copr_helper.py
nforro: No, I meant to prefix the error string so that a user knows it comes from copr.

rebasehelper/helpers/copr_helper.py
```diff
@@ -54,18 +54,27 @@ def get_client(cls):
...
+            try:
+                [[error]] = error.values()
+            except AttributeError:
+                pass
+
+            if error.startswith('You already have project named'):
+                # reuse existing project
+                pass
+            else:
+                raise RebaseHelperError('copr: ' + error)
```
nforro: Ok, but this is not very user friendly. Why not raise something like [this](https://github.com/rebase-helper/rebase-helper/blob/ec13174dda28800825354110b88fe45f78556b35/rebasehelper/specfile.py#L296-L297)?

PR #722, opened 2019-10-01
Fix --build-tasks option

FrNecas:


rebasehelper/application.py
```diff
@@ -482,7 +482,7 @@ def build_source_packages(self):
 
         for version in ['old', 'new']:
             koji_build_id = None
-            results_dir = '{}-build'.format(os.path.join(self.results_dir, version))
+            results_dir = os.path.join('{}-build'.format(os.path.join(self.results_dir, version)), 'SRPM')
```
nforro: This is ugly and unnecessary.
Why not simply `os.path.join(self.results_dir, '{}-build'.format(version), 'SRPM')`?

rebasehelper/plugins/build_tools/rpm/copr_.py
```diff
@@ -130,3 +130,16 @@ def get_detached_task(cls, task_id, results_dir):
...
+        client = CoprHelper.get_client()
+        build_id = int(task_id)
+        failed = not CoprHelper.watch_build(client, build_id)
+        rpms, logs = CoprHelper.download_build(client, build_id, results_dir)
+        build_url = CoprHelper.get_build_url(client, build_id)
+        logs.append(build_url)
+        if failed:
+            logger.error('Copr build failed %s', build_url)
+            raise BinaryPackageBuildError(logs=logs)
+        return rpms, logs,
```
nforro: Why comma at the end?

PR #725, opened 2019-10-08
Make extra version mostly work

nforro:
This should fix rebasing from pre/post-release to a stable release, and from pre/post-release to a different pre/post-release. Rebasing from stable release to pre/post-release most likely requires modifying `Source0` tag, which is currently not implemented.

### DIFF
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -164,13 +164,14 @@ class Application:
         # check if argument passed as new source is a file or just a version
         if [True for ext in Archive.get_supported_archives() if self.conf.sources.endswith(ext)]:
             logger.verbose("argument passed as a new source is a file")
-            self.rebase_spec_file.set_version_using_archive(self.conf.sources)
+            version_string = self.spec_file.extract_version_from_archive_name(self.conf.sources,
+                                                                              self.spec_file.get_main_source())
         else:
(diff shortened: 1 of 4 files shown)


rebasehelper/application.py
```diff
@@ -164,13 +164,15 @@ def _prepare_spec_objects(self):
         # check if argument passed as new source is a file or just a version
         if [True for ext in Archive.get_supported_archives() if self.conf.sources.endswith(ext)]:
             logger.verbose("argument passed as a new source is a file")
-            self.rebase_spec_file.set_version_using_archive(self.conf.sources)
+            main_source = self.spec_file._get_raw_source_string(  # pylint: disable=protected-access
```
FrNecas: Since you are using the method in Application now, wouldn't it perhaps make sense to make it public?

nforro: Maybe it would be better to make a wrapper method `SpecFile.get_main_source()`?
I didn't originally used the method at all, as I made `SpecFile.extract_version_from_archive_name()` non-static and with a single argument, but that proved problematic to unit-test.

FrNecas: Creating `get_main_source` would be great to encapsulate the call, good suggestion

PR #735, opened 2019-10-17
Add crate and gem archive types

### LINKED ISSUE #724: Fix Archive.get_supported_archives()
After PR #665, `Archive.get_supported_archives()` no longer returns `.gem` and `.crate`, which means such files can't be used as `SOURCES` option.
It would probably make sense to implement something like `get_known_archives()` instead.

### LINKED ISSUE #731: Add proper support for .gem and .crate
`.gem` and `.crate` *archives* are currently copied, but it doesn't make any sense to do so, as patch rebasing can't take place, and there is no other reason to copy the files.

But both `.gem` and `.crate` are actually (possibly compressed) tar archives, so adding support for them should be pretty straightforward. I think it would be best to just ignore the extension and try to handle all archives as tarballs first and implement fallbacks for other types.

nforro:
Fixes #731.
Fixes #724.


rebasehelper/archive.py
```diff
@@ -82,134 +79,90 @@ def extract(cls, archive=None, filename=None, path=None):
...
+    EXTENSION: str = '.tar.bz2'
 
     @classmethod
-    def open(cls, filename=None):
-        if filename is None:
-            raise TypeError("Expected argument 'filename' (pos 1) is missing")
-
-        if filename.endswith('.tar.bz2'):
+    def open(cls, filename):
+        if filename.endswith(TarBz2ArchiveType.EXTENSION):
```
FrNecas: Is there any reason not to use `cls.EXTENSION`?

nforro: Of course, the condition would be pointless in the case of `cls` being `Bz2ArchiveType`.

FrNecas: Thanks for the explanation :)

Issue #721, opened 2019-09-30
Rebase based on bugzilla ID


nforro:
There are bugs filed by *Upstream Release Monitoring*, for example [this one](https://bugzilla.redhat.com/show_bug.cgi?id=1756742).
It would be nice if **rebase-helper** could, given a BZ ID, do a complete rebase automatically.

PR #737, opened 2019-10-21
Add option to rebase based on bugzilla ID

FrNecas:
Resolves: #721

### DIFF
--- a/rebasehelper/cli.py
+++ b/rebasehelper/cli.py
@@ -35,6 +35,7 @@ from rebasehelper.application import Application
 from rebasehelper.logger import CustomLogger, LoggerHelper
 from rebasehelper.exceptions import RebaseHelperError
 from rebasehelper.helpers.console_helper import ConsoleHelper
+from rebasehelper.helpers.bugzilla_helper import BugzillaHelper
 from rebasehelper.config import Config
 from rebasehelper.argument_parser import CustomArgumentParser, CustomHelpFormatter, CustomAction
 from rebasehelper.plugins.plugin_manager import plugin_manager
@@ -123,6 +124,12 @@ class CliHelper:
             else:
                 main_handler.setLevel(logging.DEBUG)
             ConsoleHelper.use_colors = ConsoleHelper.should_use_colors(config)
+            if config.bugzilla_id:
+                repo_path, config.sources = BugzillaHelper.prepare_rebase_repository(config.bugzilla_id)
+                try:
+                    os.chdir(repo_path)
+                except OSError:
+                    raise RebaseHelperError('Could not change directory to the cloned repository')
             execution_dir, results_dir = Application.setup(config)
             app = Application(config, execution_dir, results_dir)
             app.run()
--- a/rebasehelper/helpers/bugzilla_helper.py
+++ b/rebasehelper/helpers/bugzilla_helper.py
@@ -1,0 +1,155 @@
+# -*- coding: utf-8 -*-
+#
+# This tool helps you rebase your package to the latest version
+# Copyright (C) 2013-2019 Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Authors: Petr Hráček <phracek@redhat.com>
+#          Tomáš Hozza <thozza@redhat.com>
+#          Nikola Forró <nforro@redhat.com>
+#          František Nečas <fifinecas@seznam.cz>
+
+import logging
+import os
+import re
+from typing import cast, Optional, Tuple
+
+import git  # type: ignore
+import requests
+
+from rebasehelper.exceptions import RebaseHelperError
+from rebasehelper.logger import CustomLogger
+
+
+logger: CustomLogger = cast(CustomLogger, logging.getLogger(__name__))
+
+
+class BugzillaHelper:
+
+    """Class for working with Upstream Release Monitoring on bugzilla."""
+
+    DIST_GIT_REPO_URL = 'https://src.fedoraproject.org/rpms'
+    BUGZILLA_REST_API_URL = 'https://bugzilla.redhat.com/rest'
+    UPSTREAM_RELEASE_MONITORING_USERNAME = 'upstream-release-monitoring'
+
+    @classmethod
+    def get_bugzilla_component(cls, bugzilla_id: str) -> str:
+        """Gets a component of the bugzilla.
+
+        Args:
+            bugzilla_id: ID of the bugzilla.
+
+        Returns:
+            Component of the bugzilla.
+
+        Raises:
+            RebaseHelperError: If no such bugzilla exists or if the bugzilla
+            was not created by Upstream Release Monitoring.
+
+        """
+        r = requests.get('{}/bug/{}'.format(cls.BUGZILLA_REST_API_URL, bugzilla_id))
+        if not r.ok:
+            raise RebaseHelperError('Could not obtain data from bugzilla')
+
+        response_json = r.json()
+        if 'error' in response_json:
+            logger.error('Bugzilla error: %s', response_json['error'])
+            raise RebaseHelperError('Could not obtain data from bugzilla')
+
+        bug = response_json['bugs'][0]
+        if bug['creator_detail']['email'] != cls.UPSTREAM_RELEASE_MONITORING_USERNAME:
+            raise RebaseHelperError('The given bugzilla was not created by Upstream Release Monitoring')
+        return bug['component'][0]
+
+    @classmethod
+    def get_version_from_comments(cls, bugzilla_id: str) -> Optional[str]:
+        """Gets version from bugzilla comments.
+
+        Args:
+            bugzilla_id: ID of the bugzilla.
+
+        Returns:
+            Version specified by Upstream Release Monitoring in comments
+            or None, if no version could be found.
+
+        Raises:
+            RebaseHelperError: If no such bugzilla exists.
+
+        """
+        r = requests.get('{}/bug/{}/comment'.format(cls.BUGZILLA_REST_API_URL, bugzilla_id))
+        if not r.ok:
+            raise RebaseHelperError('Could not obtain data from bugzilla')
+        version = None
+        comments = r.json()['bugs'][bugzilla_id]['comments']
+        pattern = re.compile(r'^Latest upstream release: (?P<version>.*)\n')
+        for comment in comments:
+            if comment['creator'] != cls.UPSTREAM_RELEASE_MONITORING_USERNAME:
+                continue
+            match = pattern.match(comment['text'])
+            if match:
+                version = match.group('version')
+
+        return version
+
+    @classmethod
+    def clone_repository(cls, component: str, bugzilla_id: str) -> str:
+        """Clones remote dist-git repository of a component.
+
+        Args:
+            component: Package to clone.
+            bugzilla_id: ID of the bugzilla.
+
+        Returns:
+            Path to the cloned repository.
+
+        Raises:
+            RebaseHelperError: If the directory, that the repository
+            is supposed to be cloned into, exists.
+
+        """
+        path = os.path.abspath('{}-{}'.format(bugzilla_id, component))
+        if os.path.exists(path):
+            raise RebaseHelperError('Could not clone the repository because the directory '
+                                    '{} already exists'.format(path))
+
+        url = '{}/{}.git'.format(cls.DIST_GIT_REPO_URL, component)
+        logger.info("Cloning %s into %s", url, path)
+        git.Repo.clone_from(url, path)
+        return path
+
+    @classmethod
+    def prepare_rebase_repository(cls, bugzilla_id: str) -> Tuple[str, str]:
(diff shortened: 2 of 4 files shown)


rebasehelper/cli.py
```diff
@@ -123,6 +124,10 @@ def run():
             else:
                 main_handler.setLevel(logging.DEBUG)
             ConsoleHelper.use_colors = ConsoleHelper.should_use_colors(config)
+            if config.bugzilla_id:
+                name, version = BugzillaHelper.prepare_rebase_repository(config.bugzilla_id)
+                os.chdir(name)
+                config.sources = version
```
nforro: Why not assign `config.sources` in place, there is no need for extra variable. Also, I think `component` would be better name than `name`. Or perhaps `package` or `package_name`, if you want to avoid the word component for some reason.

rebasehelper/cli.py
```diff
@@ -123,6 +124,10 @@ def run():
             else:
                 main_handler.setLevel(logging.DEBUG)
             ConsoleHelper.use_colors = ConsoleHelper.should_use_colors(config)
+            if config.bugzilla_id:
+                name, version = BugzillaHelper.prepare_rebase_repository(config.bugzilla_id)
+                os.chdir(name)
```
nforro: This can raise several exceptions, I think it would be good to handle them.

rebasehelper/helpers/bugzilla_helper.py
```diff
@@ -0,0 +1,79 @@
...
+        """Clones a repository based on Upstream Release Monitoring bug.
+
+        Args:
+            bugzilla_id: ID of the bug.
+
+        Returns:
+            Name of the cloned package and version to rebase to.
+
+        """
+        bug_data = requests.get('https://bugzilla.redhat.com/rest/bug/{}'.format(bugzilla_id)).json()
```
nforro: You expect result from `requests.get()` to contain valid json data, but that is not always the case. There could be exceptions raised, the response could be a 4xx or 5xx HTTP error and contain no parseable data at all.

rebasehelper/helpers/bugzilla_helper.py
```diff
@@ -0,0 +1,79 @@
...
+        if 'error' in bug_data:
+            raise RebaseHelperError("No bug with ID {} was found".format(bugzilla_id))
+        bug = bug_data['bugs'][0]
+        if bug['creator'] != 'Upstream Release Monitoring':
```
nforro: I think it would be a bit safer to check creator ID instead of name, or at least email. Actually, thinking about it, email is way less likely to change in the future than name or ID.

rebasehelper/helpers/bugzilla_helper.py
```diff
@@ -0,0 +1,79 @@
...
+            raise RebaseHelperError('The given bug was not created by Upstream Release Monitoring')
+        component = bug['component'][0]
```
nforro: It's really weird that you are getting an array here, when the [docs](https://bugzilla.redhat.com/docs/en/html/api/core/v1/bug.html#get-bug) state it's a string.

rebasehelper/helpers/bugzilla_helper.py
```diff
@@ -0,0 +1,79 @@
...
+        match = re.match(r'^.*-(?P<version>.*) is available$', bug['summary'])
+        if match:
+            version = match.group('version')
+        else:
+            raise RebaseHelperError('Failed to parse new version from the given bug')
```
nforro: I think it would be better to get the version string from `Latest upstream release:` field in comment(s). There can be [multiple](https://bugzilla.redhat.com/show_bug.cgi?id=1552349) version comments on the same bug, and while the summary is also updated with each such comment, I still think it's safer to do it this way. If nothing else, you don't have to deal with package name, which can be difficult to distinguish from version (especially if it contains hyphens, which I'm not sure we can rule out).

rebasehelper/helpers/bugzilla_helper.py
```diff
@@ -0,0 +1,79 @@
...
+
+        url = 'https://src.fedoraproject.org/rpms/{}.git'.format(component)
```
nforro: This shouldn't be hardcoded. It could change (it already has at least once in the recent history), a constant would be better.

rebasehelper/helpers/bugzilla_helper.py
```diff
@@ -0,0 +1,79 @@
...
+        if os.path.exists(component):
+            if InputHelper.get_message("A folder named {} found, do you want to remove it?".format(component), False):
```
nforro: Please don't use `folder`, that's Windows terminology, and not something anyone would expect to see in a CLI application. Also, the message is not very helpful, the user doesn't know what is happening right now, why would they want to remove some directory (referenced just by name, not a full path, so it could be anything).

nforro: I think we should default to failing if the directory already exists. Also, cloning to the current directory is not very smart. Making a temporary directory for it would be better choice. Just like `fedora-review` does it.

rebasehelper/helpers/bugzilla_helper.py
```diff
@@ -0,0 +1,79 @@
...
+    @classmethod
+    def prepare_rebase_repository(cls, bugzilla_id: str) -> Tuple[str, str]:
+        """Clones a repository based on Upstream Release Monitoring bug.
+
+        Args:
+            bugzilla_id: ID of the bug.
+
+        Returns:
+            Name of the cloned package and version to rebase to.
+
```
nforro: You should also add `Raises`.

rebasehelper/helpers/bugzilla_helper.py
FrNecas: I just checked again and it really seems that component is an array of strings. Perhaps the backend changed but the documentation wasn't updated?

nforro: Maybe it's worth reporting as a bug?

rebasehelper/helpers/bugzilla_helper.py
```diff
@@ -0,0 +1,152 @@
...
+logger: CustomLogger = cast(CustomLogger, logging.getLogger(__name__))
+
+
+class BugzillaHelper:
+
+    """Class for working with Upstream Release Monitoring on bugzilla."""
+
+    REPO_URL = 'https://src.fedoraproject.org/rpms/{}.git'
+    BUGZILLA_REST = 'https://bugzilla.redhat.com/rest/bug'
```
nforro: I would define just base parts of URLs, in the future it could be necessary to use also other API endpoints than just `bug`. Something like:
```python
DIST_GIT_REPO_URL = 'https://src.fedoraproject.org/rpms/'
BUGZILLA_REST_API_URL = 'https://bugzilla.redhat.com/rest/'
```

rebasehelper/helpers/bugzilla_helper.py
```diff
@@ -0,0 +1,152 @@
...
+            RebaseHelperError: If no such bugzilla exists or if the bugzilla
+            is not created by Upstream Release Monitoring.
+
+        """
+        r = requests.get('{}/{}'.format(cls.BUGZILLA_REST, bugzilla_id))
+        if not r.ok:
+            raise RebaseHelperError('Could not obtain data from bugzilla')
+
+        if 'error' in r.json():
+            raise RebaseHelperError('No bugzilla with ID {} was found'.format(bugzilla_id))
```
nforro: I'm pretty sure there are other possible reasons for errors than just this one. Maybe in such cases the actual error message could be logged and the exception raised could be the same generic message as above.

rebasehelper/helpers/bugzilla_helper.py
```diff
@@ -0,0 +1,152 @@
...
+
+        bug = r.json()['bugs'][0]
```
nforro: A nitpick, but I would avoid using `r.json()` twice, the response data have to be parsed twice as well.

rebasehelper/options.py
```diff
@@ -179,6 +179,12 @@
         "switch": True,
         "help": "update \"sources\" file and upload new sources to lookaside cache",
     },
+    {
+        "name": ["--bugzilla-id"],
+        "metavar": "BUG_ID",
+        "default": None,
+        "help": "do a rebase based on Upstream Release Monitoring bugzilla"
+    },
```
nforro: I think this option should be on top of `behavior control` list. Perhaps it would even deserve its own category.

rebasehelper/helpers/bugzilla_helper.py
```diff
@@ -0,0 +1,152 @@
...
+            is supposed to be cloned into, exists.
+
+        """
+        path = os.path.abspath('{}-{}'.format(bugzilla_id, component))
+        if os.path.exists(path):
+            raise RebaseHelperError('Could not clone the repository because the directory '
+                                    '{} already exists'.format(path))
+
+        url = cls.REPO_URL.format(component)
+        logger.info("Cloning %s into %s", url, path)
```
nforro: Verbose output now looks like this:

```
INFO: Cloning https://src.fedoraproject.org/rpms/gpxsee.git into /tmp/1756742-gpxsee
VERBOSE: Rebase-helper version: 0.18.1.dev33+g3b749cd
VERBOSE: argument passed as a new source is a version
```

I think rebase-helper version should always be the very first message.
Also the third message is obviously wrong, but that's unrelated to this PR.

rebasehelper/cli.py
```diff
@@ -124,6 +124,8 @@ def run():
             else:
                 main_handler.setLevel(logging.DEBUG)
             ConsoleHelper.use_colors = ConsoleHelper.should_use_colors(config)
+            logger.verbose("rebase-helper version: %s", VERSION)
```
nforro: Unfortunately this happens before file handers are created, so the version will not be visible in log files.
I think it would make no harm to simply log the version twice, once here and second time just after file handles are created. Also, I think this information is important enough to be present in all the verbosity levels, including the default one.

rebasehelper/helpers/bugzilla_helper.py
```diff
@@ -0,0 +1,155 @@
...
+            is not created by Upstream Release Monitoring.
+
+        """
+        r = requests.get('{}/bug/{}'.format(cls.BUGZILLA_REST_API_URL, bugzilla_id))
+        if not r.ok:
+            raise RebaseHelperError('Could not obtain data from bugzilla')
+
+        response_json = r.json()
+        if 'error' in response_json:
+            logger.error(response_json['error'])
```
nforro: Some context would be nice, something like `Bugzilla error:` prefix.

rebasehelper/cli.py
FrNecas: Perhaps we should log it to level INFO in CLI (to have version as the first message on stdout) and then log to VERBOSE or DEBUG in Application to ensure it gets into the logs? I think that having the version in verbose/debug log twice wouldn't do any harm.

nforro: It won't be in the log files twice. If you mean it will be in both `debug.log` and `verbose.log`, then yes, and I don't see any reason for it not to be also in `info.log`. The version string is now more useful than ever, there is no point it hiding it.

FrNecas: Hmm, I don't think that logging the version twice using the stream handler is a good approach. Also I just now realized that there is more information than this missing from the log files. Warnings ca be logged during `Application.setup` but at the that time, file handlers haven't been created. I think this would be worth fixing but it's out of the scope of this PR

nforro: > Hmm, I don't think that logging the version twice using the stream handler is a good approach.

Well, my idea was to log only to files, but that's not so straightforward to do. Handler-level filters could be used, but that's rather hackish. Best solution would be to buffer the log messages going to stream handlers before file handlers are created, and process the buffer just after their creation. Let's revert this commit, this all deserves its own PR.

rebasehelper/cli.py
```diff
@@ -123,6 +124,12 @@ def run():
             else:
                 main_handler.setLevel(logging.DEBUG)
             ConsoleHelper.use_colors = ConsoleHelper.should_use_colors(config)
+            if config.bugzilla_id:
+                repo_path, config.sources = BugzillaHelper.prepare_rebase_repository(config.bugzilla_id)
+                try:
+                    os.chdir(repo_path)
+                except OSError:
+                    raise RebaseHelperError("Could not change directory into the cloned repository")
```
nforro: I think `to` instead of `into` is more suitable in this context. According to google search, `to` is used about 7 times more often than `into` ;-)
Also, why double quotes? There is no reason to use them here. We should finally agree on a convention where to use which quotes, but that's for another discussion.

rebasehelper/helpers/bugzilla_helper.py
```diff
@@ -0,0 +1,155 @@
...
+
+        Args:
+            bugzilla_id: ID of the bugzilla.
+
+        Returns:
+            Component of the bugzilla.
+
+        Raises:
+            RebaseHelperError: If no such bugzilla exists or if the bugzilla
+            is not created by Upstream Release Monitoring.
```
nforro: I think past tense makes more sense here: `was not created`.

rebasehelper/helpers/bugzilla_helper.py
```diff
@@ -0,0 +1,155 @@
...
+                continue
+            match = pattern.match(comment['text'])
+            if match:
+                version = match.group('version')
+
+        return version
+
+    @classmethod
+    def clone_repository(cls, component: str, bugzilla_id: str) -> str:
+        """Clones the component package repository
```
nforro: This is rather vague and doesn't say much. How about `Clones remote dist-git repository of a component.`?

rebasehelper/cli.py
FrNecas: Good point with the quotes. Our code is inconsistent in this regard and we should change that. Perhaps using `black` or a similar tool could be a good solution? If I remember correctly, `black` uses double quotes for all strings, though I am not very keen on some aspects of the code formatting that `black` produces.

nforro: I'm not inclined to using black. My preference is to use single quotes everywhere and leave double quotes for docstrings. There is also another related question - should we quote paths, names etc. in log messages and exceptions? I think in most cases it's unnecessary, but it would be nice to be consistent in this.

PR #739, opened 2019-10-29
Add option to define macros

FrNecas:


rebasehelper/application.py
```diff
@@ -121,6 +121,7 @@ def __del__(self):
 
     @staticmethod
     def setup(cli_conf):
+        MacroHelper.add_macros(cli_conf.rpmmacros)
```
nforro: No, the option is supposed to populate `SpecFile.predefined_macros`.

FrNecas: You are right, it doesn't work. I tested that the macros get parsed correctly but didn't test it on a real package which could benefit from this option. It should be fixed now

rebasehelper/cli.py
```diff
@@ -133,6 +133,18 @@ def run():
...
                 except OSError:
                     raise RebaseHelperError('Could not change directory to the cloned repository')
+
+            macros = {}
+            for macro in config.rpmmacros:
+                try:
+                    macro = macro.strip()
+                    key, value = macro.split(' ', 1)
+                    if not key.startswith('%'):
+                        key = '%{}'.format(key)
```
nforro: I was going to say this is wrong, but it seems to work with `rpm -D`. RPM guide however says:
> Do not place the leading percent sign, %, on the macro you define with --define.

And I'm not sure about `rpm.addMacro()`. After all, you can't use `%` with `%define` or `%global`, so I would rather stay away from it.

rebasehelper/cli.py
```diff
@@ -133,6 +133,18 @@ def run():
                     os.chdir(repo_path)
                 except OSError:
                     raise RebaseHelperError('Could not change directory to the cloned repository')
+
+            macros = {}
+            for macro in config.rpmmacros:
+                try:
+                    macro = macro.strip()
```
nforro: Why strip?

FrNecas: I think it is safer to remove the leading and trailing whitespaces, not sure how rpm handles them

rebasehelper/cli.py
FrNecas: Do I understand correctly that we shouldn't prepend the percent sign to the macro? What about cases when the user has already prepended the precent sign? Should we strip it?

rebasehelper/cli.py
nforro: What if the whitespaces are there on purpose? It's unlikely, but I don't think there is a reason to do what's rpm's job.

nforro: Ok, but <pre>'macro  expression'</pre> won't work with `split()` either. And both work with `rpm -D`.

FrNecas: I implemented the parsing with a regex instead. That should be more error-prone. It also covers the 2 cases we mentioned. Although there are could be some advanced cases that won't work with this implementation (but work with `rpm -D`), I think this should suffice.

nforro: You mean less error-prone, I hope :-)

FrNecas: Oh, obviously :D

rebasehelper/cli.py
```diff
@@ -103,8 +104,32 @@ def __getattr__(self, name):
...
+        Returns:
+            The converted macros, MACRO are keys, EXPRESSION are values.
+
+        Raises:
+            RebaseHelperError if the macros don't follow the correct format.
+
+        """
+        macros = {}
+        for macro in macros_list:
+            match = re.match(r"\s*%?(?P<macro>\S+)\s+(?P<expr>\w+(\s*\w+)*)", macro)
```
nforro: This is wrong. Macro value can be anything, not just word characters and whitespaces. This won't even match a simple path containing directory separators.

rebasehelper/cli.py
```diff
@@ -103,8 +104,32 @@ def __getattr__(self, name):
...
+        Returns:
+            The converted macros, MACRO are keys, EXPRESSION are values.
+
+        Raises:
+            RebaseHelperError if the macros don't follow the correct format.
+
+        """
+        macros = {}
+        for macro in macros_list:
+            match = re.match(r"\s*%?(?P<macro>\S+)\s+(?P<expr>\S+(\s*\S+)*)", macro)
```
nforro: Why is `\S+(\s*\S+)*` necessary?

FrNecas: There are a few reasons - `.*` can't be used instead, because something like `macro    ` would match but it shouldn't so `\S+` has to be used instead. However, `rpm -D` succeeds even with `macro something1 something2`, which wouldn't work with just single '\S+' (only the first word would be taken). I am not completely positive how rpm handles such case, I assume that it takes the first word as macro name and the rest as the value but I don't know how to confirm it. So I added the optional `\S+` separated with any number of whitespaces.

nforro: How would `macro `  match? There is `\s+`.

FrNecas: Well, I tested it and it seems that with 2 spaces, `.+` takes priority over `\s+` and the regex isn't evaluated from left to right. See [this as an example](https://pythex.org/?regex=%5Cs*%25%3F(%3FP%3Cmacro%3E%5CS%2B)%5Cs%2B(%3FP%3Cexpr%3E.%2B)&test_string=%20macro%20%20%20&ignorecase=1&multiline=0&dotall=0&verbose=0)

nforro: No, that's not what happens, `.+` matches the last space (because of the `+`). How about this?
`\s*%?(?P<macro>\S+)\s+(?P<expr>\S.*)\s*`

PR #740, opened 2019-11-05
Add public API tests

FrNecas:
This covers most of packit's usage, I will add more tests later on but I want to avoid having an incredibly long PR

### DIFF
--- a/rebasehelper/tests/public_api/test_specfile.py
+++ b/rebasehelper/tests/public_api/test_specfile.py
@@ -1,0 +1,75 @@
+# -*- coding: utf-8 -*-
+#
+# This tool helps you rebase your package to the latest version
+# Copyright (C) 2013-2019 Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Authors: Petr Hráček <phracek@redhat.com>
+#          Tomáš Hozza <thozza@redhat.com>
+#          Nikola Forró <nforro@redhat.com>
+#          František Nečas <fifinecas@seznam.cz>
+
+import os
+import shutil
+
+import pytest  # type: ignore
+
+from rebasehelper.specfile import SpecFile
+from rebasehelper.tests.conftest import SPEC_FILE, TEST_FILES_DIR
+
+
+@pytest.mark.public_api
+class TestSpecFile:
+    def test_constructor(self, workdir):
+        shutil.copy(os.path.join(TEST_FILES_DIR, SPEC_FILE), workdir)
+        # check that no mandatory argument was added
+        spec = SpecFile(SPEC_FILE)
+        assert isinstance(spec, SpecFile)
+        # check that arguments were not renamed or removed
+        spec = SpecFile(path=SPEC_FILE, sources_location=workdir, predefined_macros=None)
+        assert isinstance(spec, SpecFile)
+
+    def test_update_changelog(self, spec_object):
+        assert spec_object.update_changelog('test2') is None
+        assert spec_object.update_changelog(changelog_entry='test') is None
+
+    def test_set_version(self, spec_object):
+        assert spec_object.set_version('1.2.3.4') is None
+        assert spec_object.set_version(version='1.2.3') is None
+
+    def test_get_version(self, spec_object):
+        assert isinstance(spec_object.get_version(), str)
+
+    def test_set_release_number(self, spec_object):
+        assert spec_object.set_release_number('2') is None
+        assert spec_object.set_release_number(release='1') is None
+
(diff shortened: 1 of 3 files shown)


FrNecas:
Something doesn't seem alright. Linty and Travis on my last push triggered roughly half an hour later than it should have, it hasn't triggered yet on this force push.

rebasehelper/tests/public_api/test_specfile.py
```diff
@@ -0,0 +1,87 @@
...
+        assert entry in spec_object.spec_content.section('%changelog')[1]
+
+    def test_set_version(self, spec_object):
+        NEW_VERSION = '1.2.3.4.5'
+        spec_object.set_version(version=NEW_VERSION)
+        assert spec_object.header.version == NEW_VERSION
+
+        NEW_VERSION = '1.2.3.4'
+        spec_object.set_version(NEW_VERSION)
+        assert spec_object.header.version == NEW_VERSION
```
nforro: Why the inconsistencies? Obviously version has to be different so the change can be verified, but the same applies to changelog entry. If the second call to `update_changelog()` doesn't generate a new entry at all, the second assert wouldn't catch it.

rebasehelper/tests/public_api/test_specfile.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def test_get_version(self, spec_object):
+        assert spec_object.get_version() == '1.0.2'
+
+    def test_set_release_number(self, spec_object):
+        spec_object.set_release_number(release=1)
+        assert '1' == spec_object.get_release()
```
nforro: Why the sudden swap of operators?

rebasehelper/tests/public_api/test_specfile.py
```diff
@@ -0,0 +1,87 @@
...
+
+        NEW_VERSION = '1.2.3.4'
+        spec_object.set_version(NEW_VERSION)
+        assert spec_object.header.version == NEW_VERSION
+
+    def test_get_version(self, spec_object):
+        assert spec_object.get_version() == '1.0.2'
+
+    def test_set_release_number(self, spec_object):
+        spec_object.set_release_number(release=1)
```
nforro: `release` should be of type `str`. If packit passes `int` there, we should fix the declaration of `set_release_number()`. But `set_release()` only accepts `str`.

rebasehelper/tests/public_api/test_specfile.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def test_set_release(self, spec_object):
+        spec_object.set_release(release=1)
+        assert '1' == spec_object.get_release()
+
+        spec_object.set_release(2)
+        assert '2' == spec_object.get_release()
+
+    def test_get_sources(self, spec_object):
+        assert isinstance(spec_object.get_sources(), list)
```
nforro: Again, an inconsistency. You check exact value everywhere else, but only type here. It should be either this or that. I think checking type should be enough, it will make the tests less dependent on potential changes in the testing spec file (but perhaps it would be best to make a separate spec file for API tests).

rebasehelper/tests/public_api/test_specfile.py
FrNecas: I just checked and packit uses string as well. I took this from our unit tests - they are wrong as it seems, I will fix it there as well

tox.ini
```diff
@@ -3,12 +3,13 @@ envlist=lint, py{36,37,38}
 skipsdist=True
 
 [pytest]
-addopts=-m 'standard and not integration and not long_running'
+addopts=-m 'standard and not integration and not long_running and not public_api'
```
nforro: Why do we not want to test this every time? I don't think it's that time consuming, or is it?

rebasehelper/tests/public_api/test_specfile.py
FrNecas: Hm, the thing is that with some methods (`update_changelog`) you can't really only check the type. The same goes for `set_release_number` and `set_version`. With these methods, we probably want t ocheck that the release was really set. I could check the exact value here as well but that already gets taken care of by our unit tests so it would be pointless. There are a lot of sources so listing them all out is inconvenient (this could be solved by having a separate testing spec)

nforro: > you can't really only check the type

Yes, you can. These are API (interface) tests, they are supposed to complement unit/functional tests that test the actual functionality.

FrNecas: Okay, so lets just strictly check the return value type and the arguments in this test suite, that will probably be better

nforro: While not strictly necessary, it would be more consistent to check return type of every method, even those that return `None` or instance of `SpecFile` (the constructor).

rebasehelper/tests/public_api/test_specfile.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+@pytest.mark.public_api
+class TestSpecFile:
+    def test_constructor(self, workdir):
+        shutil.copy(os.path.join(TEST_FILES_DIR, SPEC_FILE), workdir)
+        # check that no mandatory argument was added
+        assert isinstance(SpecFile(SPEC_FILE), SpecFile)
+        # check that arguments were not renamed or removed
+        assert isinstance(SpecFile(path=SPEC_FILE, sources_location=workdir, predefined_macros=None), SpecFile)
```
nforro: I would split this into 2 lines for the sake of readability.

rebasehelper/tests/public_api/test_specfile.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def test_update_changelog(self, spec_object):
+        assert not spec_object.update_changelog('test2')
```
nforro: `not` doesn't assert that the value is `None`, it could be for example `[]` or `''` or even `0`.

PR #749, opened 2019-11-14
Take subpackages and non-unique tags into account

nforro:


rebasehelper/specfile.py
```diff
@@ -352,31 +352,36 @@ def sanitize(tag):
                     return '{0}0'.format(tokens[0])
                 return '{0}{1}'.format(tokens[0], int(tokens[1]))
             return tag.capitalize()
-        result = {}
-        parsed = self.spc.parsed.split('\n')
+        result: Dict[str, List[Tag]] = {}
+        parsed = self.spc.parsed.split('\n')  # FIXME: we should split this on sections too
```
FrNecas: Do I understand correctly that the FIXME talks about handling conditionalized sections correctly? If so, the fix should be fairly simple. Checking whether the section header ( the first element in each element of `SpecContent.sections`) is in `parsed` should work in this case. Perhaps it would be better to create a `set` from `parsed` to achieve better time performance

nforro: No, it's about finding matching tags in the right section. There could be duplicate tags in different sections and the validity check could be inaccurate otherwise.

rebasehelper/specfile.py
```diff
@@ -352,47 +352,78 @@ def sanitize(tag):
...
-        index, span = self.tags[tag]
-        line = self.spec_content.section('%package')[index]
-        self.spec_content.section('%package')[index] = line[:span[0]] + value + line[span[1]:]
+        section = self._get_matching_section(subpackage)
+        tag_name, lineno, span = tag
+        line = self.spec_content.section(section)[lineno]
+        self.spec_content.section(section)[lineno] = line[:span[0]] + value + line[span[1]:]
         # update span
-        self.tags[tag] = (index, (span[0], span[0] + len(value)))
+        self.tags(subpackage)[0] = (tag_name, lineno, (span[0], span[0] + len(value)))
```
FrNecas: Why `self.tags(subpackage)[0]`? Unless I am missing something, `self.tags` returns a list of tags (tuples). The tag to be modified can be on any index in the list, can't it?

nforro: No, because `SpecFile.tag()` returns only the first one. It a bit hackish, but I don't see any other solution than returning also index from `SpecFile.tag()` (that would defeat the simplicity) or reimplementing its functionality here.

FrNecas: I understand that `SpecFile.tag()` returns only the first occurrence but this line that modifies the span calls `SpecFile.tags()`. Shouldn't it then rather be `self.tag(tag_name, subpackage)[0] = ...`

nforro: No, the code is replacing the first item in the list returned from `SpecFile.tags()`. `SpecFile.tag()` returns a tuple, that is, one, immutable, and two, even if it was mutable the change wouldn't propagate back to the list.

PR #750, opened 2019-11-15
Rework spec tags

nforro:


rebasehelper/spec_content.py
```diff
@@ -0,0 +1,166 @@
+# -*- coding: utf-8 -*-
```
FrNecas: You should also probably move its tests to a separate file

rebasehelper/specfile.py
```diff
@@ -605,7 +529,7 @@ def set_extra_version(self, extra_version: Optional[str], version_changed: bool)
         # TODO: in some cases it might be necessary to modify Source0
 
     @saves
-    def set_tag(self, tag: str, value: str, preserve_macros: bool = False, subpackage: Optional[str] = None) -> None:
```
FrNecas: Why don't we need the subpackage? What if we wanted to alter a tag in a subpackage, this wouldn't be possible after the removal, right?

nforro: Feel free to adjust `set_tag()` to work with subpackages, but I don't think it's worth it. What would be the use-case?

FrNecas: Makes sense, there is probably no point in having `set_tag` work with subpackages, at least for now

rebasehelper/spec_content.py
nforro: Done. I've also added tests for Tags.

rebasehelper/tests/test_spec_content.py
```diff
@@ -0,0 +1,72 @@
...
+#          Nikola Forró <nforro@redhat.com>
+#          František Nečas <fifinecas@seznam.cz>
+
+import pytest  # type: ignore
+
+from rebasehelper.spec_content import SpecContent
+from rebasehelper.tests.conftest import SPEC_FILE
+
+
+class TestSpecContent:
```
FrNecas: There are also public API tests for SpecContent. Although they are quite short, I think they should be moved as well for the sake of consistency.

Issue #730, opened 2019-10-11
ReplaceOldVersion hook must ignore versioned obsoletes and most likely also other dependency tags


nforro:
See for example: https://src.fedoraproject.org/rpms/protobuf/blob/master/f/protobuf.spec#_45

FrNecas:
I assume there are quite a lot of these tags (Conflicts, Requires, Recommends, etc). I would say that ignoring the whole preamble except for Patch/Source tags may be the best solution to this. Can you see any possible drawbacks?

nforro:
This is still an issue if a tag [is defined by a macro](https://src.fedoraproject.org/rpms/fedora-obsolete-packages/blob/master/f/fedora-obsolete-packages.spec#_148).

The fix should be simple, do the check after macro expansion.

nforro:
Not directly related to the original issue, but the hook assumes that there is a whitespace separating a tag from its value, and tracebacks when there is none.
I think the code should special-case preamble and avoid enumerating and parsing tags and use `SpecFile.tags` instead.

FrNecas:
I don't think that expanding the line before doing the check will do the trick here. Apparently, `%obsolete` expands to `%obsoletes`, see

```
>>> MacroHelper.expand("%obsolete gegl 0.2.0-41")
'%obsolete gegl 0.2.0-41'
```

I will rewrite the hook and use `SpecFile.tags` for preamble, though I am not sure how we are going to handle the problem with dependencies defined by macros.

nforro:
> Apparently, %obsolete expands to `%obsoletes`

What? It expands to itself, which means it doesn't expand at all, which happens I suppose because you are not doing it in the context of the spec file that defines the macro.

FrNecas:
Oh, my bad, I thought that it is a macro defined by rpm. Now I see it is defined in the spec file so expanding will work.

PR #751, opened 2019-11-19
ReplaceOldVersion fixes

FrNecas:
Fixes: #730

### DIFF
--- a/rebasehelper/plugins/spec_hooks/replace_old_version.py
+++ b/rebasehelper/plugins/spec_hooks/replace_old_version.py
@@ -22,9 +22,10 @@
 #          Nikola Forró <nforro@redhat.com>
 #          František Nečas <fifinecas@seznam.cz>
 
+import collections
 import re
 import urllib.parse
-from typing import Any, List, Pattern, Tuple
+from typing import Any, Dict, List, Pattern, Tuple, Set
 
 from rebasehelper.types import Options
 from rebasehelper.plugins.spec_hooks import BaseSpecHook
@@ -58,22 +59,6 @@ class ReplaceOldVersion(BaseSpecHook):
         # e.g. in a rebase from 2.5 to 2.5.1, version would be changed to 2.5.1.1
         'Version',
     ]
-
-    @classmethod
-    def _is_local_source(cls, line: str) -> bool:
-        """Checks if a line contains a local source.
-
-        Args:
-            line: Line to be checked.
-
-        Returns:
-            Whether the line contains a local source
-
-        """
-        if not (line.startswith('Patch') or line.startswith('Source')):
-            return False
-        source = line.split()[1]
-        return not urllib.parse.urlparse(source).scheme
 
     @classmethod
     def _create_possible_replacements(cls, old: str, new: str, use_macro: bool) -> List[Tuple[Pattern[str], str]]:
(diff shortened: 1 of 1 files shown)


rebasehelper/plugins/spec_hooks/replace_old_version.py
```diff
@@ -114,18 +98,33 @@ def run(cls, spec_file: SpecFile, rebase_spec_file: SpecFile, **kwargs: Any):
         replace_with_macro = bool(kwargs.get('replace_old_version_with_macro'))
 
         subversion_patterns = cls._create_possible_replacements(old_version, new_version, replace_with_macro)
+        examined_lines = set()
+        for tag in rebase_spec_file.tags.filter():
+            examined_lines.add(tag.line)
+            value = rebase_spec_file.spec_content.section(tag.section)[tag.line][slice(*tag.value_span)]
```
nforro: Why don't you use `SpecFile.get_raw_tag_value()`?

rebasehelper/plugins/spec_hooks/replace_old_version.py
```diff
@@ -114,18 +98,33 @@ def run(cls, spec_file: SpecFile, rebase_spec_file: SpecFile, **kwargs: Any):
...
+                continue
+            scheme = urllib.parse.urlparse(value).scheme
+            if (tag.name.startswith('Patch') or tag.name.startswith('Source')) and not scheme:
+                # skip local sources
+                continue
+
+            # replace the whole version first
+            updated_line = subversion_patterns[0][0].sub(subversion_patterns[0][1], value)
+            # replace subversions only for remote sources/patches
+            if (tag.name.startswith('Patch') or tag.name.startswith('Source')) and urllib.parse.urlparse(updated_line):
```
nforro: This condition seems rather redundant, since local sources are already skipped.

rebasehelper/plugins/spec_hooks/replace_old_version.py
```diff
@@ -114,18 +98,33 @@ def run(cls, spec_file: SpecFile, rebase_spec_file: SpecFile, **kwargs: Any):
...
+                for sub_pattern, repl in subversion_patterns[1:]:
+                    updated_line = sub_pattern.sub(repl, updated_line)
+            rebase_spec_file.set_raw_tag_value(tag.name, updated_line, tag.section)
+
         for sec_name, section in rebase_spec_file.spec_content.sections:
             if sec_name.startswith('%changelog'):
                 continue
             for index, line in enumerate(section):
-                if cls._is_local_source(line) or any(line.startswith(tag) for tag in cls.IGNORED_TAGS):
+                if index in examined_lines:
```
nforro: How is this supposed to work? `examined_lines` contains line numbers from preamble, but you are not checking section here at all.

rebasehelper/plugins/spec_hooks/replace_old_version.py
```diff
@@ -114,18 +99,33 @@ def run(cls, spec_file: SpecFile, rebase_spec_file: SpecFile, **kwargs: Any):
         replace_with_macro = bool(kwargs.get('replace_old_version_with_macro'))
 
         subversion_patterns = cls._create_possible_replacements(old_version, new_version, replace_with_macro)
+        examined_lines = collections.defaultdict(set)
+        for tag in rebase_spec_file.tags.filter():
```
nforro: This still iterates only through preamble. Is that intentional?

FrNecas: Oh, it defaults to '%package'. Is there an easy way how to iterate over all tags or do I have to use `Tags.items` attribute to do so?

rebasehelper/plugins/spec_hooks/replace_old_version.py
```diff
@@ -114,18 +99,33 @@ def run(cls, spec_file: SpecFile, rebase_spec_file: SpecFile, **kwargs: Any):
         replace_with_macro = bool(kwargs.get('replace_old_version_with_macro'))
 
         subversion_patterns = cls._create_possible_replacements(old_version, new_version, replace_with_macro)
+        examined_lines = collections.defaultdict(set)
+        for tag in rebase_spec_file.tags.items:
```
nforro: First, `.items` is completely unnecessary and shouldn't be used. Second, you are now including also invalid tags. I would use `rebase_spec_file.tags.filter(section=None)`.

rebasehelper/plugins/spec_hooks/replace_old_version.py
```diff
@@ -122,7 +123,8 @@ def run(cls, spec_file: SpecFile, rebase_spec_file: SpecFile, **kwargs: Any):
             if sec_name.startswith('%changelog'):
                 continue
             for index, line in enumerate(section):
-                if index in examined_lines[sec_name] or any(line.startswith(tag) for tag in cls.IGNORED_TAGS):
+                tag_ignored = any(line.startswith(MacroHelper.expand(tag)) for tag in cls.IGNORED_TAGS)
```
nforro: This change has no effect whatsoever. Expanding a tag name does nothing.
Also, using `MacroHelper.expand()` without `default` argument (and not handling `None` return value) is going to lead to a traceback sooner or later.

PR #754, opened 2019-12-04
Fix mypy errors for PluginCollection

FrNecas:


FrNecas:
I will rework the type-hinting in plugins a bit more as there are still some problems. Other collections have to use `cast` the same way as `CheckerCollection` does. it doesn't break tests only due to not being type-hinted. But for now, lets merge this so we can merge other PRs

PR #755, opened 2019-12-06
Release 0.20.0

nforro:


CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
 
 ## [Unreleased]
 
+## [0.20.0] - 2019-12-06
+### Added
+- Introduced `Tags` class unifying and simplifying access to SPEC tags
+- Added proper support for *crate* and *gem* archives
+- Added `--bugzilla-id` option to perform a rebase based on *Upstream Release Monitoring* bugzilla
+- Added option to define macros
```
FrNecas: I think you should also add the specific option (`-D`) to stay consistent with the line above.

nforro: But what about `--define`? :wink:

nforro: But what about `--define`? :wink:

CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
...
+- Added `-D`/`--define` option to define macros
+- Added tests for public API
+
+### Fixed
+- Fixed `--build-tasks` option
+- Fixed detecting unresolved conflicts in non-UTF-8 files
+- Prevented loss of messages logged before logging file handlers are created
+- **rebase-helper** now skips unparseable lines in *%prep* instead of tracebacking on them
+- Fixed parsing SPEC files with `-h` in *%prep*
+- Fixed processing SPEC files with zero-padded indexed tags and `%patch` macros
```
FrNecas: Shouldn't it rather be `processing of SPEC files`? The same goes for `parsing of SPEC files` and perhaps even for `detecting`.

nforro: Why?

FrNecas: It just sounds right to me. Having a look at examples for example here: https://dictionary.cambridge.org/dictionary/english/processing confirms that. I think that it is applicable to all 3 of the verbs

nforro: I see for example "parsing" being used without "of" all the time, way more often than with it.
Also, both processing and parsing are used this way multiple times in the changelog, so it would be inconsistent to change the style now.

Issue #622, opened 2019-05-27
Support new features of rpm 4.15


nforro:
Namely `%patchlist` and `%sourcelist` spec sections and automatic patch and source numbering.

nforro:
Also new `%generate_buildrequires` section.

PR #756, opened 2019-12-12
Implement parsing of %sourcelist and %patchlist

FrNecas:
Related: #622 (this just implements parsing)

This just adds parsing support, integration with other parts of rebase-helper may not be perfect. For example if rebase-helper removes a patch from preamble and `%patchlist` is present and the patches are applied in `%prep` using `%patch` macro, it is going to break. But I think this deserves its own issue/PR.

Also, I haven't found a way how to support conditionalized sources in `%sourcelist` or `%patchlist`. Say the section looks like this:

```
%sourcelist
source2.tar.gz
%if 0%{?source:1}
source3.tar
%endif
```

I don't think there is a consistent way to differentiate between the source and the `%if` macro (neither are in parsed spec object).

### DIFF
--- a/rebasehelper/plugins/spec_hooks/replace_old_version.py
+++ b/rebasehelper/plugins/spec_hooks/replace_old_version.py
@@ -100,10 +100,10 @@ class ReplaceOldVersion(BaseSpecHook):
         replace_with_macro = bool(kwargs.get('replace_old_version_with_macro'))
 
         subversion_patterns = cls._create_possible_replacements(old_version, new_version, replace_with_macro)
-        examined_lines: Dict[str, Set[int]] = collections.defaultdict(set)
+        examined_lines: Dict[int, Set[int]] = collections.defaultdict(set)
         for tag in rebase_spec_file.tags.filter():
-            examined_lines[tag.section].add(tag.line)
-            value = rebase_spec_file.get_raw_tag_value(tag.name, tag.section)
+            examined_lines[tag.section_index].add(tag.line)
+            value = rebase_spec_file.get_raw_tag_value(tag.name, tag.section_index)
             if not value or tag.name in cls.IGNORED_TAGS:
                 continue
             scheme = urllib.parse.urlparse(value).scheme
@@ -117,14 +117,14 @@ class ReplaceOldVersion(BaseSpecHook):
             if tag.name.startswith('Patch') or tag.name.startswith('Source'):
                 for sub_pattern, repl in subversion_patterns[1:]:
                     updated_value = sub_pattern.sub(repl, updated_value)
-            rebase_spec_file.set_raw_tag_value(tag.name, updated_value, tag.section)
+            rebase_spec_file.set_raw_tag_value(tag.name, updated_value, tag.section_index)
 
-        for sec_name, section in rebase_spec_file.spec_content.sections:
+        for sec_index, (sec_name, section) in enumerate(rebase_spec_file.spec_content.sections):
             if sec_name.startswith('%changelog'):
                 continue
             for index, line in enumerate(section):
                 tag_ignored = any(MacroHelper.expand(line, line).startswith(tag) for tag in cls.IGNORED_TAGS)
-                if index in examined_lines[sec_name] or tag_ignored:
+                if index in examined_lines[sec_index] or tag_ignored:
                     continue
                 start, end = spec_file.spec_content.get_comment_span(line, sec_name)
                 updated_line = subversion_patterns[0][0].sub(subversion_patterns[0][1], line[:start])
--- a/rebasehelper/spec_content.py
+++ b/rebasehelper/spec_content.py
@@ -87,6 +87,9 @@ class SpecContent:
                 content.append(line + '\n')
         return ''.join(content)
 
+    def __getitem__(self, index):
+        return self.sections[index][1]
+
     @classmethod
     def get_comment_span(cls, line: str, section: str) -> Tuple[int, int]:
         """Gets span of a comment depending on the section.
--- a/rebasehelper/spec_content.py
+++ b/rebasehelper/spec_content.py
@@ -62,6 +62,9 @@ class SpecContent:
         '%transfiletrigger',
         '%transfiletriggerun',
         '%transfiletriggerpostun',
+        '%end',
+        '%patchlist',
+        '%sourcelist',
     ]
 
     # Comments in these sections can only be on a separate line.
--- a/rebasehelper/specfile.py
+++ b/rebasehelper/specfile.py
@@ -23,6 +23,7 @@
 #          František Nečas <fifinecas@seznam.cz>
 
 import argparse
+import collections
 import enum
 import itertools
 import logging
@@ -426,9 +427,11 @@ class SpecFile:
         removed_patches = []
         inapplicable_patches = []
         modified_patches = []
-        preamble = self.spec_content.section('%package')
-        remove_lines = []
+        remove_lines: Dict[int, List[Tuple[int, int]]] = collections.defaultdict(list)
         for tag in self.tags.filter(name='Patch*'):
+            section = self.spec_content[tag.section_index]
+            if section is None:
+                continue
             patch_num = int(tag.name.split('Patch')[1])
             patch_name = self.get_raw_tag_value(tag.name) or ''
             if 'deleted' in patches:
@@ -445,14 +448,14 @@ class SpecFile:
                 removed_patches.append(patch_num)
                 # find associated comments
                 i = tag.line
-                while i > 0 and is_comment(preamble[i - 1]):
+                while i > 0 and is_comment(section[i - 1]):
                     i -= 1
-                remove_lines.append((i, tag.line + 1))
+                remove_lines[tag.section_index].append((i, tag.line + 1))
                 continue
             if patch_inapplicable:
                 if disable_inapplicable:
                     # comment out line if the patch was not applied
-                    preamble[tag.line] = '#' + preamble[tag.line]
+                    section[tag.line] = '#' + section[tag.line]
                 inapplicable_patches.append(patch_num)
             if 'modified' in patches:
                 patch = [x for x in patches['modified'] if patch_name in x]
@@ -462,8 +465,10 @@ class SpecFile:
                 name = os.path.join(constants.RESULTS_DIR, constants.REBASED_SOURCES_DIR, patch_name)
                 self.set_raw_tag_value(tag.name, name)
                 modified_patches.append(patch_num)
-        for span in sorted(remove_lines, key=lambda s: s[0], reverse=True):
-            del preamble[slice(*span)]
+        for section_index, remove in remove_lines.items():
+            content = self.spec_content[section_index]
+            for span in sorted(remove, key=lambda s: s[0], reverse=True):
+                del content[slice(*span)]
         self._process_patches(inapplicable_patches, removed_patches, disable_inapplicable)
 
     ###################################
--- a/rebasehelper/specfile.py
+++ b/rebasehelper/specfile.py
@@ -34,7 +34,7 @@ import urllib.parse
 from datetime import date
 from difflib import SequenceMatcher
 from operator import itemgetter
-from typing import List, Optional, Pattern, Tuple, Dict, cast
+from typing import List, Optional, Pattern, Tuple, Dict, Union, cast
 
 import rpm  # type: ignore
 
@@ -195,23 +195,27 @@ class SpecFile:
     # TAG HELPER METHODS #
     ######################
 
-    def tag(self, name: str, section: Optional[str] = None) -> Optional[Tag]:
+    def tag(self, name: str, section: Optional[Union[str, int]] = None) -> Optional[Tag]:
         """Returns the first non-unique tag."""
-        tags = self.tags.filter(section=section, name=name)
+        if isinstance(section, str):
+            tags = self.tags.filter(section_name=section, name=name)
+        else:
+            tags = self.tags.filter(section_index=section, name=name)
         return next(tags, None)
 
-    def get_raw_tag_value(self, tag_name: str, section: Optional[str] = None) -> Optional[str]:
+    def get_raw_tag_value(self, tag_name: str, section: Optional[Union[str, int]] = None) -> Optional[str]:
         tag = self.tag(tag_name, section)
         if not tag:
             return None
-        return self.spec_content.section(tag.section)[tag.line][slice(*tag.value_span)]
+        return self.spec_content[tag.section_index][tag.line][slice(*tag.value_span)]
 
-    def set_raw_tag_value(self, tag_name: str, value: str, section: Optional[str] = None) -> None:
+    def set_raw_tag_value(self, tag_name: str, value: str, section: Optional[Union[str, int]] = None) -> None:
         tag = self.tag(tag_name, section)
         if not tag:
             return
-        line = self.spec_content.section(tag.section)[tag.line]
-        self.spec_content.section(tag.section)[tag.line] = line[:tag.value_span[0]] + value + line[tag.value_span[1]:]
+        sec = self.spec_content[tag.section_index]
+        line = sec[tag.line]
+        sec[tag.line] = line[:tag.value_span[0]] + value + line[tag.value_span[1]:]
         # update span
         tag.value_span = (tag.value_span[0], tag.value_span[0] + len(value))
 
--- a/rebasehelper/tags.py
+++ b/rebasehelper/tags.py
@@ -25,22 +25,25 @@
 import collections.abc
 import fnmatch
 import re
-from typing import Iterator, Optional, Tuple, cast
+from typing import Dict, Iterator, List, Optional, Tuple, cast
 
 from rebasehelper.spec_content import SpecContent
 from rebasehelper.helpers.macro_helper import MacroHelper
 
 
 class Tag:
-    def __init__(self, section: str, line: int, name: str, value_span: Tuple[int, int], valid: bool) -> None:
-        self.section: str = section
+    def __init__(self, section_index: int, section_name: str, line: int, name: str, value_span: Tuple[int, int],
+                 valid: bool) -> None:
+        self.section_index: int = section_index
+        self.section_name: str = section_name
         self.line: int = line
         self.name: str = name
         self.value_span: Tuple[int, int] = value_span
         self.valid: bool = valid
 
     def __eq__(self, other):
-        return (self.section == other.section and
+        return (self.section_index == other.section_index and
+                self.section_name == other.section_name and
                 self.line == other.line and
                 self.name == other.name and
                 self.value_span == other.value_span and
@@ -56,6 +59,26 @@ class Tags(collections.abc.Sequence):
 
     def __len__(self):
         return len(self.items)
+
+    @classmethod
+    def _map_sections_to_parsed(cls, raw_content: SpecContent, parsed_content: SpecContent) -> Dict[int, int]:
+        """Creates a mapping of raw sections to parsed sections.
+
+        Returns:
+            A dict where the keys are indexes of sections in raw content and values are their
+            counterparts in parsed content. Value is set to -1 if the section is not in
+            parsed SpecContent.
+        """
+        result: Dict[int, int] = {}
+        parsed = 0
+        for index, (section, _) in enumerate(raw_content.sections):
+            if parsed < len(parsed_content.sections) and parsed_content.sections[parsed][0].lower() == section.lower():
+                result[index] = parsed
+                parsed += 1
+            else:
+                result[index] = -1
+
+        return result
 
     @classmethod
     def _parse(cls, raw_content: SpecContent, parsed_content: SpecContent) -> Tuple[Tag]:
@@ -85,20 +108,23 @@ class Tags(collections.abc.Sequence):
                 return '{0}{1}'.format(tokens[0], int(tokens[1]))
             return name.capitalize()
         result = []
+        parsed_mapping = cls._map_sections_to_parsed(raw_content, parsed_content)
         tag_re = re.compile(r'^(?P<prefix>(?P<name>\w+)\s*:\s*)(?P<value>.+)$')
-        for section, _ in raw_content.sections:
+        for section_index, (section, section_content) in enumerate(raw_content.sections):
             section = section.lower()
             if not section.startswith('%package'):
                 continue
-            parsed = parsed_content.section(section)
-            for index, line in enumerate(raw_content.section(section)):
+            parsed_section_index = parsed_mapping[section_index]
+            parsed: List[str] = [] if parsed_section_index == -1 else parsed_content.sections[parsed_section_index][1]
+            for line_index, line in enumerate(section_content):
                 expanded = MacroHelper.expand(line)
                 if not line or not expanded:
                     continue
                 valid = bool(parsed and [p for p in parsed if p == expanded.rstrip()])
                 m = tag_re.match(line)
                 if m:
-                    result.append(Tag(section, index, sanitize(m.group('name')), m.span('value'), valid))
+                    result.append(Tag(section_index, section, line_index, sanitize(m.group('name')), m.span('value'),
+                                      valid))
                     continue
                 m = tag_re.match(expanded)
                 if m:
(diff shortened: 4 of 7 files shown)


rebasehelper/tags.py
```diff
@@ -86,12 +86,17 @@ def sanitize(name):
...
             section = section.lower()
             if not section.startswith('%package'):
                 continue
-            parsed = parsed_content.section(section)
-            for index, line in enumerate(raw_content.section(section)):
+            if parsed_idx >= len(parsed_content.sections) or section != parsed_content.sections[parsed_idx][0].lower():
+                parsed: List[str] = []
+            else:
+                parsed = parsed_content.sections[parsed_idx][1]
+                parsed_idx += 1
```
nforro: I don't like this at all. First, it's not obvious what it actually does, second, it's not very robust. I think you should create a mapping between raw and parsed sections in advance, and do lookups in it.

rebasehelper/tags.py
```diff
@@ -84,26 +116,38 @@ def sanitize(name):
...
+                section_content = cls._prepend_list_with_tags(section, section_content[:], parsed, 'Patch',
+                                                              cast(Tuple[Tag], tuple(result)))
             for index, line in enumerate(section_content):
                 expanded = MacroHelper.expand(line)
                 if not line or not expanded:
                     continue
-                valid = bool(parsed and [p for p in parsed if p == expanded.rstrip()])
+                # There doesn't seem to be a way to parse conditionalized tags from sourcelist
+                # and patchlist. Just ignore such tags and consider only valid ones (this gets
+                # taken care of in _prepend_list_with_tags).
```
nforro: I don't understand this justification at all. Can you elaborate?

rebasehelper/tags.py
```diff
@@ -84,26 +116,38 @@ def sanitize(name):
...
             else:
                 parsed = parsed_content.sections[parsed_idx][1]
                 parsed_idx += 1
+            if section.startswith('%sourcelist'):
+                # copy the list, we do not want to permanently modify the content
+                section_content = cls._prepend_list_with_tags(section, section_content[:], parsed, 'Source',
+                                                              cast(Tuple[Tag], tuple(result)))
+            elif section.startswith('%patchlist'):
+                section_content = cls._prepend_list_with_tags(section, section_content[:], parsed, 'Patch',
+                                                              cast(Tuple[Tag], tuple(result)))
```
nforro: I don't like this. I think it would be better to introduce a method that would, based on `result`, return the new index, and modify the content with a simple list comprehension right here.

rebasehelper/tags.py
```diff
@@ -115,7 +159,10 @@ def sanitize(name):
                     line = line[start:].rstrip('}')  # FIXME: removing trailing braces is not very robust
                     m = tag_re.match(line)
                     if m:
-                        span = cast(Tuple[int, int], tuple(x + start for x in m.span('value')))
+                        if is_list:
+                            span = (0, len(raw_content.sections[section_index][index]) - 1)
+                        else:
+                            span = cast(Tuple[int, int], tuple(x + start for x in m.span('value')))
```
nforro: This (and the hunk above) is just too complicated. It would be better to introduce an offset variable, that would default to 0, and was set to the length of the current prefix in case something was prepended.

rebasehelper/tags.py
FrNecas: There is an example in the text of this PR. Lets say that the section looks like this:

```
%sourcelist
source2.tar.gz
%if 0%{?source:1}
source3.tar
%endif
```

Obviously if the macro `source` is not defined, then the condition won't be satisfied and the source won't be in parsed spec content. Currently Tags also parse conditionalized macros in preamble but I don't think that the same can be done for %sourcelist/%patchlist. We can't automatically prepend all lines because then the `%if` statement would be counted as Source. But then there is no way to parse `source3.tar` as a `Source` tag

nforro: > Obviously if the macro `source` is not defined, then the condition won't be satisfied and the source won't be in parsed spec content.

Exactly, that's the whole point.

> We can't automatically prepend all lines because then the `%if` statement would be counted as Source.

Then just preprend only the lines that are in `parsed`.

nforro: But I can see this prepending thing is getting more and more complicated, maybe it would be better to forget about it and just special case the list sections and implement their own parsing.

FrNecas: > Then just preprend only the lines that are in `parsed`. Or, which would be probably better, only prepend the lines that don't expand to nothing.

That's what I am doing. I was just trying to point out that unlike with tags in `%package` sections (where even conditionalized tags are parsed but with `valid=False`), the tags that aren't in `parsed` are ignored. I think that both should be use (check whether it actually expands to something and whether it is in `parsed`).

As for your second point - I thought about that and even started working on a solution before starting the "prepend" approach. I don't think that it would help. If we wanted to have a robust parsing, it would result in a lot of code duplication which would be quite hard to convert into common logic (at least from what I've tried)

nforro: > where even conditionalized tags are parsed but with `valid=False`

That's not the case for tags defined in conditional macros though. Those are ignored as well.

> If we wanted to have a robust parsing, it would result in a lot of code duplication which would be quite hard to convert into common logic (at least from what I've tried)

Yes, there would be a lot of duplicated code, but isn't that better than having a single codebase but with special-casing almost everywhere?

FrNecas: It probably is. Also it won't be that bad considering all of the above. We can keep the code for `%package` sections as it is and for `%sourcelist`/`%patchlist` it should just be a matter of going through all the lines and creating tags from the valid lines.

rebasehelper/tags.py
```diff
@@ -57,6 +57,27 @@ def __getitem__(self, index):
...
+
+        Returns:
+            A dict where the keys are indexes of sections in raw content and values are their
+            counterparts in parsed content. Value is set to -1 if the section is not in
+            parsed SpecContent.
+        """
+        result: Dict[int, int] = {}
+        parsed = 0
+        for index, (section, _) in enumerate(raw_content.sections):
+            if parsed >= len(parsed_content.sections) or parsed_content.sections[parsed][0].lower() != section.lower():
```
nforro: If the first condition is true, what's the point in continuing the iteration?

rebasehelper/tags.py
```diff
@@ -80,7 +80,52 @@ def _map_sections_to_parsed(cls, raw_content: SpecContent, parsed_content: SpecC
...
+        result = []
+        parsed_mapping = cls._map_sections_to_parsed(raw_content, parsed_content)
+        for section_index, (section, section_content) in enumerate(raw_content.sections):
+            section = section.lower()
+            parsed_index = parsed_mapping[section_index]
+            parsed: List[str] = [] if parsed_index == -1 else parsed_content.sections[parsed_index][1]
+            if section.startswith('%package'):
+                result.extend(cls._parse_package_tags(section, section_content, parsed))
+            elif section.startswith('%sourcelist') or section.startswith('%patchlist'):
+                result.extend(cls._parse_list_tags(section, section_content, parsed, cast(Tuple[Tag], result)))
```
nforro: You are casting a list to a tuple.

rebasehelper/tags.py
FrNecas: We have to go through the remaining sections to set its parsed counteparts as undefined (`-1`). If I added a break there, then the mapping wouldn't contain all of the raw section, would it?

nforro: Ok.

rebasehelper/tags.py
```diff
@@ -57,6 +57,27 @@ def __getitem__(self, index):
...
+            A dict where the keys are indexes of sections in raw content and values are their
+            counterparts in parsed content. Value is set to -1 if the section is not in
+            parsed SpecContent.
+        """
+        result: Dict[int, int] = {}
+        parsed = 0
+        for index, (section, _) in enumerate(raw_content.sections):
+            if parsed >= len(parsed_content.sections) or parsed_content.sections[parsed][0].lower() != section.lower():
+                result[index] = -1
+                continue
```
nforro: This is redundant.

rebasehelper/tags.py
nforro: Wouldn't it be a bit more readable if the condition was reversed?

rebasehelper/tags.py
```diff
@@ -80,7 +80,52 @@ def _map_sections_to_parsed(cls, raw_content: SpecContent, parsed_content: SpecC
...
-        """Parses all tags from provided SPEC content and determines if they are valid.
+        result = []
+        parsed_mapping = cls._map_sections_to_parsed(raw_content, parsed_content)
+        for section_index, (section, section_content) in enumerate(raw_content.sections):
+            section = section.lower()
+            parsed_index = parsed_mapping[section_index]
+            parsed: List[str] = [] if parsed_index == -1 else parsed_content.sections[parsed_index][1]
+            if section.startswith('%package'):
+                result.extend(cls._parse_package_tags(section, section_content, parsed))
+            elif section.startswith('%sourcelist') or section.startswith('%patchlist'):
```
nforro: Why `startswith()` here?

rebasehelper/tags.py
```diff
@@ -80,7 +80,52 @@ def _map_sections_to_parsed(cls, raw_content: SpecContent, parsed_content: SpecC
...
+                result.extend(cls._parse_list_tags(section, section_content, parsed, cast(Tuple[Tag], tuple(result))))
+        return cast(Tuple[Tag], tuple(result))
+
+    @classmethod
+    def _parse_list_tags(cls, section: str, section_content: List[str], parsed: List[str],
+                         parsed_tags: Tuple[Tag]) -> List[Tag]:
+        """Parses all tags from a %sourcelist or %patchlist section.
+
+        Only parses tags that are valid (AKA are in parsed), nothing more can
```
nforro: I don't think AKA is the proper term here, `(that is - are in parsed)` would look better.

rebasehelper/tags.py
```diff
@@ -80,7 +80,52 @@ def _map_sections_to_parsed(cls, raw_content: SpecContent, parsed_content: SpecC
...
+            if not expanded or not line or is_comment or not [p for p in parsed if p == expanded.rstrip()]:
+                continue
+            result.append(Tag(section, i, tag + str(index), (0, len(line) - 1), True))
+            index += 1
+
+        return result
+
+    @classmethod
+    def _parse_package_tags(cls, section: str, section_content: List[str], parsed: List[str]) -> List[Tag]:
+        """Parses all tags from provided by a %package section and determines if they are valid.
```
nforro: I guess `from provided by` is a typo. I would rather use `in a %package section`.

nforro: And the same above, to stay consistent.
BTW, why is `_parse_package_tags()` below `_parse_list_tags()`? It would feel more natural if their order was opposite.

rebasehelper/tags.py
FrNecas: Currently `%sourcelist` and `%patchlist` don't support any arguments but I think they potentially could in the future so I used `startswith()` just to be safe if that makes sense

nforro: What arguments would that be? But fine, it doesn't hurt anything.

rebasehelper/tags.py
```diff
@@ -22,29 +22,33 @@
...
 
 from rebasehelper.spec_content import SpecContent
 from rebasehelper.helpers.macro_helper import MacroHelper
 
 
 class Tag:
-    def __init__(self, section: str, line: int, name: str, value_span: Tuple[int, int], valid: bool) -> None:
+    def __init__(self, section: str, line: int, name: str, value_span: Tuple[int, int], valid: bool,
+                 section_index: int) -> None:
         self.section: str = section
```
nforro: I know it's backwards incompatible change, but I would prefer `section_name` here for clarity. Also I think `section_index` should be positioned above it, because it's the canonical reference.

rebasehelper/spec_content.py
```diff
@@ -104,6 +104,25 @@ def get_comment_span(cls, line: str, section: str) -> Tuple[int, int]:
...
+
+        """
+        index = 0
+        for header, section in self.sections:
+            if header.lower() == name.lower():
+                if index == n:
+                    return section
+                index += 1
+        return None
+
```
nforro: This is not what I had in mind. SpecContent should behave like an iterable and the index should be absolute index in that iterable. not relative to section name. That's confusing and unnecessarily complicated.

rebasehelper/specfile.py
```diff
@@ -195,19 +195,21 @@ def guess_category():
     # TAG HELPER METHODS #
     ######################
 
-    def tag(self, name: str, section: Optional[str] = None) -> Optional[Tag]:
+    def tag(self, name: str, section: Optional[str] = None, section_index: Optional[int] = None) -> Optional[Tag]:
```
nforro: This is a convenience method returning the first matching tag, so it doesn't make much sense to make it more complex by allowing to specify a particular section. If you want to do that anyway, I think the `section` argument should accept both name (`str`) and index (`int`). But `get_raw_tag_value()` and `set_raw_tag_value()` probably shouldn't use this method at all.

rebasehelper/tags.py
FrNecas: Changing the order definitely makes sense. But do we really want to make a backwards incompatible change? The `section` attribute is used several times. I don't think that the difference is that big to be worth renaming the attribute.

nforro: > The `section` attribute is used several times.

What do you mean by that?

FrNecas: For example: https://github.com/rebase-helper/rebase-helper/blob/master/rebasehelper/specfile.py#L207

There are some more uses of the attribute in `ReplaceOldVersionHook`

nforro: Well of course it is used, otherwise it wouldn't be there. But that doesn't mean it makes sense not to rename it. Your examples need to be rewritten anyway, they wouldn't work with non-unique section names.

rebasehelper/plugins/spec_hooks/replace_old_version.py
```diff
@@ -102,8 +102,8 @@ def run(cls, spec_file: SpecFile, rebase_spec_file: SpecFile, **kwargs: Any):
         subversion_patterns = cls._create_possible_replacements(old_version, new_version, replace_with_macro)
         examined_lines: Dict[str, Set[int]] = collections.defaultdict(set)
         for tag in rebase_spec_file.tags.filter():
-            examined_lines[tag.section].add(tag.line)
-            value = rebase_spec_file.get_raw_tag_value(tag.name, tag.section)
+            examined_lines[tag.section_name].add(tag.line)
```
nforro: This is supposed to use `section_index`, `section_name` is not unique.

rebasehelper/plugins/spec_hooks/replace_old_version.py
```diff
@@ -102,8 +102,8 @@ def run(cls, spec_file: SpecFile, rebase_spec_file: SpecFile, **kwargs: Any):
...
+            value = rebase_spec_file.get_raw_tag_value(tag.name, tag.section_name)
```
nforro: Same here.

rebasehelper/spec_content.py
```diff
@@ -87,6 +87,12 @@ def __str__(self):
                 content.append(line + '\n')
         return ''.join(content)
 
+    def __getitem__(self, index):
+        try:
+            return self.sections[index]
+        except IndexError:
+            raise KeyError
```
nforro: What's the reasoning behind changing `IndexError` to `KeyError`?

rebasehelper/tags.py
```diff
@@ -25,22 +25,25 @@
...
 
 from rebasehelper.spec_content import SpecContent
 from rebasehelper.helpers.macro_helper import MacroHelper
 
 
 class Tag:
-    def __init__(self, section: str, line: int, name: str, value_span: Tuple[int, int], valid: bool) -> None:
-        self.section: str = section
+    def __init__(self, section: str, line: int, name: str, value_span: Tuple[int, int], valid: bool,
+                 section_index: int) -> None:
```
nforro: Constructor arguments don't match the attributes now.

rebasehelper/tags.py
nforro: This should be fixed in this commit, not the following one.

rebasehelper/tags.py
FrNecas: Should I move `section_index` to the first position as well?

rebasehelper/spec_content.py
FrNecas: I thought it was a convention to raise `KeyError` if the used key can't be found. Or at least that's what I have seen most of the time

rebasehelper/tags.py
nforro: Yes, of course. Why should it be last?

rebasehelper/spec_content.py
nforro: What key? It's index. You even named it `index`.

rebasehelper/specfile.py
```diff
@@ -195,23 +195,28 @@ def guess_category():
     # TAG HELPER METHODS #
     ######################
 
-    def tag(self, name: str, section: Optional[str] = None) -> Optional[Tag]:
+    def tag(self, name: str, section: Optional[Union[str, int]] = None) -> Optional[Tag]:
         """Returns the first non-unique tag."""
-        tags = self.tags.filter(section=section, name=name)
+        if isinstance(section, str):
+            tags = self.tags.filter(section=section, name=name)
```
nforro: Again, this should be `section_name`, or `section` should accept `str`/`int`.

rebasehelper/specfile.py
```diff
@@ -195,23 +195,28 @@ def guess_category():
...
+            tags = self.tags.filter(section_index=section, name=name)
         return next(tags, None)
 
-    def get_raw_tag_value(self, tag_name: str, section: Optional[str] = None) -> Optional[str]:
+    def get_raw_tag_value(self, tag_name: str, section: Optional[Union[str, int]] = None) -> Optional[str]:
         tag = self.tag(tag_name, section)
         if not tag:
             return None
-        return self.spec_content.section(tag.section)[tag.line][slice(*tag.value_span)]
+        section = self.spec_content.section(section) if isinstance(section, str) else self.spec_content[section]
```
nforro: This is not so great, perhaps `SpecContent` could do the same and accept both types.

rebasehelper/tags.py
```diff
@@ -85,20 +108,23 @@ def sanitize(name):
...
-        for section, _ in raw_content.sections:
+        for section_index, (section, section_content) in enumerate(raw_content.sections):
             section = section.lower()
             if not section.startswith('%package'):
                 continue
-            parsed = parsed_content.section(section)
-            for index, line in enumerate(raw_content.section(section)):
+            parsed_index = parsed_mapping[section_index]
+            parsed: List[str] = [] if parsed_index == -1 else parsed_content.sections[parsed_index][1]
+            for index, line in enumerate(section_content):
```
nforro: This is more of a nitpick, but there are 3 index variables already, so using just `index` is not very clear. `line_index` would be better. And `parsed_index` should probably be `parsed_section_index`. I don't like long variable names but I think clarity matters more.

rebasehelper/specfile.py
FrNecas: I thought about this one. `SpecContent` could accept string or int and either call `SpecContent.section()` or just return by index. What would then be the point of `SpecContent.section()`? Another thing to note is that `__getitem__` could then return None which would need to be handled when using it.

nforro: `__getitem__` shoudn't return `None`, it shoud raise `KeyError` (or `IndexError`). There is no point in preserving the `section()` method, unless removing it would break the API. But I think we should strive to define a clean and consistent API, even at the cost of occasional incompatibility.

FrNecas: Looking at this now, the line didn't make sense at all. The different behaviour of string and int in `section` is already handled in `SpecFile.tag()`. This line should then just use the obtained tag to determine the correct section. So it seems like there is no need to change `SpecContent`. Btw removing `section()` would break the API. IIRC packit uses the method somewhere.

rebasehelper/tags.py
```diff
@@ -111,32 +137,35 @@ def sanitize(name):
...
                         span = cast(Tuple[int, int], tuple(x + start for x in m.span('value')))
-                        result.append(Tag(section, index, sanitize(m.group('name')), span, valid))
+                        result.append(Tag(section_index, section, index, sanitize(m.group('name')), span, valid))
         return cast(Tuple[Tag], tuple(result))
 
     @classmethod
-    def _filter(cls, tags: Tuple[Tag], section: Optional[str] = None, name: Optional[str] = None,
-                valid: Optional[bool] = True) -> Iterator[Tag]:
+    def _filter(cls, tags: Tuple[Tag], section_name: Optional[str] = None, name: Optional[str] = None,
+                valid: Optional[bool] = True, section_index: int = None) -> Iterator[Tag]:
```
nforro: `section_index` should be `Optional[int]` and it would make more sense if it was the first argument.

rebasehelper/tags.py
```diff
@@ -111,32 +137,35 @@ def sanitize(name):
...
+    def filter(self, section_name: Optional[str] = None, name: Optional[str] = None,
+               valid: Optional[bool] = True, section_index: int = None) -> Iterator[Tag]:
         """Filters tags based on section, name or validity. Defaults to all valid tags in all sections.
 
         Args:
-            section: If specified, includes tags only from this section.
+            section_name: If specified, includes tags only from this section.
             name: If specified, includes tags matching this name. Wildcards are supported.
             valid: If specified, includes tags of this validity.
+            section_index: If specified, includes tags from the repeating section of this index.
```
nforro: This description doesn't make sense to me. I would just use one line with both `section_index` and `section_name`.

FrNecas: I don't think that 2 arguments on one line is a valid syntax. I fixed the description

rebasehelper/tags.py
```diff
@@ -111,32 +137,35 @@ def sanitize(name):
...
-        if section is not None:
-            result = filter(lambda t: t.section == section.lower(), result)  # type: ignore
+        if section_name is not None:
+            result = filter(lambda t: t.section_name == section_name.lower(), result)  # type: ignore
         if name is not None:
             result = filter(lambda t: fnmatch.fnmatchcase(t.name, name.capitalize()), result)  # type: ignore
         if valid is not None:
             result = filter(lambda t: t.valid == valid, result)
+        if section_index is not None:
+            result = filter(lambda t: t.section_index == section_index, result)
```
nforro: This should also be first, what sense does it make to have it at the end?

rebasehelper/tags.py
```diff
@@ -111,32 +137,35 @@ def sanitize(name):
...
-    def filter(self, section: Optional[str] = None, name: Optional[str] = None,
-               valid: Optional[bool] = True) -> Iterator[Tag]:
+    def filter(self, section_index: Optional[int] = None, section_name: Optional[str] = None,
+               name: Optional[str] = None, valid: Optional[bool] = True) -> Iterator[Tag]:
         """Filters tags based on section, name or validity. Defaults to all valid tags in all sections.
 
         Args:
-            section: If specified, includes tags only from this section.
+            section_index: If specified, includes tags only from section of this index.
+            section_name: If specified, includes tags only from this section.
```
nforro: I think the correct description here would be: `If specified, includes tags only from sections of this name.`.

rebasehelper/tags.py
nforro: This comment is still valid.

rebasehelper/tags.py
```diff
@@ -111,32 +137,35 @@ def sanitize(name):
...
-    def filter(self, section: Optional[str] = None, name: Optional[str] = None,
-               valid: Optional[bool] = True) -> Iterator[Tag]:
+    def filter(self, section_index: Optional[int] = None, section_name: Optional[str] = None,
+               name: Optional[str] = None, valid: Optional[bool] = True) -> Iterator[Tag]:
         """Filters tags based on section, name or validity. Defaults to all valid tags in all sections.
 
         Args:
-            section: If specified, includes tags only from this section.
+            section_index: If specified, includes tags only from section of this index.
+            section_name: If specified, includes tags only from this sections of this name.
```
nforro: There is a typo, double `this`.

rebasehelper/tags.py
```diff
@@ -108,37 +125,62 @@ def sanitize(name):
...
+                    result.append(Tag(section_index, section, line_index, sanitize(m.group('name')), span, valid))
+
+        return result
+
+    @classmethod
+    def _parse_list_tags(cls, section: str, section_content: List[str], parsed: List[str],
+                         parsed_tags: Tuple[Tag], section_index: int) -> List[Tag]:
+        """Parses all tags in a %sourcelist or %patchlist section.
+
+        Only parses tags that are valid (that is -  are in parsed), nothing more can
```
nforro: Nitpick - double space.

rebasehelper/spec_content.py
```diff
@@ -88,7 +91,7 @@ def __str__(self):
         return ''.join(content)
 
     def __getitem__(self, index):
-        return self.sections[index]
+        return self.sections[index][1]
```
nforro: Isn't this supposed to be part of previous commit?

rebasehelper/specfile.py
```diff
@@ -462,8 +465,12 @@ def is_comment(line):
                 name = os.path.join(constants.RESULTS_DIR, constants.REBASED_SOURCES_DIR, patch_name)
                 self.set_raw_tag_value(tag.name, name)
                 modified_patches.append(patch_num)
-        for span in sorted(remove_lines, key=lambda s: s[0], reverse=True):
-            del preamble[slice(*span)]
+        for section_index, remove in remove_lines.items():
+            content = self.spec_content[section_index]
+            if content is None:
```
nforro: How can `content` be `None` here?

PR #758, opened 2019-12-17
Handle null release name in CommitHashUpdater

FrNecas:
Example of a problematic release: https://api.github.com/repos/KhronosGroup/glslang/releases/15961606 (caused traceback in batch rebase)


rebasehelper/plugins/spec_hooks/commit_hash_updater.py
```diff
@@ -62,7 +62,7 @@ def _get_commit_hash_from_github(cls, spec_file):
         version = spec_file.header.version
         tag_name = None
         for release in data:
-            if version in release.get('name'):
+            if release.get('name') and version in release.get('name'):
```
nforro: Why not `if version in release.get('name', '')`?

Issue #759, opened 2020-01-06
Incorrect (incomplete) paths when using --bugzilla-id


nforro:
When using `--bugzilla-id`, the rebase is being done in a new subdirectory, but the paths listed in the summary don't point to it:

```
Available logs:
Debug log:
rebase-helper-results/logs/debug.log
Old build logs and (S)RPMs:
rebase-helper-results/old-build
New build logs and (S)RPMs:
rebase-helper-results/new-build

Rebased sources:
rebase-helper-results/rebased-sources
Generated patch:
rebase-helper-results/changes.patch

text report:
rebase-helper-results/report.txt
```

FrNecas:
This doesn't work properly with `--results-dir` option either, I will fix it for both

PR #761, opened 2020-01-13
Refactor output tools

FrNecas:
Resolves: #759


rebasehelper/plugins/output_tools/__init__.py
```diff
@@ -51,7 +51,12 @@ def get_report_path(cls, app):
         return os.path.join(app.results_dir, REPORT + '.' + cls.EXTENSION)
 
     @classmethod
-    def prepend_results_dir_name(cls, *path_members):
+    def prepend_results_dir_name(cls, app, *path_members):
+        if app.conf.results_dir:
+            return os.path.join(app.conf.results_dir, RESULTS_DIR, *path_members)
+        elif app.conf.bugzilla_id:
```
nforro: Are the two options mutually exclusive?

rebasehelper/plugins/output_tools/__init__.py
```diff
@@ -51,7 +51,12 @@ def get_report_path(cls, app):
...
+            subdir = os.getcwd().split('/')[-1]
```
nforro: You should use `os.path.split`. But I don't like this special-casing too much. Knowing current working directory and the directory from where `rebase-helper` was started it should be possible to determine the proper path.

rebasehelper/plugins/output_tools/__init__.py
FrNecas: No, both can be used at the same. This should work fine, `--results-dir` has higher priority if we look at where the results directory will be placed. But you brought up a good point. When using both of the options and passing `--results-dir` as a relative path, it's not going to work as expected.

I think the path should be converted to absolute before change directory here:
https://github.com/rebase-helper/rebase-helper/blob/master/rebasehelper/cli.py#L158

rebasehelper/plugins/output_tools/__init__.py
FrNecas: It would be enough but can we somehow determine where `rebase-helper` started? I assume it would be necessary to pass it as an attribute of `Application`.

rebasehelper/plugins/output_tools/__init__.py
nforro: Yes, and the original directory should be saved so it can be used to determine the paths.

rebasehelper/plugins/output_tools/__init__.py
```diff
@@ -51,7 +51,17 @@ def get_report_path(cls, app):
...
-    def prepend_results_dir_name(cls, *path_members):
+    def prepend_results_dir_name(cls, app, *path_members):
+        if app.conf.results_dir and app.conf.bugzilla_id:
+            # revert back from path relative to execution dir to path
+            # relative to starting dir
+            prepend = os.path.relpath(os.path.join(app.execution_dir, app.conf.results_dir), app.start_dir)
+            return os.path.join(prepend, RESULTS_DIR, *path_members)
+        if app.conf.results_dir:
+            return os.path.join(app.conf.results_dir, RESULTS_DIR, *path_members)
+        if app.conf.bugzilla_id:
```
nforro: This special case is not needed anymore.

rebasehelper/plugins/output_tools/__init__.py
```diff
@@ -51,7 +51,17 @@ def get_report_path(cls, app):
         return os.path.join(app.results_dir, REPORT + '.' + cls.EXTENSION)
 
     @classmethod
-    def prepend_results_dir_name(cls, *path_members):
+    def prepend_results_dir_name(cls, app, *path_members):
+        if app.conf.results_dir and app.conf.bugzilla_id:
```
nforro: Same as below, this shouldn't be special-cased.

FrNecas: How come? I feel like there are 4 cases with different behaviour that can occur.

1. just `--bugzilla-id` is used, we should prepend the path with `created_repo/rebase-helper-results`
2. just `--results-dir` is used, we should prepend the path with `config.results_dir/rebase-helper-results`
3. both of the above are used, we should use what was passed in `--results-dir`. The problem is when `--results-dir` is a relative path. We could obviously convert all of the paths to absolute but I think if the user passed a relative path, we should keep it as relative, if possible (hence the special case)
4. every other case, results directory is in the current working directory.

nforro: The point of `start_dir` was to avoid special casing of `--bugzilla-id` (and other potential future options that would change working directory). There are only two cases, `--results-dir` and no `--results-dir`.

rebasehelper/plugins/output_tools/__init__.py
```diff
@@ -51,8 +51,13 @@ def get_report_path(cls, app):
         return os.path.join(app.results_dir, REPORT + '.' + cls.EXTENSION)
 
     @classmethod
-    def prepend_results_dir_name(cls, *path_members):
-        return os.path.join(RESULTS_DIR, *path_members)
+    def prepend_results_dir_name(cls, app, *path_members):
+        if app.conf.results_dir:
+            # revert back from path relative to execution dir to path
+            # relative to starting dir
+            prepend = os.path.relpath(os.path.join(app.execution_dir, app.conf.results_dir), app.start_dir)
```
nforro: What if `results_dir` is absolute? Also, I don't like the expression `revert back`, even though it is technically correct. I think `make the path relative to start_dir` would be enough. You can add `(as opposed to being relative to execution_dir)` if you think it's necessary to mention that.

rebasehelper/plugins/output_tools/json_.py
```diff
@@ -34,14 +34,14 @@ class JSON(BaseOutputTool):
...
     @classmethod
-    def print_summary(cls, path, results):
+    def print_summary(cls, report_path, results):
         """
         Print JSON summary
 
-        :param path: to the report file
-        :param results: dictionary containing info about rebase
+        :param report_path path to the report
+        :param results ResultsStore instance
```
nforro: If you want to keep this style, you should add the missing semicolons (as this is a [field list](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#field-lists)).

rebasehelper/plugins/output_tools/__init__.py
FrNecas: If `results_dir` is absolute, it correctly makes the paths relative to the current directory. The output is correct and makes sense but perhaps it would be better to use absolute paths if the user used absolute path...

nforro: But you are prepending absolute `results_dir` with `execution_dir`. Maybe it really does work, I didn't try it, but it definitely looks wrong. Anyway, yes, if `results_dir` was specified as absolute, it should be absolute in the output as well.

rebasehelper/plugins/output_tools/json_.py
FrNecas: Hm, I will rewrite the documentation in output tools, there are not that many docstrings and we should slowly work towards replacing all of these old-style docstrings with napoleon.

Issue #760, opened 2020-01-08
Avoid false-positive reports of ABI changes


nforro:
For example when rebasing `libjpeg-turbo` from **2.0.3** to **2.0.4**, `rebase-helper` reports ABI changes in `libjpeg-turbo-utils`, but `abipkgdiff` output contains this:

```
==== Error happened during processing of cjpeg: ====
could not find alternate debug info
==== End of error for cjpeg ====
==== Error happened during processing of djpeg: ====
could not find alternate debug info
==== End of error for djpeg ====
==== Error happened during processing of jpegtran: ====
could not find alternate debug info
==== End of error for jpegtran ====
==== Error happened during processing of wrjpgcom: ====
could not find alternate debug info
==== End of error for wrjpgcom ====
==== Error happened during processing of rdjpgcom: ====
could not find alternate debug info
==== End of error for rdjpgcom ====
```

Errors in the output should be definitely ignored, but perhaps they could be prevented.

PR #763, opened 2020-01-14
Abipkgdiff changes

FrNecas:
Resolves: #760

### DIFF
--- a/rebasehelper/plugins/checkers/abipkgdiff.py
+++ b/rebasehelper/plugins/checkers/abipkgdiff.py
@@ -39,11 +39,7 @@ logger: CustomLogger = cast(CustomLogger, logging.getLogger(__name__))
 
 
 class AbiPkgDiff(BaseChecker):
-    """abipkgdiff compare tool
-
-    Attributes:
-        abi_changes(bool): True if ABI changes were detected.
-    """
+    """abipkgdiff compare tool."""
 
     DEFAULT: bool = True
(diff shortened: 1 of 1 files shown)


rebasehelper/plugins/checkers/abipkgdiff.py
```diff
@@ -237,6 +239,8 @@ def format(cls, data):
 
     @classmethod
     def get_important_changes(cls, checker_output):
+        if any(x & cls.ABIDIFF_ABI_INCOMPATIBLE_CHANGE for x in checker_output['ret_codes'].values()):
+            return ['Incompatible ABI changes occurred. Check abipkgdiff output.']
```
nforro: Wouldn't it be better if "normal" ABI changes were detected the same way?

rebasehelper/plugins/checkers/abipkgdiff.py
```diff
@@ -131,15 +126,13 @@ def run_check(cls, results_dir, **kwargs):
                 raise RebaseHelperError('Execution of {} failed.\nCommand line is: {}'.format(cls.CMD, cmd))
             reports[old_name] = int(ret_code)
         return dict(packages=cls.parse_abi_logs(reports),
-                    abi_changes=cls.abi_changes,
-                    path=cls.get_checker_output_dir_short())
+                    path=cls.get_checker_output_dir_short(),
+                    ret_codes=reports)
```
nforro: Perhaps the variable shouldn't be called `reports`?

PR #765, opened 2020-01-20
Update %patch indexes when using %patchlist

FrNecas:
Resolves: #622 (from what I can see, lists should now be fully functional and looking at the changelog there aren't any remaining features that we haven't implemented)

### DIFF
--- a/rebasehelper/specfile.py
+++ b/rebasehelper/specfile.py
@@ -376,28 +376,31 @@ class SpecFile:
         if not prep:
             return
 
-        patch_re = re.compile(r'^%patch(?P<index>\d+).*')
+        patch_re = re.compile(r'^%patch(?P<index>\d+)(.*)')
 
         i = 0
+        removed = 0
         while i < len(prep):
             line = prep[i]
             match = patch_re.match(line)
             if match:
                 index = int(match.group('index'))
-                for num in reversed(comment_out):
-                    if num == index:
-                        if disable_inapplicable_patches:
-                            prep[i] = '#%{}'.format(line)
-                        prep.insert(i, '# The following patch contains conflicts')
-                        comment_out.remove(num)
-                        i += 1
-                        break
-                for num in reversed(remove_patches):
-                    if num == index:
-                        del prep[i]
-                        remove_patches.remove(num)
-                        i -= 1
-                        break
+                if index in comment_out:
+                    if disable_inapplicable_patches:
+                        prep[i] = '#%{}'.format(line)
+                        removed += 1
+                    prep.insert(i, '# The following patch contains conflicts')
+                    comment_out.remove(index)
+                    i += 1
+                elif index in remove_patches:
+                    del prep[i]
+                    remove_patches.remove(index)
+                    i -= 1
+                    removed += 1
+                # When combining Patch tags and %patchlist, if a Patch is removed, the indexes
+                # of %patchlist patches change and %patch macros need to be modified.
+                elif self.tag('Patch{}'.format(index)).section_name.startswith('%patchlist'):
+                    prep[i] = patch_re.sub(r'%patch{}\2'.format(index - removed), prep[i])
             i += 1
 
     @saves
--- a/rebasehelper/tests/conftest.py
+++ b/rebasehelper/tests/conftest.py
@@ -28,6 +28,8 @@ import shutil
 import pytest  # type: ignore
 
 from rebasehelper.specfile import SpecFile
+from rebasehelper.spec_content import SpecContent
+from rebasehelper.tags import Tags
 
 
 TESTS_DIR: str = os.path.dirname(__file__)
@@ -59,3 +61,16 @@ def pytest_collection_modifyitems(items):
             item.add_marker(pytest.mark.functional)
         else:
             item.add_marker(pytest.mark.standard)
+
+
+@pytest.fixture
+def mocked_spec_object(spec_attributes):
+    spec = SpecFile.__new__(SpecFile)
+    spec.save = lambda: None
+    for attribute, value in spec_attributes.items():
+        if attribute == 'spec_content' and isinstance(value, str):
+            value = SpecContent(value)
+        setattr(spec, attribute, value)
+    if hasattr(spec, 'spec_content') and not hasattr(spec, 'tags'):
+        spec.tags = Tags(spec.spec_content, spec.spec_content)
+    return spec
--- a/rebasehelper/tests/test_specfile.py
+++ b/rebasehelper/tests/test_specfile.py
@@ -24,6 +24,7 @@
 
 import os
 from typing import List
+from textwrap import dedent
 
 import pytest  # type: ignore
 
@@ -198,6 +199,95 @@ class TestSpecFile:
         assert target == 'test-1.0.2'
         target = spec_object.find_archive_target_in_prep('misc.zip')
         assert target == 'test-1.0.2/misc'
+
+    @pytest.mark.parametrize('spec_attributes, kwargs, expected_content', [
+        (
+            {
+                'removed_patches': [],
+                'spec_content': dedent("""\
+                    Patch0:    0.patch
+                    Patch1:    1.patch
+
+                    %patchlist
+                    2.patch
+                    3.patch
+
+                    %prep
+                    %patch0 -p0
+                    %patch1 -p1
+                    %patch2 -p2
+                    %patch3 -p3
+                    """),
+            },
+            {
+                'patches':
+                    {
+                        'deleted': ['2.patch'],
+                        'inapplicable': ['1.patch'],
+                    },
+                'disable_inapplicable': False,
+            },
+            dedent("""\
+                Patch0:    0.patch
+                Patch1:    1.patch
+
+                %patchlist
+                3.patch
+
+                %prep
+                %patch0 -p0
+                # The following patch contains conflicts
+                %patch1 -p1
+                %patch2 -p3
+                """)
+        ),
+        (
+            {
+                'removed_patches': [],
+                'spec_content': dedent("""\
+                    Patch0:    0.patch
+                    Patch1:    1.patch
+
+                    %patchlist
+                    2.patch
+                    3.patch
+
+                    %prep
+                    %patch0 -p0
+                    %patch1 -p1
+                    %patch2 -p2
+                    %patch3 -p3
+                    """),
+            },
+            {
+                'patches':
+                    {
+                        'deleted': ['2.patch'],
+                        'inapplicable': ['1.patch'],
+                    },
+                'disable_inapplicable': True,
+            },
+            dedent("""\
+        Patch0:    0.patch
+        #Patch1:    1.patch
+
+        %patchlist
+        3.patch
+
+        %prep
+        %patch0 -p0
+        # The following patch contains conflicts
+        #%%patch1 -p1
+        %patch1 -p3
+        """),
+        ),
+    ], ids=[
+        'do_not_disable_inapplicable',
+        'disable_inapplicable',
+    ])
+    def test_write_updated_patches(self, mocked_spec_object, kwargs, expected_content):
+        mocked_spec_object.write_updated_patches(**kwargs)
+        assert expected_content == str(mocked_spec_object.spec_content)
 
     def test_update_paths_to_patches(self, spec_object):
         """


rebasehelper/specfile.py
```diff
@@ -391,13 +392,19 @@ def _process_patches(self, comment_out=None, remove_patches=None, disable_inappl
...
                     if num == index:
                         del prep[i]
                         remove_patches.remove(num)
                         i -= 1
+                        removed += 1
                         break
+                # When combining Patch tags and %patchlist, if a Patch is removed, the indexes
+                # of %patchlist patches change and %patch macros need to be modified.
+                if self.tag('Patch{}'.format(index)).section_name.startswith('%patchlist'):
+                    prep[i] = patch_re.sub(r'%patch{}\2'.format(index - removed), prep[i])
```
nforro: Are you sure this works as expected? I'd like to see a test that proves it.

FrNecas: I've tried it on a simple case (some Patch tags and a patchlist and a patch defined by Patch tag was removed) and it worked correctly. I couldn't think of a more complicated scenario.

As for testing, I'd really like to write some tests, the problem is that we are running tests even for the older version of rpm. And since we need access to tags, the test spec has to be parseable by rpm... I may try to solve it by mocking.. Or perhaps we could create a separate spec file just for this test and run it only on rpm version >= 4.15?

nforro: > some Patch tags and a patchlist and a patch defined by Patch tag was removed

And what if a tag defined by %patchlist is removed?

> Or perhaps we could create a separate spec file just for this test and run it only on rpm version >= 4.15?

No, mocking is the correct way here.

rebasehelper/tests/conftest.py
```diff
@@ -59,3 +61,15 @@ def pytest_collection_modifyitems(items):
             item.add_marker(pytest.mark.functional)
         else:
             item.add_marker(pytest.mark.standard)
+
+
+@pytest.fixture
+def mock_spec_object_generator(monkeypatch):
+    def _mock_spec(content):
```
FrNecas: Looking at this, perhaps I could make the fixture a bit more general. There could be an argument for passing methods to monkeypatch and an argument for passing attributes to set. This would make the fixture usable for a lot of test cases. What do you think @nforro ?

nforro: Sure, but maybe we should consider using something like [flexmock](https://flexmock.readthedocs.io/en/latest/).

rebasehelper/tests/test_specfile.py
```diff
@@ -199,6 +199,34 @@ def test_find_archive_target_in_prep(self, spec_object):
...
+%patchlist
+2.patch
+
+%prep
+%patch0 -p0
+%patch1 -p0
+"""
+        spec = mock_spec_object_generator(content)
+        spec.removed_patches = []
+        spec.write_updated_patches({'deleted': ['1.patch', '3.patch']}, False)
```
nforro: This test ignores inapplicable patches, but that's exactly the part I suspect to be a bit buggy.

FrNecas: Updated the test to also disable an inapplicable patch. Seems to work correctly

rebasehelper/specfile.py
```diff
@@ -397,7 +397,6 @@ def _process_patches(self, comment_out=None, remove_patches=None, disable_inappl
                     remove_patches.remove(index)
                     i -= 1
                     removed += 1
-                    break
```
nforro: This is not supposed to be here.

rebasehelper/tests/test_specfile.py
```diff
@@ -199,6 +199,36 @@ def test_find_archive_target_in_prep(self, spec_object):
...
+#3.patch
+
+%prep
+%patch0 -p0
+%patch1 -p0
+# The following patch contains conflicts
+#%%patch3 -p0
+"""
+        spec = mock_spec_object_generator({'save': lambda _: None}, {'spec_content': content, 'removed_patches': []})
+        spec.write_updated_patches({'deleted': ['1.patch'], 'inapplicable': ['3.patch']}, True)
```
nforro: And now swap deleted with inapplicable and watch the test fail.

FrNecas: It seems to me that even after swapping, the code for modifying prep works correctly. Here is the spec content after swapping inapplicable and deleted.

```
Patch0:    0.patch
#Patch1:   1.patch

%patchlist
2.patch

%prep
%patch0 -p0
# The following patch contains conflicts
#%%patch1 -p1
%patch1 -p0
```

So 3.patch was removed from patchlist, it was removed from prep. 1.patch was commented out and so the index of 2.patch changed to 1 and it is reflected in prep. Am I missing something?

nforro: Sorry, my bad. But what if you set `disable_inapplicable` to `False`?

rebasehelper/tests/conftest.py
```diff
@@ -59,3 +62,20 @@ def pytest_collection_modifyitems(items):
             item.add_marker(pytest.mark.functional)
         else:
             item.add_marker(pytest.mark.standard)
+
+
+@pytest.fixture
+def mock_spec_object_generator(monkeypatch):
```
nforro: I don't like the name. The fixture doesn't return a generator. Also, I'd rather see `mocked` instead of `mock`.
Also, wouldn't it be more convenient to not to return a "decorator", but `SpecFile` instance with attributes set up via traditional method arguments (that can be directly parametrized)?

rebasehelper/tests/conftest.py
```diff
@@ -59,3 +62,20 @@ def pytest_collection_modifyitems(items):
...
+    def _mock_spec(override_methods: Dict[str, Callable], set_attributes: Dict[str, Any]):
```
nforro: What's the difference between method and attribute? Why not just use a single argument?

rebasehelper/tests/conftest.py
FrNecas: If we want to keep this as a pytest fixture, I don't think that there is a way to directly return SpecFile while using arguments to influence the `SpecFile`. It could be changed to a normal function.. Wouldn't it have any downsides when using it in our tests?

rebasehelper/tests/test_specfile.py
FrNecas: Good catch, there was indeed a bug. I think some more tests for this method should be added to test various possibilities (at least test both states of `disable_inapplicable`). But I'd ideally want to avoid having long strings in the tests, I think they decrease readability quite a lot (those 2 10-line strings are not ideal). Do you have any ideas how we could do it in a better way?

rebasehelper/tests/conftest.py
nforro: What makes you think so? [This fixture](https://github.com/rebase-helper/rebase-helper/blob/master/rebasehelper/tests/test_config.py#L38) does something like that.

rebasehelper/tests/test_specfile.py
nforro: Other than dividing the tests into more individual files? No, I don't think there is a way.

rebasehelper/tests/conftest.py
```diff
@@ -59,3 +61,15 @@ def pytest_collection_modifyitems(items):
...
+
+@pytest.fixture
+def mocked_spec_object(monkeypatch, spec_attributes):
+    spec = SpecFile.__new__(SpecFile)
+    for attribute, value in spec_attributes.items():
+        monkeypatch.setattr(spec, attribute, value, raising=False)
+    if 'spec_content' in spec_attributes and isinstance(spec_attributes['spec_content'], str):
+        spec.spec_content = SpecContent(spec_attributes['spec_content'])
+    if hasattr(spec, 'spec_content') and not hasattr(spec, 'tags'):
+        spec.tags = Tags(spec.spec_content, spec.spec_content)
```
nforro: More pythonic would be to just try to assign to `spec.tags` and handle `AttributeError`.

rebasehelper/tests/test_specfile.py
```diff
@@ -199,6 +200,97 @@ def test_find_archive_target_in_prep(self, spec_object):
         target = spec_object.find_archive_target_in_prep('misc.zip')
         assert target == 'test-1.0.2/misc'
 
+    @pytest.mark.parametrize('spec_attributes, kwargs, expected_content', [
+        (
+            {
+                'save': lambda: None,
+                'removed_patches': [],
+                'spec_content': dedent("""
```
nforro: I think there should be a backslash at the end of this line. An empty line perhaps doesn't matter here, but it could matter in some other test, and this can't serve as a template in such case.

rebasehelper/tests/test_specfile.py
```diff
@@ -199,6 +200,97 @@ def test_find_archive_target_in_prep(self, spec_object):
...
+        # The following patch contains conflicts
+        #%%patch1 -p1
+        %patch1 -p1
+        """),
+        ),
+    ], ids=[
+        'do_not_disable_inapplicable',
+        'disable_inapplicable',
+    ])
+    def test_write_updated_patches(self, kwargs, expected_content, mocked_spec_object):
```
nforro: Just a nitpick, but I would prefer `mocked_spec_object` to be the first argument, since it is being used/accessed first.

nforro: Also, it would match the order of parametrized arguments (`spec_attributes` directly influences `mocked_spec_object`).

rebasehelper/tests/test_specfile.py
```diff
@@ -199,6 +200,97 @@ def test_find_archive_target_in_prep(self, spec_object):
...
+
+                    %patchlist
+                    2.patch
+                    3.patch
+
+                    %prep
+                    %patch0 -p0
+                    %patch1 -p1
+                    %patch2 -p0
+                    %patch3 -p1
```
nforro: I see the point in these `-p` options, but wouldn't it be even more obvious if it was like this?
```
%patch0 -p0
%patch1 -p1
%patch2 -p2
%patch3 -p3
```

rebasehelper/tests/conftest.py
```diff
@@ -59,3 +61,18 @@ def pytest_collection_modifyitems(items):
             item.add_marker(pytest.mark.functional)
         else:
             item.add_marker(pytest.mark.standard)
+
+
+@pytest.fixture
+def mocked_spec_object(monkeypatch, spec_attributes):
+    spec_attributes['save'] = lambda: None
```
nforro: How can one redefine the `save` method now? This is not setting default value, this is overriding any passed one.

rebasehelper/tests/conftest.py
```diff
@@ -59,3 +61,18 @@ def pytest_collection_modifyitems(items):
...
+    spec = SpecFile.__new__(SpecFile)
+    for attribute, value in spec_attributes.items():
+        monkeypatch.setattr(spec, attribute, value, raising=False)
```
nforro: What's the point of setting the `spec_content` attribute twice? Wouldn't it be better to convert `value` first if `attribute` is `'spec_content'`, and only then call `monkeypatch.setattr()`?

rebasehelper/tests/conftest.py
```diff
@@ -59,3 +61,18 @@ def pytest_collection_modifyitems(items):
...
+        if attribute == 'spec_content' and isinstance(value, str):
+            monkeypatch.setattr(spec, attribute, SpecContent(value), raising=False)
+    try:
+        spec.tags = Tags(spec.spec_content, spec.spec_content)
+    except AttributeError:
+        pass
```
nforro: Actually, `tags` could have been passed in, so `hasattr()` can't be avoided it seems. Sorry.

rebasehelper/tests/conftest.py
```diff
@@ -59,3 +61,17 @@ def pytest_collection_modifyitems(items):
             item.add_marker(pytest.mark.functional)
         else:
             item.add_marker(pytest.mark.standard)
+
+
+@pytest.fixture
+def mocked_spec_object(monkeypatch, spec_attributes):
+    if 'save' not in spec_attributes:
+        spec_attributes['save'] = lambda: None
```
nforro: Ok, but why not simply do `spec.save = lambda: None` after creating `spec`?

nforro: But of course using `monkeypatch.setattr()`.

nforro: Actually, is `monkeypatch.setattr()` even necessary? Looking at the implementation, it doesn't to anything special except remembering the old value, which we don't need here.

FrNecas: Okay, lets just use `setattr()`, that should suffice in this case.

rebasehelper/tests/conftest.py
```diff
@@ -59,3 +63,26 @@ def pytest_collection_modifyitems(items):
...
+            value = types.MethodType(value, instance)
+    except TypeError:
+        pass
+    setattr(instance, name, value)
+
+
+@pytest.fixture
+def mocked_spec_object(spec_attributes):
+    spec = SpecFile.__new__(SpecFile)
+    spec.save = lambda: None
```
nforro: I think you need to use `setattr_` with `lambda self: None` here, otherwise it will create a static method.

PR #768, opened 2020-01-28
Add support for automatic patch/source numbering

FrNecas:


nforro:
This is overcomplicated. The last source/patch indexes should be kept as local variables in `Tags._parse()`, to avoid filtering and parsing interim results. `Tag` probably should gain `index` attribute, only used with indexed tags, to allow filtering on it and to avoid parsing it from its name every time. The `sanitize()` method should be renamed and it shoud return both sanitized name and index.

rebasehelper/tags.py
```diff
@@ -82,23 +84,43 @@ def _map_sections_to_parsed(cls, raw_content: SpecContent, parsed_content: SpecC
 
     @classmethod
     def _parse(cls, raw_content: SpecContent, parsed_content: SpecContent) -> Tuple[Tag]:
-        result = []
+        result: List[Tag] = []
         parsed_mapping = cls._map_sections_to_parsed(raw_content, parsed_content)
+        patch_source_indices = {tag: 0 for tag in ('source', 'patch')}
```
nforro: I don't like using dict for this. I'd rather see two (function?) attributes, called `next_source_index` and `next_patch_index` for example. Not to mention `indices` is inconsistent with the rest of this module.

rebasehelper/tags.py
```diff
@@ -82,23 +84,43 @@ def _map_sections_to_parsed(cls, raw_content: SpecContent, parsed_content: SpecC
...
 
+    @classmethod
+    def _sanitize_tag(cls, name: str, indices_dict: Dict[str, int]) -> Tuple[str, Optional[int]]:
+        name = name.lower()
+        index = None
+        if name in ('source', 'patch'):
+            # automatic numbering
+            index = indices_dict[name]
+            indices_dict[name] += 1
+            name = name + str(index)
```
nforro: Is it really necessary to special-case this? Either you have explicit index or you don't, that's the only difference between these two branches.

rebasehelper/tags.py
```diff
@@ -82,23 +84,43 @@ def _map_sections_to_parsed(cls, raw_content: SpecContent, parsed_content: SpecC
...
+                                                      section_index))
             elif section.startswith('%sourcelist') or section.startswith('%patchlist'):
-                result.extend(cls._parse_list_tags(section, section_content, parsed,
-                                                   cast(Tuple[Tag], tuple(result)), section_index))
+                result.extend(cls._parse_list_tags(section, section_content, parsed ,patch_source_indices,
+                                                   section_index))
         return cast(Tuple[Tag], tuple(result))
 
+    @classmethod
+    def _sanitize_tag(cls, name: str, indices_dict: Dict[str, int]) -> Tuple[str, Optional[int]]:
```
nforro: I don't particularly like this name, it's not descriptive enough, but unfortunately I can't think of anything better.

rebasehelper/tags.py
FrNecas: By using dict we can easily modify the indices inside the methods (for example in `_sanitize`) and the changes "are propagated" to the calling method (`Tags._parse`). That wouldn't be the case with just integers, would it? Perhaps the 2 attributes you mentioned could be attributes of `Tags` but that doesn't seem right to me

nforro: Yes, I know, that's what I don't like about it. Why not? Also, I suggested function attributes, because their scope would be exactly just `Tags._parse()`.

rebasehelper/tags.py
```diff
@@ -82,23 +84,49 @@ def _map_sections_to_parsed(cls, raw_content: SpecContent, parsed_content: SpecC
 
     @classmethod
     def _parse(cls, raw_content: SpecContent, parsed_content: SpecContent) -> Tuple[Tag]:
-        result = []
+        result: List[Tag] = []
         parsed_mapping = cls._map_sections_to_parsed(raw_content, parsed_content)
+        patch_index = 0
+        source_index = 0
```
nforro: I still think `next_` prefix should be here.

rebasehelper/tags.py
```diff
@@ -82,23 +84,49 @@ def _map_sections_to_parsed(cls, raw_content: SpecContent, parsed_content: SpecC
...
         # counts number of occurrences of one particular section
         for section_index, (section, section_content) in enumerate(raw_content.sections):
             section = section.lower()
             parsed_section_index = parsed_mapping[section_index]
             parsed: List[str] = [] if parsed_section_index == -1 else parsed_content.sections[parsed_section_index][1]
             if section.startswith('%package'):
-                result.extend(cls._parse_package_tags(section, section_content, parsed, section_index))
+                new_tags, patch_index, source_index = cls._parse_package_tags(section, section_content, parsed,
```
nforro: Nitpick, but what's so new about these tags? Why not just `tags`?

rebasehelper/tags.py
nforro: Also, why this order? Sources are usually before patches in a spec file, aren't they?

rebasehelper/tags.py
```diff
@@ -82,23 +84,49 @@ def _map_sections_to_parsed(cls, raw_content: SpecContent, parsed_content: SpecC
...
+        tokens = re.split(r'(\d+)', name, 1)
+        if name.startswith('source'):
+            index = source_index if len(tokens) == 1 else int(tokens[1])
+            source_index = index + 1
+            name = '{}{}'.format(tokens[0], index)
+        elif name.startswith('patch'):
+            index = patch_index if len(tokens) == 1 else int(tokens[1])
+            patch_index = index + 1
+            name = '{}{}'.format(tokens[0], index)
+        return name.capitalize(), patch_index, source_index, index
```
nforro: I think this would be a bit better:

```python
name, *rest = re.split(r'(\d+)', name, 1)
sanitized_name = name.capitalize()
next_index = next_source_index if name == 'source' else next_patch_index
index = int(next(iter(rest), next_index))
if name == 'source':
    return sanitized_name, index, index + 1, patch_index
if name == 'patch':
    return sanitized_name, index, source_index, index + 1
return sanitized_name, None, source_index, patch_index
```

PR #769, opened 2020-01-28
Reset macros when SpecFile is destroyed

FrNecas:


rebasehelper/specfile.py
```diff
@@ -113,7 +113,7 @@ def wrapper(spec, *args, **kwargs):
 
 class SpecFile:
 
-    """Class representing a SPEC file"""
+    """Class representing a SPEC file. **Using SpecFile wipes all macros.**"""
```
nforro: I'm afraid this is not formulated too well. How about: `Be aware that SpecFile modifies RPM macros in global context.`?

rebasehelper/specfile.py
```diff
@@ -133,6 +133,9 @@ def __init__(self, path: str, sources_location: str = '', predefined_macros: Opt
         # Load rpm information
         self._update_data()
 
+    def __del__(self):
+        rpm.reloadConfig()
```
nforro: I'd add a comment, something like `# make sure there are no leftover macros`.

rebasehelper/specfile.py
FrNecas: Okay, I will change it :) I've just noticed that the tests failed in Travis but the fails seem rather weird to me... It passes on F30, works correctly locally.

PR #774, opened 2020-02-07
Correctly handle nested macros in set_tag

FrNecas:


rebasehelper/specfile.py
```diff
@@ -608,16 +608,19 @@ def _find_macros(s):
...
-                rep = _expand_macros(value)
+                macros.add(macro)
+                rep, new_macros = _expand_macros(value)
+                macros |= new_macros
                 if _find_macros(rep):
                     replace.append((rep, span))
             for rep, span in reversed(replace):
                 s = s[:span[0]] + rep + s[span[1]:]
-            return s
+            return s, macros
```
nforro: I think some extra documentation is needed here. With `_expand_macros()` returning a `str`, it's obvious what it is and what it does, but now it returns two values and that doesn't make much sense. At least the docstring should be extended.

rebasehelper/specfile.py
```diff
@@ -606,18 +606,23 @@ def _find_macros(s):
             return result
 
         def _expand_macros(s):
-            """Expands all redefinable macros containing redefinable macros"""
+            """Expands all redefinable macros containing redefinable macros while keeping
+            track of all expanded macros. Returns the expanded string and a set of expanded
+            macro names."""
```
nforro: The closing `"""` should be on its own line.

PR #775, opened 2020-02-14
Tests refactoring

FrNecas:


rebasehelper/tests/test_specfile.py
```diff
@@ -79,20 +81,43 @@ def test_set_version(self, spec_object):
...
-    def test__write_spec_content(self, spec_object):
+    @pytest.mark.parametrize('spec_attributes, content', [
+        (
+            {
+                'path': SPEC_FILE,
+            },
+            'testing line1\ntesting line 2\n'
+        )
+    ])
+    def test__write_spec_content(self, mocked_spec_object, content):
```
nforro: This makes me wonder - why do we test this? I mean, what does it actually test? `SpecContent.__str__()` or `file.write()`? I don't think the latter needs to be tested, and the former is already being tested by `TestSpecContent.test_string_representation()`.

rebasehelper/tests/test_specfile.py
```diff
@@ -79,20 +81,43 @@ def test_set_version(self, spec_object):
...
         # pylint: disable=protected-access
-        new_content = 'testing line 1\ntesting line 2\n'
-        spec_object.spec_content = SpecContent(new_content)
-        spec_object._write_spec_content()
-        with open(SPEC_FILE) as spec:
-            assert new_content == spec.read()
+        mocked_spec_object.spec_content = SpecContent(content)
```
nforro: Why is this not a `spec_content` attribute inside `spec_attributes`?

rebasehelper/tests/test_specfile.py
```diff
@@ -115,11 +140,6 @@ def test_get_patches(self, workdir, spec_object):
             patches[index] = [p.path, p.index]
         assert patches == expected_patches
 
-    def test_get_requires(self, spec_object):
-        expected = {'openssl-devel', 'pkgconfig', 'texinfo', 'gettext', 'autoconf'}
-        req = spec_object.header.requires
-        assert len(expected.intersection(req)) == len(expected)
-
```
nforro: I agree that this test makes no sense, however, there is nothing about removal of meaningless/redundant tests in the commit message ;-)

rebasehelper/tests/test_specfile.py
FrNecas: I think that the original intention was to check whether replacing the whole `SpecContent` has an effect on writing the spec. You are right that there is probably no point in testing a method which just writes a string representation of `SpecContent` (which is already covered by tests). I will remove the test.

rebasehelper/tests/test_specfile.py
```diff
@@ -177,23 +243,50 @@ def test_set_extra_version(self, spec_object):
...
+    def test_update_setup_dirname(self, mocked_spec_object):
+        # mock %{version} and %{name} macros, set their level to -3
+        macros = {'name': 'test', 'version': '1.0.2'}
+        for macro, value in macros.items():
+            rpm.addMacro(macro, value)
+        mocked_spec_object.macros = MacroHelper.dump()
+        for m in macros:
+            for macro in mocked_spec_object.macros:
+                if macro['name'] == m:
+                    macro['level'] = -3
```
nforro: This definitely shouldn't be part of a test. `mocked_spec_object` fixture should instead accept another argument like `macros`.

rebasehelper/tests/test_specfile.py
```diff
@@ -177,23 +243,50 @@ def test_set_extra_version(self, spec_object):
...
         assert '-n %{name}-%{version}rc1' in setup
 
-        spec_object.update_setup_dirname('test-1.0.2-rc1')
-        prep = spec_object.spec_content.section('%prep')
+        mocked_spec_object.update_setup_dirname('test-1.0.2-rc1')
+        prep = mocked_spec_object.spec_content.section('%prep')
         setup = [l for l in prep if l.startswith('%setup')][0]
         assert '-n %{name}-%{version}-rc1' in setup
 
+        rpm.reloadConfig()
```
nforro: Same as above. This should be the *teardown* part of `mocked_spec_object` fixture.

rebasehelper/tests/test_specfile.py
nforro: Or maybe, to stay consistent, it should process `macros` key in `spec_attributes` analogically to how `spec_content` is being processed.

rebasehelper/tests/plugins/test_spec_hooks.py
```diff
@@ -70,28 +112,59 @@ def test_escape_macros_spec_hook(self, spec_object):
...
+
+                %changelog
+                * Wed Apr 26 2017 Nikola Forró <nforro@redhat.com> - 1.0.2-34
+                - Update to 1.0.2
+                """),
+            'header': SimpleNamespace(version='1.0.2'),
+        }
+    ])
+    def test_replace_old_version_spec_hook(self, mocked_spec_object, replace_with_macro):
+        new_spec = copy.deepcopy(mocked_spec_object)
```
nforro: How about another fixture, called e.g. `mocked_spec_object_copy`, for this?

rebasehelper/tests/plugins/test_spec_hooks.py
```diff
@@ -70,28 +112,59 @@ def test_escape_macros_spec_hook(self, spec_object):
...
         # Check if the version has been updated
-        test_source = [line for line in new_spec.spec_content.section('%package') if line.startswith('Source9')]
+        test_source = new_spec.get_raw_tag_value('Source9')
         assert test_source
         if replace_with_macro:
             expected_result = 'https://test.com/#/1.1/%{name}-hardcoded-version-%{version}.tar.gz'
         else:
             expected_result = 'https://test.com/#/1.1/%{name}-hardcoded-version-1.1.0.tar.gz'
-        assert test_source[0].split()[1] == expected_result
+        assert test_source == expected_result
```
nforro: Now it doesn't make much sense not to assert directly in the condition. The change also makes `assert test_source` redundant.

rebasehelper/tests/plugins/test_spec_hooks.py
```diff
@@ -70,28 +112,59 @@ def test_escape_macros_spec_hook(self, spec_object):
...
+    @pytest.mark.parametrize('spec_attributes', [
+        {
+            'spec_content': dedent("""\
+                URL: https://pypi.python.org/pypi/%{name}
+                Source0: https://pypi.python.org/.../%{name}.%{version}.tar.gz
+                """)
+        }
+    ])
+    def test_pypi_to_python_hosted_url_trans(self, mocked_spec_object):
         # pylint: disable=protected-access
```
nforro: This is no longer necessary, is it?

rebasehelper/tests/conftest.py
```diff
@@ -70,21 +71,26 @@ def mocked_spec_object(spec_attributes):
...
+            for macro, properties in value.items():
+                rpm.addMacro(macro, properties.get('value', ''))
+            macros = MacroHelper.dump()
+            for macro, properties in value.items():
+                for m in macros:
+                    if m['name'] == macro:
+                        for prop, v in properties.items():
+                            if prop != 'value':
+                                m[prop] = v
+            value = macros
```
nforro: Why this move? I mean, why not do it this way already in commit 0e7473d?

rebasehelper/tests/conftest.py
```diff
@@ -70,21 +71,26 @@ def mocked_spec_object(spec_attributes):
...
-                        if prop != 'value':
-                            m[prop] = value
     if hasattr(spec, 'spec_content') and not hasattr(spec, 'tags'):
         spec.tags = Tags(spec.spec_content, spec.spec_content)
     return spec
+
+
+@pytest.fixture
+def mocked_spec_object_copy(mocked_spec_object):  # pylint: disable=redefined-outer-name
+    return copy.deepcopy(mocked_spec_object)
```
nforro: This is a nitpick, but still. Why are these new fixtures below `pytest_collection_modifyitems`  hook and not below `spec_object` fixture?

rebasehelper/tests/plugins/test_spec_hooks.py
```diff
@@ -33,12 +36,38 @@
...
+                /usr/share/man/man1/*
+                /usr/bin/%{name}
+                %config(noreplace) /etc/test/test.conf
+
+                %files devel
+                %{_bindir}/test_example
+                %{_libdir}/my_test.so
+                /usr/share/test1.txt
+                /no/macros/here
+
```
nforro: Why the empty line here?

rebasehelper/tests/plugins/test_spec_hooks.py
```diff
@@ -33,12 +36,38 @@
...
+                """),
+            'macros': {},
```
nforro: Why is this necessary?

rebasehelper/tests/conftest.py
FrNecas: Hmm, I amended to a wrong commit :/ I didn't really like setting the `macros` attribute to spec (which aren't really the macros intended for use) and wanted to modify the attribute correctly inside the `for` loop, to stay consistent with how `SpecContent` is handled.

rebasehelper/tests/plugins/test_spec_hooks.py
FrNecas: The hook requires `SpecFile.macros` to be set. `mocked_spec_content` only sets the macros if `macros` key is a part of `spec_attributes`. By setting it to empy dict, `SpecFile.macros` is set to `MacroHelper.dump()`

rebasehelper/tests/test_specfile.py
```diff
@@ -170,20 +232,49 @@ def test_set_extra_version(self, spec_object):
...
+                        {
+                            'value': 'test',
+                            'level': -3,
+                        },
+                    'version':
+                        {
+                            'value': '1.0.2',
+                            'level': -3,
+                        },
+                },
```
nforro: I think this would actually be more readable if it was squashed into 2 lines.

rebasehelper/tests/conftest.py
```diff
@@ -53,24 +55,36 @@ def spec_object(workdir):  # pylint: disable=redefined-outer-name
...
+                for m in macros:
+                    if m['name'] == macro:
+                        for prop, v in properties.items():
+                            if prop != 'value':
+                                m[prop] = v
+            value = macros
         if attribute == 'spec_content' and isinstance(value, str):
             value = SpecContent(value)
         setattr(spec, attribute, value)
+
```
nforro: Extra newline :-)

PR #785, opened 2020-02-28
Simplify packit config

jpopelka:
https://packit.dev/docs/configuration/

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -1,30 +1,13 @@
 specfile_path: rebase-helper.spec
-synced_files:
-    - rebase-helper.spec
-    - .packit.yml
-upstream_package_name: rebase-helper
-downstream_package_name: rebase-helper
 current_version_command: ["python3", "setup.py", "--version"]
 create_tarball_command: ["python3", "setup.py", "sdist", "--base-name", "rebase-helper", "--dist-dir", "."]
 jobs:
-    - job: copr_build
-      trigger: pull_request
-      metadata:
-          targets:
-              - fedora-rawhide-x86_64
-              - fedora-31-x86_64
-              - fedora-30-x86_64
-    - job: propose_downstream
-      trigger: release
-    - job: propose_downstream
-      trigger: release
-      metadata:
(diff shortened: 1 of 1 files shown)


jpopelka:
strange, `packit srpm` works ok locally

nforro:
Well, `strictness: veryhigh` obviously comes from `.prospector.yml`, if that tells you anything.

nforro:
I'm not sure what you are trying to accomplish, but both upstream and downstream package name needs to be `rebase-helper`, with the dash. And so does stdist's base-name.

jpopelka:
Both [upstream_package_name](https://packit.dev/docs/configuration/#upstream-package-name) and [downstream_package_name](https://packit.dev/docs/configuration/#downstream-package-name) now default (if not specified) to the name of the git repository, i.e. `rebase-helper` in this case.
So we can omit `downstream_package_name`.
`upstream_package_name` means the PyPI package name which is [rebasehelper](https://pypi.org/project/rebasehelper/) - that's why I changed it.
I changed the stdist's base-name as well because that's what made `packit srpm` work locally after t changing the `upstream_package_name`.

I don't know prospector, but will take a look.

nforro:
> `upstream_package_name` means the PyPI package name which is rebasehelper - that's why I changed it.

But PyPI is not the upstream for the Fedora package, github is. Maybe it doesn't really matter, but I suspect spec file discovery could be affected by it.

jpopelka:
Ok, I've removed `upstream_package_name` too.
`packit srpm` works just fine locally so I have no clue how the prospector (which IMHO lints just python files) is related to this yaml only change and how it can affect srpm build.
Let's leave this to Monday :-)

nforro:
I'm not suggesting prospector has something to do with it, but it looks to me like packit is trying to parse the `.prospector.yml` as a spec file (instead of `rebase-helper.spec`).

jpopelka:
I've returned the `specfile_path` and will take a look at the prospector weirdness later.

PR #787, opened 2020-03-05
Fix regression in koji log analysing

FrNecas:


rebasehelper/application.py
```diff
@@ -39,6 +39,7 @@
 from rebasehelper.patcher import Patcher
 from rebasehelper.plugins.plugin_manager import plugin_manager
 from rebasehelper.plugins.checkers import CheckerCategory
+from rebasehelper.plugins.build_tools.rpm.mock import Mock
```
nforro: This completely defeats the purpose of a plugin, making rebase-helper non-functional without mock.

FrNecas: In that case, the method used should be moved somewhere else, perhaps into a helper? I suggest RpmHelper, it doesn't completely fit there but it's the most related in my opinion

nforro: If the method is specific to mock, it has to remain where it is. None of this is supposed to be in `Application`.

FrNecas: It is specific to mock but doesn't require mock to be run. And since koji and copr use mock as well, even these build tools should be able to use the method. Looking at the code, it should be possible to avoid doing anything in `Application` (by using `get_mock_logfile_path` inside koji buildtool) but the problem remains, we have to move the method somewhere else.

FrNecas: Perhaps we could create a class for mock-based buildtools that koji, copr and mock would inherit from and the class could contain the method for determining the correct logfile. However, I am not sure whether this isn't overcomplicated

nforro: Good points. Technically, koji and copr should depend on mock, but that would be impractical, since they don't need it to run (or to be installed at all). Perhaps it could be in `rebasehelper.plugins.build_tools`.

PR #788, opened 2020-03-12
Allow completely disabling checkers

FrNecas:
The problem  doesn't seem to be solvable any other way if
we want to keep the default choices and keep our CLI the
way it is now.


nforro:
I don't think this is a good solution. Let me try to figure out something better.

nforro:
This requires empty string to be explicitly specified. It should work also (preferably) e.g. like this:
```
rebase-helper --pkgcomparetool --buildtool rpmbuild
```

nforro:
And of course also:
```
rebase-helper --pkgcomparetool= --buildtool=rpmbuild
```

rebasehelper/argument_parser.py
```diff
@@ -51,10 +51,12 @@ def __init__(self, option_strings,
                  switch=False,
                  counter=False,
                  append=False,
+                 allow_empty=False,
```
nforro: This works great, but does `allow_empty` make sense now? Shouldn't combination of `nargs` and `const` be enough?

FrNecas: For `pkgcomparetool`, `nargs` and `const` would suffice but what about other options with choices? Without `allow_empty` `--color=""` would be successfully parsed which doesn't make much sense to me.

Also thinking about it now, we should perhaps add `nargs`, `const` and `allow_empty` to some other options (e.g `--versioneer-blacklist`). The option is empty by-default but we should allow users to explicitly specify it...

nforro: > Without `allow_empty` `--color=""` would be successfully parsed which doesn't make much sense to me.

Why would it be, with no `nargs`?

> Also thinking about it now, we should perhaps add `nargs`, `const` and `allow_empty` to some other options (e.g `--versioneer-blacklist`).

Sure, why not.

FrNecas: > Why would it be, with no `nargs`?

Not sure if I understand what you mean by this. With the changes in this PR, if `--pkgcomparetool` is used without additional argument, its value is set to `const` (so `['']`). If an additional argument is used, even if it is an empty string, it is normally read (and due to `type: lambda s: s.split(',')` it's read as `['']`).

So we have to check for empty strings but as I have pointed out in the previous comment, we don't want to enable empty strings by default, it doesn't make sense with some options. `--color` uses choices and its nargs are set to 1 AFAIK.

nforro: > `--color` uses choices and its nargs are set to 1 AFAIK.

Exactly, so it's not `?`, which means it's not optional. `allow_empty` is equivalent to `nargs == '?'`.

rebasehelper/argument_parser.py
```diff
@@ -95,7 +97,8 @@ class CustomArgumentParser(argparse.ArgumentParser):
     def _check_value(self, action, value):
         if isinstance(value, list):
             # converted value must be subset of the choices (if specified)
-            if action.choices is not None and not set(value).issubset(action.choices):
+            empty = value == ['']
```
nforro: Are you sure about this? You are setting `const` to `[]`. Also, shouldn't you compare to `const`?

FrNecas: Good point, `const` should also be set to `['']` (just tested and with an empty list, it would work properly.

nforro: You mean it doesn't work properly with empty list? Why?

FrNecas: `rebase-helper --pkgcomparetool=` is interpreted the same way as `rebase-helper --pkgcomparetool=""`. In such case, `value` is equal to `[""]` so comparing to empty list wouldn't work

nforro: Ok.

PR #790, opened 2020-03-20
Fix some typing issues detected by mypy

FrNecas:
These issues were detected  by mypy with `--check-untyped-defs` argument.

### DIFF
--- a/rebasehelper/plugins/build_log_hooks/files.py
+++ b/rebasehelper/plugins/build_log_hooks/files.py
@@ -28,16 +28,20 @@ import fnmatch
 import logging
 import os
 import re
-from typing import Dict, List, Optional, cast
+from typing import Any, Dict, List, Optional, Tuple, Union, cast
 
 from rebasehelper.logger import CustomLogger
 from rebasehelper.plugins.build_log_hooks import BaseBuildLogHook
 from rebasehelper.types import PackageCategories
 from rebasehelper.helpers.macro_helper import MacroHelper
 from rebasehelper.constants import NEW_BUILD_DIR
+from rebasehelper.specfile import SpecFile
 
 
 logger: CustomLogger = cast(CustomLogger, logging.getLogger(__name__))
+
+AddedFiles = Dict[str, List[str]]
+RemovedFiles = Union[List[str], Dict[str, List[str]]]
 
 
 class Files(BaseBuildLogHook):
@@ -204,7 +208,7 @@ class Files(BaseBuildLogHook):
                     if new_best_match:
                         # the new match is a closer match
                         if new_best_match[0] != best_match:
-                            best_match = new_best_match[0]
+                            best_match = str(new_best_match[0])
                             best_match_section = sec_name
 
         return best_match_section or rebase_spec_file.get_main_files_section() or (files[0] if files else None)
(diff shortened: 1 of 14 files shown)


rebasehelper/plugins/build_log_hooks/files.py
```diff
@@ -39,6 +39,9 @@
 
 logger: CustomLogger = cast(CustomLogger, logging.getLogger(__name__))
 
+AddedFiles = Dict[str, Dict[str, List[str]]]
+RemovedFiles = Dict[str, Union[List[str], Dict[str, List[str]]]]
```
nforro: I'd rather see this in `types.py`, on the other hand, it's local to this build hook (which begs a question - if it's local, why does it need to be typed?).

FrNecas: Exactly my thoughts, since it's local, `types.py` wouldn't make much sense but putting it next to the variable (or function) would make the code quite unreadable. 

I don't exactly know the reason why mypy requires typing in such cases but it seems to me that variables that are returned from methods have to be typed or their types have to be obvious (which isn't the case with lists and dicts).

nforro: Ok, but the method is internal. The only "public" method is `run()`, which returns `Dict[Any, Any]` (and that doesn't seem right either).

FrNecas: Apparently mypy doesn't differentiate between internal and external methods. Perhaps we could just ignore the error since it's internal. I think that the return value of `run` should be `Union[AddedFiles, RemovedFiles]` with that it would even make sense to have types in the internal methods. However mypy throws a different error:
```
rebasehelper/plugins/build_log_hooks/files.py:337: error: Incompatible types in assignment (expression has type "Dict[<nothing>, <nothing>]", variable has type "Union[Dict[str, Dict[str, List[str]]], Dict[str, Union[List[str], Dict[str, List[str]]]]]")```

I will try to find a way how to fix it

Comment (FrNecas) [reply rebasehelper/plugins/build_log_hooks/files.py]: I fixed the type of the returned dict and added typing to the `run` method. I can't find a way to make mypy ignore the internal methods, it just checks everything...

File: rebasehelper/plugins/build_log_hooks/files.py [hunk 2]
```diff
@@ -323,15 +327,16 @@ def _correct_deleted_files(cls, rebase_spec_file, files):
         return result
 
     @classmethod
-    def run(cls, spec_file, rebase_spec_file, results_dir, **kwargs):
+    def run(cls, spec_file: SpecFile, rebase_spec_file: SpecFile, results_dir: str,
+            **kwargs: Dict[str, Any]) -> Tuple[Dict[str, Union[RemovedFiles, AddedFiles]], bool]:
```
nforro: I don't think `**kwargs` is supposed to be `Dict`. It should be `**kwargs: Any` instead.

rebasehelper/spec_content.py
```diff
@@ -168,7 +172,10 @@ def _split_sections(cls, content):
         sections = [('%package', lines[:section_beginnings[0]])]
 
         for i in range(len(section_beginnings) - 1):
-            start = section_beginnings[i] + 1
+            start = section_beginnings[i]
+            if start is None:
+                continue
+            start += 1
```
nforro: This is rather out of scope of this commit. It may be a bug that was revealed by type checking, but that doesn't mean it's related to it.

FrNecas: I don't think that's a bug... `None` will always be the last element and not any other. And as you can see, the `for` loop iterates only through N-1 elements, so `section_beginnings[i]` cannot be `None`. There doesn't seem to be a way how to describe this in type checking. `section_beginnings` has to be `List[Optional[int]]` since we are appending None and due to being `Optional`, we have to check for None in the loop...

We could get rid of appending `None` but then it would require an if-else in the loop (different behaviour at the end of the list).

nforro: Ok, then I think adding `mypy: ignore` and a comment with explanation would be better.

Issue #767, opened 2020-01-28
Unhandled exception in LookasideCacheHelper._upload_source()


nforro:
```
Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/requests/adapters.py", line 479, in send
    r = low_conn.getresponse(buffering=True)
TypeError: getresponse() got an unexpected keyword argument 'buffering'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/requests/adapters.py", line 482, in send
    r = low_conn.getresponse()
  File "/usr/lib64/python3.7/http/client.py", line 1344, in getresponse
    response.begin()
  File "/usr/lib64/python3.7/http/client.py", line 306, in begin
    version, status, reason = self._read_status()
  File "/usr/lib64/python3.7/http/client.py", line 275, in _read_status
    raise RemoteDisconnected("Remote end closed connection without"
http.client.RemoteDisconnected: Remote end closed connection without response

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/nforro/devel/rebase-helper/rebasehelper/cli.py", line 178, in run
    app = Application(config, start_dir, execution_dir, results_dir)
  File "/home/nforro/devel/rebase-helper/rebasehelper/application.py", line 117, in __init__
    upload=not self.conf.skip_upload)
  File "/home/nforro/devel/rebase-helper/rebasehelper/helpers/lookaside_cache_helper.py", line 205, in update_sources
    cls._upload_source(url, package, source_dir, filename, hashtype, hsh)
  File "/home/nforro/devel/rebase-helper/rebasehelper/helpers/lookaside_cache_helper.py", line 181, in _upload_source
    post()
  File "/home/nforro/devel/rebase-helper/rebasehelper/helpers/lookaside_cache_helper.py", line 168, in post
    r = requests.post(url, data=cd, headers=cd.headers, auth=auth)
  File "/usr/lib/python3.7/site-packages/requests/api.py", line 116, in post
    return request('post', url, data=data, json=json, **kwargs)
  File "/usr/lib/python3.7/site-packages/requests/api.py", line 60, in request
    return session.request(method=method, url=url, **kwargs)
  File "/usr/lib/python3.7/site-packages/requests/sessions.py", line 533, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/lib/python3.7/site-packages/requests/sessions.py", line 653, in send
[...]

PR #792, opened 2020-03-25
Fix unhandled exception in LookasideCacheHelper

FrNecas:
Resolves: #767

### DIFF
--- a/rebasehelper/helpers/lookaside_cache_helper.py
+++ b/rebasehelper/helpers/lookaside_cache_helper.py
@@ -179,6 +179,9 @@ class LookasideCacheHelper:
         logger.info('Uploading %s to lookaside cache', path)
         try:
             post()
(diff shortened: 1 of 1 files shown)


rebasehelper/helpers/lookaside_cache_helper.py
```diff
@@ -179,6 +179,9 @@ def post(check_only=False):
         logger.info('Uploading %s to lookaside cache', path)
         try:
             post()
+        except ConnectionError as e:
```
nforro: Did you test this? Shouldn't it be `requests.exceptions.ConnectionError`?

FrNecas: Good catch, didn't realize python has its own `ConnectionError`. I didn't manage to reproduce the issue but your remark makes sense

PR #796, opened 2020-05-12
Update videos in README

nforro:


README.md
```diff
@@ -87,8 +87,12 @@ The following steps describe a rebase process:
...
+## Videos
 
-At DevConf.cz 2016, Tomáš Hozza, Petr Hráček presented on rebase-helper.
+A talk about **rebase-helper** at DevConf.CZ 2020 given by František Nečas:
+
+[![Rebasing RPM packages with rebase-helper - DevConf.CZ 2020](https://img.youtube.com/vi/xI6edOXqk8U/0.jpg)](https://www.youtube.com/watch?v=xI6edOXqk8U)
+
+Another presentation, this time from DevConf.CZ 2016, by Petr Hráček and Tomáš Hozza:
 
 [![rebase-helper presentation at DevConf.cz 2016](https://img.youtube.com/vi/Y-5Qiwaujd8/0.jpg)](https://www.youtube.com/watch?v=Y-5Qiwaujd8)
```
FrNecas, line 98: Since we are using the full presentation name for the first link, I think we should use it here as well to make it consistent. So perhaps `Rebase-helper and Upstream Release Monitoring - DevConf.CZ 2016`?

nforro: Good point, I don't like the capitalized *R*, but I don't think we should change that.

PR #802, opened 2020-06-23
Add rpminspect checkers

FrNecas:


rebasehelper/plugins/checkers/rpminspect.py
```diff
@@ -0,0 +1,141 @@
...
+    }
+    """
+    DEFAULT: bool = True
+
+    CMD: str = 'rpminspect'
+
+    BAD_CHECK: str = 'BAD'
+    VERIFY_CHECK: str = 'VERIFY'
+    OK_CHECK: str = 'OK'
+    INFO_CHECK: str = 'INFO'
```
nforro, line 71: I think it would be better to group these into a list, so you can iterate over all of them later.

rebasehelper/plugins/checkers/rpminspect_rpm.py
```diff
@@ -0,0 +1,66 @@
...
+        os.makedirs(cls.results_dir)
+
+        result = {'path': cls.get_checker_output_dir_short(), 'files': [], 'checks': {}}
+        old_pkgs = results_store.get_old_build()['rpm']
+        new_pkgs = results_store.get_new_build()['rpm']
+        for old_pkg in old_pkgs:
+            if 'debuginfo' in old_pkg:
+                continue
+            name = RpmHelper.split_nevra(os.path.basename(old_pkg))['name']
+            find = [x for x in new_pkgs if RpmHelper.split_nevra(os.path.basename(x))['name'] == name]
```
nforro, line 57: Just a nitpick, I think `found` would make more sense.

rebase-helper.spec
```diff
@@ -42,6 +42,8 @@ Recommends:     licensecheck
 Recommends:     rpmlint
 Recommends:     libabigail
 Recommends:     pkgdiff >= 1.6.3
+Recommends:     rpminspect
+Recommends:     rpminspect-data-fedora
```
nforro, line 46: Is `rpminspect-data-fedora` really needed? As far as I can see, it is required by `librpminspect`, which is a dependency of `rpminspect`.

nforro:
Could `rpminspect-srpm` be listed before `rpminspect-rpm` in the report? It would make more sense that way.

nforro:
Ok, I see, checker reports are sorted alphabetically, not by type.

FrNecas:
That would definitely make sense. It would make sense to order the report as: source checkers, SRPM checkers, RPM checkers. But that's out of scope of this PR, I can fix it in another one (has to be fixed in Text output tool). However, I am not sure if the ordering can be changed in JSON output tool

rebase-helper.spec
FrNecas: Good catch, I followed the README of rpminspect but `rpminspect-data-fedora` doesn't seem to be necessary.

nforro:
> However, I am not sure if the ordering can be changed in JSON output tool

No, and it doesn't make sense there.

nforro:
I just got this while testing:

```
  File "/tmp/rebase-helper/rebasehelper/cli.py", line 179, in run
    app.run()
  File "/tmp/rebase-helper/rebasehelper/application.py", line 776, in run
    self.run_package_checkers(self.results_dir, category=CheckerCategory.SRPM)
  File "/tmp/rebase-helper/rebasehelper/application.py", line 637, in run_package_checkers
    data = plugin_manager.checkers.run(os.path.join(results_dir, 'checkers'),
  File "/tmp/rebase-helper/rebasehelper/plugins/checkers/__init__.py", line 134, in run
    return checker.run_check(results_dir, **kwargs)
  File "/tmp/rebase-helper/rebasehelper/plugins/checkers/rpminspect_srpm.py", line 48, in run_check
    os.makedirs(cls.results_dir)
  File "/usr/lib64/python3.8/os.py", line 223, in makedirs
    mkdir(name, mode)
FileExistsError: [Errno 17] File exists: '/home/nforro/tmp/libtiff/rebase-helper-results/checkers/rpminspect-srpm'
```

FrNecas:
That's weird... Creation of the directory is done in the same way as in other checkers. The directory shouldn't exist as the `rebase-helper-results` directory is freshly created. Any steps to reproduce?

nforro:
Yes, rebase libtiff from 4.0.9 to  4.0.10 with `--favor-on-conflict=upstream`.

nforro:
The check is run for the second time after **files** build log hook fixes the spec.

rebasehelper/plugins/checkers/rpminspect.py
```diff
@@ -0,0 +1,142 @@
...
+    CHECK_RESULTS = (BAD_CHECK, VERIFY_CHECK, OK_CHECK, INFO_CHECK)
```
nforro, line 72: Perhaps `POSSIBLE_CHECK_RESULTS`, for clarity?

PR #803, opened 2020-06-24
Checkers adjustments

FrNecas:
- ordering by type in text report
- remove outputs of SRPM checkers when retrying the build

### DIFF
--- a/rebasehelper/plugins/checkers/__init__.py
+++ b/rebasehelper/plugins/checkers/__init__.py
@@ -25,6 +25,7 @@
 import enum
 import logging
 import os
+import shutil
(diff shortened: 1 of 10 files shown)


rebasehelper/plugins/checkers/__init__.py
```diff
@@ -55,6 +56,16 @@ class BaseChecker(Plugin):
     CATEGORY: Optional[CheckerCategory] = None
     results_dir: Optional[str] = None
 
+    @classmethod
+    def create_results_dir(cls):
```
nforro, line 60: I think `prepare_results_dir` would be a bit better.

PR #805, opened 2020-06-29
Remove excessive blank lines when removing patches

FrNecas:
This is mainly a cosmetic change to avoid leaving 2 or more blank lines
in the specfile after removing a patch that is surrounded by blank
lines.


FrNecas:
Just realised we already have a test for this method. I will add a simple test case for removing comments and empty lines before merging

PR #806, opened 2020-07-08
Rename sources and patches if necessary

FrNecas:
Names of local Sources and Patches may change between 2 versions (for example when %{version} macro is used in their names). This could lead to a failed rebase, SRPM build will fail.

Try to handle the situation by renaming the patches and sources in rebased-sources directory, changing their paths in specfile and then reverting the names back.

### DIFF
--- a/rebasehelper/application.py
+++ b/rebasehelper/application.py
@@ -158,9 +158,7 @@ class Application:
                                         'and no SOURCES argument specified!')
 
         # Prepare rebased_sources_dir
-        self.rebased_repo = self._prepare_rebased_repository(self.spec_file.patches,
-                                                             self.execution_dir,
-                                                             self.rebased_sources_dir)
+        self.rebased_repo = self._prepare_rebased_repository()
 
         # check if argument passed as new source is a file or just a version
         if [True for ext in Archive.get_supported_archives() if self.conf.sources.endswith(ext)]:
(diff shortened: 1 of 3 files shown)


rebasehelper/application.py
```diff
@@ -457,6 +458,12 @@ def _prepare_rebased_repository(self):
         if os.path.isfile(gitignore):
             shutil.copy(gitignore, self.rebased_sources_dir)
 
+        for source, _, source_type in self.spec_file.spc.sources:
+            if not urllib.parse.urlparse(source).scheme and source_type == 1:
+                source_path = os.path.join(self.execution_dir, source)
+                if os.path.isfile(source_path):
+                    shutil.copy(source_path, self.rebased_sources_dir)
+
```
nforro: A nitpick, but I would move this to the top, before copying patches, just so it matches the usual ordering in a spec file.

rebasehelper/application.py
```diff
@@ -457,6 +458,12 @@ def _prepare_rebased_repository(self):
         if os.path.isfile(gitignore):
             shutil.copy(gitignore, self.rebased_sources_dir)
 
+        for source, _, source_type in self.spec_file.spc.sources:
+            if not urllib.parse.urlparse(source).scheme and source_type == 1:
```
nforro, line 505: A comment would be nice. Will you remember what source type `1` means in a few months? :wink:

rebasehelper/specfile.py
```diff
@@ -405,12 +405,13 @@ def _process_patches(self, comment_out=None, remove_patches=None, disable_inappl
 
     @saves
     def update_paths_to_patches(self) -> None:
-        """Fixes paths of patches to make them usable in SPEC file location"""
+        """Fixes paths of patches and sources to make them usable in SPEC file location"""
```
nforro, line 408: I think the method name should be updated as well.

rebasehelper/application.py
```diff
@@ -193,6 +195,47 @@ def _prepare_spec_objects(self):
...
+                continue
+            # Find matching source in the old spec
+            source_old = None
+            for source_old, index_old, source_type_old in self.spec_file.spc.sources:
+                if index_old == index and source_type_old == source_type:
+                    break
+            if not source_old:
+                logger.error('Failed to find the corresponding source to %s in old specfile', source)
+                continue
+
```
nforro, line 222: You could do this instead:

```python
sources = [n for n, i, t in self.spec_file.spc.sources if i == index and t == source_type]
if not sources:
    logger.error('Failed to find the source corresponding to %s in old version spec', source)
    continue
source_old = sources[0]
```

rebasehelper/application.py
```diff
@@ -193,6 +195,45 @@ def _prepare_spec_objects(self):
...
+            new_source_path = os.path.join(self.rebased_sources_dir, source)
+            logger.debug('Renaming %s to %s', old_source_path, new_source_path)
+            try:
+                os.rename(old_source_path, new_source_path)
+            except OSError:
+                logger.error('Failed to rename %s to %s while sanitizing sources', old_source_path, new_source_path)
+
+            # prepend the Source/Path with rebased-sources directory in the specfile
+            to_prepend = os.path.relpath(self.rebased_sources_dir, self.execution_dir)
+            tag = ('Patch{}' if source_type == 2 else 'Source{}').format(index)
```
nforro, line 232: I would prefer this: `tag = '{0}{1}'.format('Patch' if source_type == 2 else 'Source', index)`

Issue #807, opened 2020-07-08
Add functional test for source/path renaming


FrNecas:
Add a functional test (whole rebase) for the feature implemented in https://github.com/rebase-helper/rebase-helper/pull/806

PR #808, opened 2020-07-15
Add a test for patch renaming

FrNecas:
Incorporating the test for the new feature into the existing test turned out to be easier than I expected. I can't think of any other use cases that would need to be tested, so it isn't necessary to create a separate test (at least for now)

Resolves: #807

### DIFF
--- a/rebasehelper/tests/functional/test_rebase.py
+++ b/rebasehelper/tests/functional/test_rebase.py
@@ -54,8 +54,9 @@ class TestRebase:
     TEST_FILES: List[str] = [
         'rebase/test.spec',
         'rebase/applicable.patch',
-        'rebase/conflicting.patch',
         'rebase/backported.patch',
+        'rebase/conflicting.patch',
+        'rebase/renamed-0.1.patch',
     ]
 
     @pytest.mark.parametrize('buildtool', [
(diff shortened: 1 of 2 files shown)


rebasehelper/tests/functional/test_rebase.py
```diff
@@ -56,6 +56,7 @@ class TestRebase:
         'rebase/applicable.patch',
         'rebase/conflicting.patch',
         'rebase/backported.patch',
+        'rebase/0.1.patch',
```
nforro, line 59: Looking at the list, I think it would be better to order the patches alphabetically, as that's how they are ordered in the diff.

rebasehelper/tests/testing_files/rebase/test.spec
```diff
@@ -8,6 +8,7 @@ Source0:        https://integration:4430/pkgs/%{name}/%{name}-%{version}.tar.gz
 Patch0:         applicable.patch
 Patch1:         conflicting.patch
 Patch2:         backported.patch
+Patch3:         %{version}.patch
```
nforro, line 11: I think it would be better to name the patch `renamed-%{version}.patch` for example.

rebasehelper/tests/functional/test_rebase.py
```diff
@@ -90,16 +91,22 @@ def test_rebase(self, buildtool, favor_on_conflict):
...
+            backported_patch, renamed_patch, spec_file = patch
             # Non interactive mode - inapplicable patches are only commented out.
             assert [h for h in spec_file if '+#Patch1:         conflicting.patch\n' in h.target]
             assert [h for h in spec_file if '+#%%patch1 -p1\n' in h.target]
+        assert renamed_patch.is_rename  # renamed patch 0.1.patch to 0.2.patch
+        assert os.path.basename(renamed_patch.source_file) == 'renamed-0.1.patch'
+        assert os.path.basename(renamed_patch.target_file) == 'renamed-0.2.patch'
+        # Check that the renamed patch path is unchanged
+        assert [h for h in spec_file if ' Patch3:         renamed-%{version}.patch\n' in h.source]
+        assert [h for h in spec_file if ' Patch3:         renamed-%{version}.patch\n' in h.target]
```
nforro, line 109: I don't like this very much, I think it's fragile and can easily break if the generated diff changes (you are checking for something in the context of the diff). On the other hand, I think this check makes sense, and I can't think oif any better solution.

FrNecas: Good point, checking unchanged lines in a diff isn't very robust. Perhaps we could instead check that no such line was added or removed in the whole diff? So gather lines `+Patch3: ... ` and `-Patch3: ...` and check if it is an empty list, does this make more sense? The purpose of this check should be the same

nforro: Yes, that's probably better.

FrNecas: Fixed - in the end checking for Patch removal in source of the hunk should be sufficient

nforro: LGTM.

PR #810, opened 2020-07-31
Test adjustments

FrNecas:


rebasehelper/constants.py
```diff
@@ -42,6 +42,7 @@
 VERBOSE_LOG: str = 'verbose.log'
 INFO_LOG: str = 'info.log'
 REPORT: str = 'report'
+PATCH: str = 'changes.patch'
```
nforro, line 45: This feels a bit too general to me, what about `CHANGES_PATCH`?

rebasehelper/tests/functional/conftest.py
```diff
@@ -26,12 +26,14 @@
...
 
-from rebasehelper.constants import RESULTS_DIR, DEBUG_LOG
+from rebasehelper.constants import RESULTS_DIR, DEBUG_LOG, PATCH, REPORT
 
 
 def make_logs_report():
     logs = [
         DEBUG_LOG,
+        PATCH,
+        REPORT + '.json',
```
nforro, line 36: This is a little confusing. Either change the variable from `logs` to `files` for example, or make another list for these two files and iterate through both.

nforro: And the function name should be changed as well.

pylintrc
```diff
@@ -59,7 +59,9 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-#disable=
+
+# Temporary change to avoid E1136: Value 'Optional' is unsubscriptable
+disable=E1136
```
nforro, line 64: I'd add this to `tox.ini` instead, and use a literal constant (I think it is `unsubscriptable-object`) instead of the error code. Also, `Union` triggers the same error, it's not just `Optional`.

rebasehelper/tests/functional/conftest.py
FrNecas: One list seems like the better way to go. I find the patch and report more important than build logs so I'd prefer to have them closer to the top

rebasehelper/tests/functional/conftest.py
```diff
@@ -42,7 +44,7 @@ def make_logs_report():
         'new-build/RPM/mock_output.log',
     ]
     report = []
-    for log in logs:
+    for log in files:
```
nforro, line 47: `for file in files`?

rebasehelper/tests/functional/conftest.py
```diff
@@ -57,4 +59,4 @@ def make_logs_report():
 def pytest_runtest_makereport():
     outcome = yield
     report = outcome.get_result()
-    report.sections.append(('Logs', make_logs_report()))
+    report.sections.append(('Logs', make_files_report()))
```
nforro, line 62: The section should be also renamed. Actually, now that I think about it, various files created by tests are usually referred to as artifacts. Perhaps that would be the best term to use here, instead of files or logs.

tox.ini
```diff
@@ -40,6 +40,7 @@ commands=
            --disable=C,R \
            --disable=fixme \
            --disable=locally-disabled \
+           --disable=unsubscriptable-object \
```
nforro, line 43: I would still add a comment explaining this is a temporary workaround, the issue is reported here: https://github.com/PyCQA/pylint/issues/3760

FrNecas:
The commits have been merged as a part of #815 - closing

PR #820, opened 2020-09-16
Fix failing uploads to lookaside cache

### LINKED ISSUE #801: Lookaside cache upload fails with HTTP code "400 Bad Request"

nforro:
I'm not really sure about this. The requests still fail randomly with `Remote end closed connection without response`.
I'm still considering implementing the upload with `pycurl` after all, because it seems to work without issues.

Fixes #801.

### DIFF
--- a/rebasehelper/helpers/lookaside_cache_helper.py
+++ b/rebasehelper/helpers/lookaside_cache_helper.py
@@ -143,6 +143,7 @@ class LookasideCacheHelper:
                 if check_only:
                     fields.append(('filename', filename))
                 else:
+                    fields.append(('mtime', str(int(os.stat(filename).st_mtime))))
                     with open(path, 'rb') as f:
                         rf = RequestField('file', f.read(), filename)
                         rf.make_multipart()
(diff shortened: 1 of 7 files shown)


nforro:
I think I solved that issue by enabling [opportunistic authentication](https://pypi.org/project/requests-gssapi/#opportunistic-authentication). So the only downside now is the infinite progressbar, which looks a bit weird (especially because it doesn't show transferred bytes).

nforro:
Excerpt from the output:
```
INFO: Uploading vim-go-1.24.tar.gz to lookaside cache
    [     <=>                      ]     in 00:00:01 
INFO: Running checks on source files using 'licensecheck'
```

rebasehelper/helpers/lookaside_cache_helper.py
```diff
@@ -167,24 +167,33 @@ def post(check_only=False):
...
         if state.strip() == b'Available':
             # already uploaded
+            logger.info('Source is already present in lookaside cache')
             return
 
         logger.info('Uploading %s to lookaside cache', path)
         try:
-            post()
-        except requests.exceptions.ConnectionError as e:
+            try:
```
FrNecas, line 216: Is this inner try-finally block necessary? Shouldn't it be sufficient to keep the finally part connected to the outer `try-except-finally` block?

FrNecas (approved):
If the problem with `Remote end closed connection without response` was solved by opportunistic authentication, it looks good to me and we can merge :)

PR #822, opened 2020-09-23
Allow turning off macro preserving in set release or version

dhodovsk:
We (https://github.com/packit/) hit the [issue](https://github.com/packit/dist-git-to-source-git/issues/72) with macros being modified when creating srpm's on converted source-git repositories. We have a [workaround](https://github.com/packit/packit/pull/970) but crave for cleaner solution :)


nforro (approved):
Thanks, just a nitpick, the commit message doesn't make much sense, I think there should be either `turning off`, or `turning off of`.

Issue #826, opened 2020-11-02
Associated comment not removed with Patch tag


nforro:
Rebasing man-pages from 5.08 to 5.09:

```diff
--- a/man-pages.spec
+++ b/man-pages.spec
@@ -6,7 +6,7 @@

 Summary: Linux kernel and C library user-space interface documentation
 Name: man-pages
-Version: 5.08
+Version: 5.09
 Release: 1%{?dist}
 License: GPL+ and GPLv2+ and BSD and MIT and Copyright only and IEEE
 URL: http://www.kernel.org/doc/man-pages/
@@ -42,7 +42,6 @@ Patch0: man-pages-posix-2013-a-pthread_once.patch
 # https://bugzilla.kernel.org/show_bug.cgi?id=53781
 Patch21: man-pages-3.42-close.patch
 # resolves: #1797591
-Patch22: man-pages-5.04-kernel_lockdown.patch

 %description
 A large collection of manual pages from the Linux Documentation Project (LDP).
@@ -52,7 +51,6 @@ A large collection of manual pages from the Linux Documentation Project (LDP).

 %patch0 -p1
 %patch21 -p1
-%patch22 -p1

 # rename posix README so we don't have conflict
 mv %{posix_name}/README %{posix_name}/%{posix_name}.README
```

I'd expect `# resolves: #1797591` to be removed as well.

Issue #811, opened 2020-07-31
Test rebase is broken


FrNecas:
Rpmbuild test rebase with `--favor-on-conflict=upstream` sometimes throws an error due to this assertion: https://github.com/rebase-helper/rebase-helper/blob/master/rebasehelper/tests/functional/test_rebase.py#L110

The error has only appeared recently and appears only once in a few runs. It is caused by %version macro not being expanded in the changelog entry, see the following relevant part of changes.patch from the failed run:

```
 %changelog

+* Fri Jul 31 2020 rebase-helper <rebase-helper@localhost.local> - 0.2-1

+- New upstream release %{version}

+
```

nforro:
A few more details:

I was able to track this down to `MacroHelper.dump()` (but that could be just coincidental and unrelated to the underlying issue), where after a certain (semi-random) number of iterations of the loop the macros seem to just disappear from the global namespace.

PR #829, opened 2020-11-06
Improve comment removal

FrNecas:
Resolves: #826 

There is a problem with the tests, issue with macros seems to have appeared again (probably related to #811 )

### DIFF
--- a/rebasehelper/options.py
+++ b/rebasehelper/options.py
@@ -182,6 +182,13 @@ OPTIONS: Options = [
         "help": "disable inapplicable patches in rebased SPEC file",
     },
     {
+        "name": ["--keep-comments"],
+        "default": False,
+        "switch": True,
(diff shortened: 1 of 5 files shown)


rebasehelper/options.py
```diff
@@ -181,6 +181,13 @@
         "dest": "disable_inapplicable_patches",
         "help": "disable inapplicable patches in rebased SPEC file",
     },
+    {
+        "name": ["--keep-comments"],
+        "default": False,
+        "switch": True,
+        "dest": "keep_comments",
+        "help": "do not remove comments associated to patches when removing them",
```
nforro, line 189: I think it should be `associated with`. But maybe it would be better to describe it as `do not remove associated comments when removing patches from spec file`.

nforro (approved):

PR #836, opened 2020-12-04
update packit.yaml

TomasTomecek:
* dropping outdated command overrides

* renamed dist-git-branch to dist_git_branches


nforro:
I don't know what's the current default for `create_tarball_command`, but I think it needs to be preserved, because of a conflict between Python package name (rebasehelper) and project name (rebase-helper).

https://github.com/rebase-helper/rebase-helper/blob/c67c9140207091e971ac19ffb377a4751b06cf83/setup.py#L66-L86

TomasTomecek:
oh right, I forgot (only build SRPM locally and not the whole thing), let me fix

nforro (approved):

PR #840, opened 2021-01-22
Refactor SpecFile._process_patches()

nforro:
Make it public and usable on its own, not only from within `SpecFile.write_updated_patches()`.

### DIFF
--- a/rebasehelper/specfile.py
+++ b/rebasehelper/specfile.py
@@ -363,18 +363,19 @@ class SpecFile:
         """
         return self.patches['not_applied']
 
-    def _process_patches(self, comment_out=None, remove_patches=None, disable_inapplicable_patches=None):
-        """
-        Comment out and delete patches from SPEC file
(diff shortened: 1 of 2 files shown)


nforro:
Related to https://github.com/packit/packit/issues/1081.

rebasehelper/specfile.py
```diff
@@ -455,7 +460,8 @@ def is_empty(line):
             if patch_removed:
                 # remove the line of the patch that was removed
                 self.removed_patches.append(patch_name)
-                removed_patches.append(tag.index)
+                if tag.index:
```
FrNecas, line 463: Are these added `if`s necessary? As far as I am aware, only patches should get up to this point and they should always have an index, shouldn't they?

nforro: Of course, but the type is still `Optional[int]`.

FrNecas (approved):

FrNecas (approved):
Re-approving with the public API test :)

PR #851, opened 2021-05-20
Use C locale for updating %changelog

stefwalter:
Having the wrong locale running on your machine shouldn't
result in a changelog with invalid syntax or foreign characters.

Example of a bad date here:

https://src.fedoraproject.org/rpms/oci-kvm-hook/pull-request/2#request_diff

Technically setlocale() is not threadsafe. However it is unclear
whether this is a problem for rebase-helper. If it is a problem,
then we must replace strftime() usage in order to fix this problem.

The test aims to be reproducible, despite every system having all
sorts of different locales installed.

### DIFF
--- a/rebasehelper/tests/public_api/test_specfile.py
+++ b/rebasehelper/tests/public_api/test_specfile.py
@@ -22,7 +22,6 @@
 #          Nikola Forró <nforro@redhat.com>
 #          František Nečas <fifinecas@seznam.cz>
 
-import locale
 import os
 import shutil
 
@@ -48,18 +47,6 @@ class TestSpecFile:
     def test_update_changelog(self, spec_object):
         assert spec_object.update_changelog('test2') is None
         assert spec_object.update_changelog(changelog_entry='test') is None
-
-    def test_get_new_log_with_non_c_locale(self, spec_object):
-        # Changelogs should be identical no matter what locale
-        changelog = spec_object.get_new_log("test2")
-        for l in locale.locale_alias:
-            try:
-                locale.setlocale(locale.LC_TIME, l)
-                # Prevents us from trying out strange locale aliases that fail
-                locale.setlocale(locale.LC_TIME, locale.getlocale(locale.LC_TIME))
-            except:
-                continue
-            assert changelog == spec_object.get_new_log("test2")
 
     def test_set_version(self, spec_object):
         assert spec_object.set_version('1.2.3.4') is None
--- a/rebasehelper/tests/public_api/test_specfile.py
+++ b/rebasehelper/tests/public_api/test_specfile.py
@@ -22,6 +22,7 @@
 #          Nikola Forró <nforro@redhat.com>
 #          František Nečas <fifinecas@seznam.cz>
 
+import locale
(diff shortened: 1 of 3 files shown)


rebasehelper/tests/public_api/test_specfile.py
```diff
@@ -48,6 +49,18 @@ def test_update_changelog(self, spec_object):
         assert spec_object.update_changelog('test2') is None
         assert spec_object.update_changelog(changelog_entry='test') is None
 
+    def test_get_new_log_with_non_c_locale(self, spec_object):
```
nforro, line 52: Could you please move the test to `rebasehelper/tests/test_specfile.py`? Public API tests are meant to test the stability of the interface of certain methods, not the functionality.

rebasehelper/tests/public_api/test_specfile.py
```diff
@@ -48,6 +49,18 @@ def test_update_changelog(self, spec_object):
...
+        # Changelogs should be identical no matter what locale
+        changelog = spec_object.get_new_log("test2")
+        for l in locale.locale_alias:
```
nforro, line 55: There is no other locale but C in Fedora container images, so this doesn't actually test anything when run on Travis for example.

rebasehelper/tests/public_api/test_specfile.py
```diff
@@ -48,6 +49,18 @@ def test_update_changelog(self, spec_object):
...
+            try:
+                locale.setlocale(locale.LC_TIME, l)
+                # Prevents us from trying out strange locale aliases that fail
+                locale.setlocale(locale.LC_TIME, locale.getlocale(locale.LC_TIME))
+            except:
```
nforro, line 60: This is making pylint complain, you could use `except Exception:` instead.

rebasehelper/tests/public_api/test_specfile.py
stefwalter: Makes sense. Done in the FIXUP.

rebasehelper/tests/public_api/test_specfile.py
stefwalter: Does this mean we should add another locale to the images to be used in Travis testing? Or that we're fine with this test being a noop in Travis?

rebasehelper/tests/public_api/test_specfile.py
stefwalter: Changed in FIXUP to ```except locale.Error```

stefwalter:
Fixed two changes ... one question above remains.

rebasehelper/tests/public_api/test_specfile.py
nforro: I think it would make sense to add at least one (non-English) locale, but let me take care of that.

nforro (approved):

nforro:
Thanks!

rebasehelper/tests/public_api/test_specfile.py
stefwalter: Thank you.

PR #852, opened 2021-05-20
CONTRIBUTING.md: Start a contributing file

stefwalter:
Here's a file for new contributors. Obviously it's not (yet)
exhaustive, and starts with a few commansd I had to figure out
myself in order to contribute to rebase-helper.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,22 @@
+# Contributing Guidelines
+
+## Running your changes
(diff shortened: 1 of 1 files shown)


nforro:
Thank you. This is definitely a good idea, though there are better ways to selectively run some tests for example, let me take this as it is as a starting point and we'll improve it later.

nforro (approved):

Issue #855, opened 2021-05-26
specfile.set_version(version, preserve_macros=True) recurses infinitely when "Version" has no numbers


TomasTomecek:
```
  File "/home/tt/g/user-cont/packit/packit/base_git.py", line 367, in set_specfile_content                                                                                                    
    self.specfile.set_version(this_version)                                                                                                                                                   
  File "/home/tt/g/rebase-helper/rebase-helper/rebasehelper/specfile.py", line 535, in set_version                                                                                            
    self.set_tag('Version', version, preserve_macros=preserve_macros)                                                                                                                         
  File "/home/tt/g/rebase-helper/rebase-helper/rebasehelper/specfile.py", line 108, in wrapper                                                                                                
    func(spec, *args, **kwargs)                                                                                                                                                               
  File "/home/tt/g/rebase-helper/rebase-helper/rebasehelper/specfile.py", line 846, in set_tag                                                                                                
    value = _process_value(self.get_raw_tag_value(tag) or '', value)                                                                                                                          
  File "/home/tt/g/rebase-helper/rebase-helper/rebasehelper/specfile.py", line 754, in _process_value                                                                                         
    value, _ = _expand_macros(curval)                                                                                                                                                         
  File "/home/tt/g/rebase-helper/rebase-helper/rebasehelper/specfile.py", line 643, in _expand_macros                                                                                         
    rep, new_macros = _expand_macros(value)
  File "/home/tt/g/rebase-helper/rebase-helper/rebasehelper/specfile.py", line 643, in _expand_macros
    rep, new_macros = _expand_macros(value)
  File "/home/tt/g/rebase-helper/rebase-helper/rebasehelper/specfile.py", line 643, in _expand_macros
    rep, new_macros = _expand_macros(value)
  [Previous line repeated 974 more times]
  File "/home/tt/g/rebase-helper/rebase-helper/rebasehelper/specfile.py", line 638, in _expand_macros
    for macro, span in _find_macros(s):
  File "/home/tt/g/rebase-helper/rebase-helper/rebasehelper/specfile.py", line 617, in _find_macros
    macro_re = re.compile(r'%(?P<brace>{\??)?(?P<name>\w+)(?(brace)})')
```

```
Version: %{general_version}%{?prerel:~%{prerel}}
```

I'm wondering what would be the right thing to do here - that's why I haven't sent a PR with a fix because I'm not sure what to do here.

[...]

PR #856, opened 2021-06-28
Fix macro value detection in set_tag

FrNecas:
Resolves: https://github.com/rebase-helper/rebase-helper/issues/855

Tried on the repo @TomasTomecek linked in the issue and it seems to work correctly now (also added a test that struggled from the same problem)

Previously, the regular expression for finding macro definitions didn't
correctly consider the possibility of having multiple macros inside
value of a macro (e.g. %{major}.%{minor}) resulting in incorrect value
parsing in some cases where macro definitions ended with regular string
characters as found in the following reproducer:

    %global general_version %{pybasever}.5
    #global prerel ...
    %global upstream_version %{general_version}%{?prerel}

The value of 'general_version' macro would previously span across all
three lines due to greedy matching and re.DOTALL which resulted in
incorrect behaviour (infinite recursion in this case).

Change the regex to correctly expect the possibility of multiple macros
along with normal text around as well. Allow line breaks only in the
body of a macro (that could for example be %{lua: ...}) using [\s\S]
instead of re.DOTALL and prohibit line breaks elsewhere.

### DIFF
--- a/rebasehelper/specfile.py
+++ b/rebasehelper/specfile.py
@@ -577,21 +577,25 @@ class SpecFile:
         macro_def_re = re.compile(
             r'''
             ^
-            (?P<cond>%{!?\?\w+:\s*)?
-            (?(cond)%global|%(global|define))
+            (?P<cond>%{!?\?\w+:\s*)?          # conditional macro, keep track of the opening brace
+            (?(cond)%global|%(global|define)) # only global can be conditional, not define
             \s+
-            (?P<name>\w+)
-            (?P<options>\(.+?\))?
+            (?P<name>\w+)                     # macro name
+            (?P<options>\(.+?\))?             # optional macro arguments
             \s+
             (?P<value>
-                (%((?P<b>{)|(?P<s>\()))?
-                .+?
-                (?(b)})(?(s)\))
+                (                             # the value consists of multiple macros and/or other text
+                    %(?![{(]) |               # match % but we don't want to consume macros or shell, ignore %( and %{
+                    [^%] |                    # other regular text, do not match %, force the line above to match it
+                    (%((?P<b>{)|(?P<s>\()))?  # macro or shell, track the parenthesis type
+                    [\s\S]+?                  # macro and shell may be multiline, match ASAP using non-greedy matching
+                    (?(b)})(?(s)\))           # parenthesis must be closed if it was used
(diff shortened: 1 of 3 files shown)


rebasehelper/specfile.py
```diff
@@ -584,14 +584,18 @@ def set_tag(self, tag: str, value: str, preserve_macros: bool = False) -> None:
             (?P<options>\(.+?\))?
             \s+
             (?P<value>
-                (%((?P<b>{)|(?P<s>\()))?
-                .+?
-                (?(b)})(?(s)\))
+                (
+                    %(?!{) |
```
nforro, line 588: What about `%()` (shell expression)? Than can be multiline and can contain pretty much whatever character.

nforro: Could you perhaps add comments, since this is a verbose regexp?

FrNecas: Nice catch, forgot to consider `%(` in the lookahead and it would be consumed by `[^%]` and hence wouldn't allow multiline. Should be easy to fix. And comments are a great idea, will add some. Btw travis seems not to respond, hope it will wake up once I force push the changes...

nforro (approved):
Looks good, thanks!

TomasTomecek (commented):
Thanks!

TomasTomecek:
seems like docutils is borked in rawhide:
```
Exception occurred:
  File "/usr/lib/python3.10/site-packages/docutils/nodes.py", line 625, in __getitem__
    return self.attributes[key]
KeyError: 'domain'
```

nforro:
> seems like docutils is borked in rawhide:
> 
> ```
> Exception occurred:
>   File "/usr/lib/python3.10/site-packages/docutils/nodes.py", line 625, in __getitem__
>     return self.attributes[key]
> KeyError: 'domain'
> ```

Yes, that's being taken care of :)

nforro:
> > seems like docutils is borked in rawhide:
> > ```
> > Exception occurred:
> >   File "/usr/lib/python3.10/site-packages/docutils/nodes.py", line 625, in __getitem__
> >     return self.attributes[key]
> > KeyError: 'domain'
> > ```
> 
> Yes, that's being taken care of :)

Hopefully fixed in #858.

Issue #859, opened 2021-07-07
Switch from distutils to setuptools (or something else)


nforro:
> The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives

PR #861, opened 2021-07-12
Replace soon-to-be deprecated distutils

FrNecas:
Resolves: https://github.com/rebase-helper/rebase-helper/issues/859

### DIFF
--- a/containers/webhooks.py
+++ b/containers/webhooks.py
@@ -8,8 +8,8 @@ import tempfile
 import threading
 import traceback
 
-import distutils.core
 import git
+import setuptools.sandbox
 import twine.commands.upload
 
 
@@ -22,8 +22,8 @@ class PyPI:
             repo = git.Repo.clone_from(url, wd)
             repo.git.checkout(tag)
             sys.path.insert(0, wd)
-            distutils.core.run_setup(os.path.join(wd, 'setup.py'),
-                                     ['sdist', 'bdist_wheel', '--universal'])
+            setuptools.sandbox.run_setup(os.path.join(wd, 'setup.py'),
+                                         ['sdist', 'bdist_wheel', '--universal'])
             twine.commands.upload.main([os.path.join(wd, 'dist', '*')])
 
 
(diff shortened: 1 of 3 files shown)


rebasehelper/helpers/input_helper.py
```diff
@@ -36,6 +35,29 @@ class InputHelper:
 
     """Class for command line interaction with the user."""
 
+    @staticmethod
+    def strtobool(message):
```
nforro, line 39: Shoudn't this be case-insensitive?

nforro: And perhaps it would be better to make it a nested function of `get_message()`.

containers/webhooks.py
```diff
@@ -22,8 +21,20 @@ def release(url, tag):
...
+            global_vars = {'__file__': setup_path}
+            try:
+                sys.argv[0] = setup_path
+                sys.argv[1:] = ['sdist', 'bdist_wheel', '--universal']
+                with open(setup_path, 'rb') as f:
+                    exec(f.read(), global_vars)
+            except SystemExit:
+                pass
+            finally:
+                sys.argv = argv_backup
```
nforro, line 37: Couldn't we perhaps use `setuptools.sandbox.run_setup()`?

https://github.com/pypa/setuptools/blob/38974172d0ff7ead2a2b212c92f303438bcdf1a2/setuptools/sandbox.py#L241-L242

FrNecas: Thanks, I didn't find an equivalent but this seems to be exactly what we need

rebasehelper/helpers/input_helper.py
FrNecas: Good catch with the sensitivity. Not completely sure about making it nested, I feel like it would make it a bit less readable but I see your point that it is most likely not going to be used anywhere but in `get_message`.

nforro: If you want to leave it as is, I'd at least make `get_message()` a classmethod.

nforro (approved):

PR #862, opened 2021-07-13
Make container with tests privileged

FrNecas:
Without making it privileged, tests using mock fail in Github Actions.
This used to work in Travis CI but apparently Github uses a newer OS or
docker version.

This fixes the behaviour, see https://github.com/rebase-helper/rebase-helper/actions/runs/1026615062

### DIFF
--- a/containers/docker-compose.yml
+++ b/containers/docker-compose.yml
@@ -6,8 +6,7 @@ services:
         build:
             context: ..
             dockerfile: containers/Containerfile.tests
-        cap_add:
-            - SYS_ADMIN
+        privileged: true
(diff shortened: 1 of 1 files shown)


containers/docker-compose.yml
```diff
@@ -6,6 +6,7 @@ services:
         build:
             context: ..
             dockerfile: containers/Containerfile.tests
+        privileged: true
         cap_add:
             - SYS_ADMIN
```
nforro, line 11: I think this is redundant now.

nforro (approved):

PR #866, opened 2021-07-26
Fix release checkout reference syntax

FrNecas:


.github/workflows/ci.yml
```diff
@@ -10,8 +10,8 @@ jobs:
     strategy:
       matrix:
         include:
-          - distro: fedora-latest
-          - distro: fedora-rawhide
+          - distro: latest-723d477
+          - distro: rawhide-723d477
```
nforro, line 14: Perhaps you could add a comment that this is a temporary workaround.

PR #870, opened 2021-09-14
Update copr buildtool based on new copr responses

FrNecas:
@nforro I didn't manage to reproduce the issue that you pointed out today, I tried building `vim-go` to avoid waiting for hours for the build to finish. I found a different problem - the responses from copr changed and the error is now in a dict-like structure so rebase-helper threw an error that the copr repository already exists and cannot be created while trying to build new packages.

I am not sure what's the better way to handle this - either we can parse the error (the current implementation in the PR) or just modify the if below (to `if 'You already have project named' in error:`). I like this approach a bit more as it is more "precise" and could give us some more options for error checking in the future if we need it.

As for the problem you messaged me about, could it have been a copr outage or something similar? Looks like rebase-helper failed to fetch the build status but it seems to work for my simple case (see https://copr.fedorainfracloud.org/coprs/fnecas/rebase-helper-vim-go/builds/ - the 2 newest builds are from a single rebase after this fix). The error message isn't saying much, the API URL is correct. You could have also lost internet connection - rebase-helper checks the status every 10 seconds and if it fails to check, it errors out. Maybe we could change this?

### DIFF
--- a/rebasehelper/helpers/copr_helper.py
+++ b/rebasehelper/helpers/copr_helper.py
@@ -29,7 +29,9 @@ import urllib.parse
 from typing import cast
 
 import pyquery  # type: ignore
-from copr.v3 import Client, CoprConfigException, CoprNoConfigException, CoprRequestException  # type: ignore
+from copr.v3 import (  # type: ignore
+    Client, CoprConfigException, CoprNoConfigException, CoprRequestException, CoprNoResultException)
+
 
 from rebasehelper.exceptions import RebaseHelperError
 from rebasehelper.helpers.download_helper import DownloadHelper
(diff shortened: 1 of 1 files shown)


nforro:
> As for the problem you messaged me about, could it have been a copr outage or something similar? Looks like rebase-helper failed to fetch the build status but it seems to work for my simple case (see https://copr.fedorainfracloud.org/coprs/fnecas/rebase-helper-vim-go/builds/ - the 2 newest builds are from a single rebase after this fix). The error message isn't saying much, the API URL is correct. You could have also lost internet connection - rebase-helper checks the status every 10 seconds and if it fails to check, it errors out. Maybe we could change this?

I'm not sure, the copr build was stuck in queue for a while and then it was in `Importing` state when rebase-helper threw the error. I couldn't reproduce it again today (this time the copr build itself timeouted). It could have been a network problem as well, my ethernet is a bit unstable sometimes.

rebasehelper/helpers/copr_helper.py
```diff
@@ -64,6 +65,15 @@ def create_project(cls, client, project, chroots, description, instructions, per
...
             error = e.result.error
+            # In newer versions of copr, the response is JSON-like, not
+            # a simple text response but contains apostrophes instead of
+            # quotes, hence json.loads() is not usable. Read it as dict literal.
+            try:
+                response = ast.literal_eval(error)
+                if isinstance(response, dict):
+                    error = response['name'][0]
+            except (ValueError, IndexError, KeyError):
+                pass
```
nforro, line 76: This is rather strange, and is bound to break eventually, but it's no worse than just checking for a substring I suppose.

nforro (approved):

rebasehelper/helpers/copr_helper.py
FrNecas: Agreed. The problem is that the `error` in this case looked something like `"{'name': "You already have project named 'rebase-helper-....'"}"` (escaping of quotes is left out as simplification). I initially wanted to replace `'` with `"` and parse it as json but due to the value, the replacement creates an invalid JSON so parsing fails. That's just odd formatting. Can you see a better way of handling this? I guess we could also ask around the copr team...

FrNecas: Could it be a bug in the python binding? I wouldn't expect the service itself to return such a weird format

nforro: > Could it be a bug in the python binding? I wouldn't expect the service itself to return such a weird format

I'm not sure, we can ask copr team for sure.

There is another copr-related problem I just found. rebase-helper creates a temporary project for an old build, but then it tries to create another project for a new build, but with the same name, and that fails. Should I open a separate issue for that?

nforro (approved):

PR #872, opened 2021-09-23
spec: sources may not be defined

TomasTomecek:
there can be spec files which are meant to define dependencies only thus
they ship no files and have no sources

Please let me know where should I create a test case.

https://sentry.io/share/issue/1e27f7ad5acd49c29f2c27957741ebb4/

### DIFF
--- a/rebasehelper/specfile.py
+++ b/rebasehelper/specfile.py
@@ -235,7 +235,10 @@ class SpecFile:
     ###########################
 
     @staticmethod
-    def _identify_main_source(spec: rpm.spec) -> int:
+    def _identify_main_source(spec: rpm.spec) -> Optional[int]:
+        # a spec file does not need to have sources
+        if not spec.sources:
+            return None
         # lowest index is the main source
         return min([s[1] for s in spec.sources if s[2] == 1])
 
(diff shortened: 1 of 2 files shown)


rebasehelper/specfile.py
```diff
@@ -270,11 +273,14 @@ def get_archive(self):
         """
         return os.path.basename(self.get_sources()[0])
 
-    def _get_raw_source_string(self, source_num: int) -> Optional[str]:
+    def _get_raw_source_string(self, source_num: Optional[int]) -> Optional[str]:
+        if source_num is None:
+            return ""
```
nforro, line 278: I think it would be better to return `None`, since `SpecFile.get_raw_tag_value()` also returns `None` if there is no matching tag.

nforro:
Thanks. I think it would make sense to modify/extend `TestSpecFile.test__get_raw_source_string()` to cover this case. Perhaps just adding `assert mocked_spec_object._get_raw_source_string(None) == None` would do it.

TomasTomecek:
@nforro thanks for the review!

I pushed those updates

nforro (approved):

TomasTomecek:
I'd appreciate a release with this fix :) since the referenced PR cannot be built. (although there should be a workaround with defining a dummy main source)

nforro:
Sure, I'll do a new release ASAP.

PR #877, opened 2021-11-16
Warn the user about a potential rename

FrNecas:
In case of renames of files prepended by some directives such as %doc or
%license, we cannot automatically detect the rename. The directive
handles the installation of the file so removing the file makes rpmbuild
or mock not complain but the file is not there if it was renamed. This
can be common in case of renames such as README -> README.md. Warn the
user that this is happening and that it may be necessary to check the
specfile.

Signed-off-by: František Nečas <fnecas@redhat.com>

### DIFF
--- a/rebasehelper/plugins/build_log_hooks/files.py
+++ b/rebasehelper/plugins/build_log_hooks/files.py
@@ -41,7 +41,9 @@ from rebasehelper.specfile import SpecFile
 logger: CustomLogger = cast(CustomLogger, logging.getLogger(__name__))
 
 AddedFiles = Dict[str, List[str]]
-RemovedFiles = Union[List[str], Dict[str, List[str]]]
+RemovedFromSections = Dict[str, List[str]]
+RemainingFiles = List[str]
+RemovedFiles = Union[RemainingFiles, RemovedFromSections]
 
 
 class Files(BaseBuildLogHook):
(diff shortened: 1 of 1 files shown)


rebasehelper/plugins/build_log_hooks/files.py
```diff
@@ -277,35 +277,52 @@ def _correct_deleted_files(cls, rebase_spec_file, files):
...
                             del split_line[j]
                             del original_line[len(directives) + j]
                             files.remove(deleted_file)
                             result['removed'][sec_name].append(original_file)
                             logger.info("Removed %s from '%s' section", original_file, sec_name)
+                            if warn_about_rename:
+                                logger.warning("The installation of %s was done through a %s directive and has now "
+                                               "been removed. The file may have been renamed and rebase-helper cannot "
+                                               "automatically detect it. A common example of this problem is renaming "
+                                               "of README to README.md", original_file, prepended_directive)
```
nforro, line 325: I would change the formulation a bit:

```
The installation of %s was handled by %s directive and the file has now been removed.
The file may have been renamed and rebase-helper cannot automatically detect it.
A common example of this is renaming README to README.md.
```

nforro: And maybe add another sentence, something like: `It might be necessary to re-add such renamed file to the rebased specfile manually`.

rebasehelper/plugins/build_log_hooks/files.py
```diff
@@ -256,75 +258,130 @@ def _correct_missing_files(cls, rebase_spec_file, files):
...
+    def _get_line_directives(cls, split_line) -> Tuple[List[str], Optional[str]]:
+        """Gathers directives present in the line.
+
+        Args:
+            split_line: Line split in spaces. Each element must either be
+                a directive or a file path.
+
+        Returns:
+            A tuple of 2 elements. The first is a list of used directives.
+            The second is a directive which influences the following paths
```
nforro, line 270: I'd rather say subsequent paths, following is a bit ambiguous.

rebasehelper/plugins/build_log_hooks/files.py
```diff
@@ -256,75 +258,130 @@ def _correct_missing_files(cls, rebase_spec_file, files):
         return result
 
     @classmethod
-    def _correct_deleted_files(cls, rebase_spec_file, files):
+    def _get_line_directives(cls, split_line) -> Tuple[List[str], Optional[str]]:
+        """Gathers directives present in the line.
+
+        Args:
+            split_line: Line split in spaces. Each element must either be
```
nforro, line 265: Shouldn't it be split at spaces? And I think whitespace(s) would be more accurate.

rebasehelper/plugins/build_log_hooks/files.py
```diff
@@ -256,75 +258,130 @@ def _correct_missing_files(cls, rebase_spec_file, files):
...
+            # expand to empty string.
+            expanded = MacroHelper.expand(sec_content[i])
+            if not original_line or not expanded or any(k in expanded for k in cls.PROHIBITED_KEYWORDS):
+                i += 1
+                continue
+            split_line = original_line[:]
+            # Keep track of files which could possibly be renamed but not
+            # detected by the hook. %doc and %license files are the 2 examples
+            # of this. If %doc README is renamed to README.md, the hook will
+            # simply remove it but README.md # won't be added (it is installed
```
nforro, line 310: There seems to be an extra # here.

rebasehelper/plugins/build_log_hooks/files.py
```diff
@@ -256,75 +258,130 @@ def _correct_missing_files(cls, rebase_spec_file, files):
...
+                    original_file = original_line[len(directives) + j]
+
+                    del possible_rename[j]
+                    del split_line[j]
+                    del original_line[len(directives) + j]
+                    files.remove(deleted_file)
+                    result['removed'][sec_name].append(original_file)
+                    logger.info("Removed %s from '%s' section", original_file, sec_name)
+                    if warn_about_rename:
+                        logger.warning("The installation of %s was handled by %s directive and the file has now bee n"
```
nforro, line 340: There is an extra space here

nforro: Actually, it's not extra, it's supposed to be after "been".

nforro (approved):
Seems to work fine, thanks!

PR #879, opened 2021-12-13
Fix parsing of %dump output

nforro:
Macros can have (and do have) multiline bodies. Account for that.


FrNecas (approved):
Sorry for the late review, LGTM.

PR #887, opened 2022-03-21
packit: use bumpspec to update spec

TomasTomecek:
...during the CI builds

Resolves: https://github.com/packit/packit/issues/1520

The problem is that a macro `%dump` which comes from a commit title is
interpreted and makes the spec unparsable.


nforro (approved):

TomasTomecek:
once you create a new release, we can revert :)

PR #890, opened 2022-06-07
Add support for rpmuncompress

nforro:
RPM >= 4.18 uses *rpmuncompress* rather than calling *tar* or *unzip* directly. Account for that in `SpecFile.find_archive_target_in_prep()`.

Solves FTBFS in Fedora Rawhide.


FrNecas (approved):
Sorry, I somehow missed this one.

PR #898, opened 2022-07-13
Switch to podman for CI

nforro:


nforro:
Unfortunately there seems to be an issue with podman in GitHub Actions right now:

https://github.community/t/error-running-container-error-from-usr-bin-crun-creating-container-for-bin-sh-c-dnf-update-y-nobest-sd-bus-call-connection-reset-by-peer-1994/261522/2

nforro:
> Unfortunately there seems to be an issue with podman in GitHub Actions right now

Since it's still not fixed, I've added a workaround for that.

nforro:
@FrNecas could you please take a look?

.github/workflows/ci.yml
```diff
@@ -30,4 +25,4 @@ jobs:
       - name: Run the tests
         env:
           DISTRO: ${{ matrix.distro }}
-        run: make PYTEST_ADDOPTS="${PYTEST_ADDOPTS}" test-docker
```
FrNecas, line 33: Why drop the `PYTEST_ADDOPTS` part? That should still be relevant to running in podman, is it not? This way, functional tests are run when running in cronjob.

containers/Makefile.podman
```diff
@@ -2,7 +2,8 @@ DISTRO ?= fedora-latest fedora-rawhide
 
 PYTEST_ADDOPTS ?= -m 'standard and not long_running'
 
-PODMAN := sudo podman
+# https://github.com/containers/podman/issues/14920
+PODMAN := sudo -E XDG_RUNTIME_DIR= podman
```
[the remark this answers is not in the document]
nforro: Good point, that's probably better, I'll fix it.

.github/workflows/ci.yml
FrNecas: You are right, it's not necessary

FrNecas (approved):

PR #899, opened 2022-07-30
Use rpkg for lookaside cache client implementation

nforro:


tox.ini
```diff
@@ -27,10 +27,12 @@ passenv=
     PYTEST_ADDOPTS
     REQUESTS_CA_BUNDLE
 commands=
+    pip install https://github.com/pythongssapi/python-gssapi/archive/refs/tags/v1.8.0.tar.gz
```
FrNecas, line 30: Is this a temporary workaround (is there some hope that this is fixed in rawhide)? If so, a comment would be nice to not forget to remove it somewhere down the line.

nforro: Oh, yes, sorry. The comment was there, but I removed the workaround because it seemed not to be needed anymore, but I had to add it back and forgot about the comment :slightly_smiling_face:

PR #900, opened 2022-12-12
Switch to specfile library

### LINKED ISSUE #884: Migrate to specfile library

nforro:
I've tried to split the changes to more chunks, but it's still a lot to go through.

Fixes #884.


pylintrc
```diff
@@ -1,314 +1,619 @@
-[MASTER]
+[MAIN]
```
FrNecas, line 1: Are there any significant changes to the pylintrc or is it just an update to the current config schema?

nforro: This is the only notable config option:
https://github.com/rebase-helper/rebase-helper/blob/8c06c0a33cd0de4a9657d5ca42345fdd8e0e4410/pylintrc#L540-L541

nforro: The rest should be unchanged/default.

rebasehelper/options.py
```diff
@@ -181,13 +181,6 @@
         "dest": "disable_inapplicable_patches",
         "help": "disable inapplicable patches in rebased SPEC file",
     },
-    {
-        "name": ["--keep-comments"],
```
FrNecas, line 185: Specfile keeps the comments by default right? I am wondering if there shouldn't be an option to override this behaviour. On the other hand, keeping the comments is usually desired and will result in a "false positive" not that often, I assume

rebasehelper/plugins/build_log_hooks/files.py
```diff
@@ -202,16 +204,16 @@ def _get_best_matching_files_section(cls, rebase_spec_file, file):
         best_match = ''
         best_match_section = ''
         files = []
-        for sec_name, sec_content in rebase_spec_file.spec_content.sections:
-            if sec_name.startswith('%files'):
-                files.append(sec_name)
-                for line in sec_content:
-                    new_best_match = difflib.get_close_matches(file, [best_match, MacroHelper.expand(line)])
+        for section in rebase_spec_file.spec.sections().content: # pylint: disable=no-member
```
FrNecas, line 207: Is the pylint disable still necessary when the new context managers are already released?

rebasehelper/plugins/spec_hooks/pypi_url_fix.py
```diff
@@ -49,14 +49,10 @@ class PyPIURLFix(BaseSpecHook):
...
     def run(cls, spec_file, rebase_spec_file, **kwargs):
-        for tag in rebase_spec_file.tags.filter(name='URL'):
-            value = rebase_spec_file.get_raw_tag_value(tag.name)
-            if value:
-                rebase_spec_file.set_raw_tag_value(tag.name, cls._transform_url(value))
-        for tag in rebase_spec_file.tags.filter(name='Source*'):
-            value = rebase_spec_file.get_raw_tag_value(tag.name)
-            if value:
-                rebase_spec_file.set_raw_tag_value(tag.name, cls._transform_sources_url(value))
+        rebase_spec_file.spec.url = cls._transform_url(rebase_spec_file.spec.url)
```
FrNecas, line 52: I am still amazed at how much specfile simplifies some things :)

rebasehelper/plugins/spec_hooks/replace_old_version.py
```diff
@@ -116,34 +115,35 @@ def run(cls, spec_file: SpecFile, rebase_spec_file: SpecFile, **kwargs: Any):
...
+                            if tag.normalized_name.startswith('Source') or tag.normalized_name.startswith('Patch'):
+                                for sub_pattern, repl in subversion_patterns[1:]:
+                                    updated_value = sub_pattern.sub(repl, updated_value)
+                            tag.value = updated_value
+                    continue
+                for index, line in enumerate(section):
+                    start, end = spec_file.get_comment_span(line, section.normalized_name)
+                    updated_line = subversion_patterns[0][0].sub(subversion_patterns[0][1], line[:start])
+                    section[index] = updated_line + line[start:end]
+            rebase_spec_file.save()
```
FrNecas, line 149: Is save still necessary? Isn't this handled by autosave in the context manager?

rebasehelper/specfile.py
```diff
@@ -23,35 +23,58 @@
...
+    '_mandir',
+    '_sbindir',
+    '_sharedstatedir',
+    '_sysconfdir',
+    'python2_sitelib',
+    'python3_sitelib',
+]
+
+# Comments in these sections can only be on a separate line.
+DISALLOW_INLINE_COMMENTS: List[str] = [
```
FrNecas, line 70: I am wondering if this shouldn't somehow be included in specfile since it deals with specfile parsing but looks like it's not needed in the library itself?

rebasehelper/specfile.py
```diff
@@ -568,297 +521,7 @@ def set_extra_version(self, extra_version: Optional[str], version_changed: bool)
         self.set_release(release)
         # TODO: in some cases it might be necessary to modify Source0
 
-    @saves
-    def set_tag(self, tag: str, value: str, preserve_macros: bool = False) -> None:
```
FrNecas, line 572: So happy that this beast is gone :D

FrNecas (commented):
Very nice! A few comments and questions. I will surely do another pass in January, it was a bit too much info to comprehend everything right now.

We should also update the documentation, mainly here: https://rebase-helper.readthedocs.io/en/latest/user_guide/writing_plugins.html

rebasehelper/options.py
nforro: No, vice versa, specfile removes associated comments when removing a source/patch. And it doesn't make much sense to me that someone would want the comments to be left in the spec file (since they are tied to the source/patch being removed).

nforro: I see you added the option in https://github.com/rebase-helper/rebase-helper/commit/4eef5e263ae8892dbbf2d82c07a2e507ab6fb282. Do you remember what for?

rebasehelper/plugins/build_log_hooks/files.py
nforro: I think so, it's a false positive by pylint.

rebasehelper/plugins/spec_hooks/replace_old_version.py
nforro: I kept it there because it was there originally, but it probably was unnecessary even then. I'll remove it.

rebasehelper/specfile.py
nforro: The thing is, comments in spec files can never be inline. But sections such as `%prep`, `%build` etc. actually contain shell scripts, where inline comments are valid.
Perhaps it would make sense if `Section` had something like a `shell` attibute, and then we could use that rather than this constant?

rebasehelper/specfile.py
nforro: Same here :slightly_smiling_face: The new implementation in specfile is way better IMHO, though not 100%.

nforro:
> We should also update the documentation, mainly here: https://rebase-helper.readthedocs.io/en/latest/user_guide/writing_plugins.html

Good point, I'll update it.

rebasehelper/options.py
FrNecas: Oh, yes, I got a bit confused. Right, it doesn't make sense to keep the comments. The commit was trying to solve https://github.com/rebase-helper/rebase-helper/issues/826 but I am not sure why I added the option to keep the comments. Maybe for backwards compatibility so that the users could change the behavior to the old one if the removal made some mess (e.g. the comment looked to be related to the patch but semantically, it wasn't). I am fine with dropping the option.

rebasehelper/plugins/spec_hooks/replace_old_version.py
nforro: Actually, autosaving is not enabled and I'd rather keep the explicit save here.

rebasehelper/specfile.py
nforro: https://github.com/packit/specfile/pull/170

nforro:
> > We should also update the documentation, mainly here: https://rebase-helper.readthedocs.io/en/latest/user_guide/writing_plugins.html
> 
> Good point, I'll update it.

https://rebase-helper.readthedocs.io/en/specfile/user_guide/writing_plugins.html

nforro:
Tests will be green once https://github.com/packit/specfile/pull/170 is merged.

setup.py
```diff
@@ -54,6 +54,7 @@ def get_requirements():
         'rpkg',
         'unidiff',
         'GitPython',
+        'specfile @ git+https://github.com/packit/specfile',
```
FrNecas, line 57: Is installing the latest release from github intended?

FrNecas (approved):
The code looks good as far as I can tell (I definitely may have missed something, a lot to review at once). Two questions/remarks  about dependencies and building, otherwise LGTM

.packit.yml
```diff
@@ -1,32 +1,45 @@
...
+
 jobs:
   - job: copr_build
     trigger: pull_request
-    metadata:
-      targets:
-        - fedora-all
-        - epel-8-x86_64
+    additional_repos:
+      - copr://packit/packit-dev # for the latest specfile
```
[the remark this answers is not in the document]
nforro: Both of these were only temporary, I've removed them.

PR #909, opened 2023-02-27
Switch to pyproject.toml

nforro:


FrNecas (approved):
Wow, really nice simplification, especially of the specfile. Looks good as far as I can tell. We should probably also update this doc page https://rebase-helper.readthedocs.io/en/latest/user_guide/installation.html which refers to setup.py.

FrNecas:
Hm, `Writing plugins` also contains a setup.py example. I am not sure if combination of pyproject with setup.py will work :thinking:

nforro:
Good point, I'll update the documentation. I don't know if entrypoints can be specified in pyproject.toml directly (but I'll find out), but they can be defined in setup.cfg.

nforro:
Updated docs: https://rebase-helper.readthedocs.io/en/pyproject.toml/

FrNecas (approved):
:+1:

